PR #731, opened 2015-01-14
Add "Learning Hypertext Transfer Protocol" to Guides.pod

henryan:
This pull request is more indented to see if it is can be considered to add the HTTP protocol item for newbies to start with Mojolicious.
I once wanted to learn Mojolicous, but I had a hard time to understand
the example code, I tried again two days ago, still didn't get it,
later, I realized that the problem is that I didn't have a basic
understanding of the HTTP.  I'm a newbie to web development.  I probably
would have started my current little project with Mojolicioius instead
of asking colleagues(who has Perl programming experience but only has some experience in developing web site using PHP ) to develop the web site using PHP if I realized this earlier.

### DIFF
--- a/lib/Mojolicious/Guides.pod
+++ b/lib/Mojolicious/Guides.pod
@@ -31,6 +31,9 @@ All web development starts with HTML, CSS and JavaScript, to learn the basics
 we recommend the
 L<Mozilla Developer Network|https://developer.mozilla.org/en-US/docs/Web>.
 
+=item Learning Hypertext Transfer Protocol
+http://en.wikipedia.org/wiki/Hypertext_Transfer_Protocol
+
 =back
 
 =head1 TUTORIAL


lib/Mojolicious/Guides.pod
```diff
@@ -31,6 +31,9 @@ All web development starts with HTML, CSS and JavaScript, to learn the basics
 we recommend the
 L<Mozilla Developer Network|https://developer.mozilla.org/en-US/docs/Web>.
 
+=item Learning Hypertext Transfer Protocol
```
kraih: There's a blank line missing here.

lib/Mojolicious/Guides.pod
```diff
@@ -31,6 +31,9 @@ All web development starts with HTML, CSS and JavaScript, to learn the basics
...
+http://en.wikipedia.org/wiki/Hypertext_Transfer_Protocol
```
kraih: This is not a POD link, and it's missing some context.

kraih:
I'm not opposed to the idea, but if we link to a learning resource for HTTP, it should actually be something good. And in fact, the Mozilla Developer Network we already link to has [something](https://developer.mozilla.org/en-US/docs/Web/HTTP). So :+1: for the idea, and :-1: for the implementation from me.

Grinnz:
More complete suggestion?

    =item Learning HTTP
    
    Browsers interact with web servers through HTTP, so it is helpful to have a
    basic understanding of the protocol. We recommend the documentation at the
    L<Mozilla Developer Network|https://developer.mozilla.org/en-US/docs/Web/HTTP>.

kraih:
@Grinnz Can't say i'm a huge fan of using the same link description twice.

Grinnz:
I agree. But that is where it is from, and it doesn't seem to be easy to find from the previous link. Perhaps:

    We recommend the L<MDN HTTP documentation|https://developer.mozilla.org/en-US/docs/Web/HTTP>.

kraih:
Closing this issue, since it looks like the discussion has ended and the pull request has not been updated recently.

PR #737, opened 2015-01-31
Use highlight.js instead of prettify for syntax coloring

dotandimet:
Note that the highlight.js Perl definition seems to tag less tokens than prettify, so I couldn't reproduce the current colors in full.


kraih:
Thanks, the results look pretty good mostly, i'm just a little worried about the `<script>` block of our frontpage example.
w7ag8h9

kraih:
This one is from the tutorial, so less critical, but would still be nice if the highlighting worked everywhere.
re

kraih:
It might also be worth looking into [prism.js](http://prismjs.com/), i hear good things about it.

dotandimet:
I just saw it mentioned in the Hacker News thread today, thinking about it.

marcusramberg:
Yeah, I think based on https://github.com/isagalaev/highlight.js/pull/722 we should probably look for something else. Did you test prism, @dotandimet ?

kraih:
I'm with @marcusramberg on that.

kraih:
Closing this issue, since its "work in progress" status has caused some confusion.

PR #741, opened 2015-02-17
Fix regression - some web frameworks terminate cookies with semicolons

dmw397:
For example, given an HTTP response containing:

Set-Cookie: a=b; expires=Wednesday, 17-Feb-2016 23:59:59 GMT;
Set-Cookie: c=d; expires=Wednesday, 17-Feb-2016 23:59:59 GMT;

Mojo::Message::Response::cookies in 5.77 returns both cookies but 5.79 only returns the first one.


kraih:
Thanks, applied!

PR #743, opened 2015-02-19
Test::Harness enables global warnings by default. Disable this behavior

olegwtf:
See also [this](https://github.com/Perl-Toolchain-Gang/Test-Harness/commit/18794a8565633ae08f048b9cf6c25b03be6e1020) and [this](https://rt.cpan.org/Public/Bug/Display.html?id=58229) discussions


kraih:
Thanks, applied.

PR #746, opened 2015-02-24
Updated format detection FAQ

Grinnz:


lib/Mojolicious/Guides/FAQ.pod
```diff
@@ -136,6 +136,18 @@ connections concurrently.
 
 In L<Mojolicious> this event loop is L<Mojo::IOLoop>.
 
+=head2 Why is the file extension missing from my placeholder value?
+
+Standard route placeholders will not match a C<.> character, however
```
jberger: The Routing guide calls these "Generic placeholders"

jberger: Or else rename them in the Routing guide. "Generic" might be misleading as it could be thought of as the most general, ie. the one that matches the most.

jberger:
Merged, thanks.

PR #748, opened 2015-02-28
Ignore expression's comment in template.

karupanerura:
We need to write comments with perl expression in template.

e.g.)
```perl
#!/usr/bin/env perl
use Mojolicious::Lite;

my @entries = ();

get '/' => sub {
    my $c = shift;
    $c->stash(entries => \@entries);
    $c->render(template => 'index');
};

post '/post' => sub {
    my $c = shift;
    my $entry = $c->param('body');
    push @entries => $entry;
    $c->stash(body => $entry);
    $c->redirect_to('/');
};

app->start;
__DATA__

@@ index.html.ep
% layout 'default';
% title 'note';
%= form_for '/post' => method => 'POST' => begin # comment...
  %= text_field 'body'
  %= submit_button 'POST'
% end

@@ layouts/default.html.ep
<!DOCTYPE html>
<html>
  <head><title><%= title %></title></head>
  <body><%= content %></body>
</html>
```

But, This application is failure.
And, I got error:
```
syntax error at template index.html.ep from DATA section line 6, near "} $_O "
Global symbol "$_O" requires explicit package name at template index.html.ep from DATA section line 6, <DATA> line 36.
Unmatched right curly bracket at template index.html.ep from DATA section line 6, at end of line
```

I fixed it.

### DIFF
--- a/lib/Mojo/Template.pm
+++ b/lib/Mojo/Template.pm
@@ -133,17 +133,17 @@ sub parse {
     = qr/^(\s*)\Q$start\E(?:(\Q$replace\E)|(\Q$cmnt\E)|(\Q$expr\E))?(.*)$/;
   my $token_re = qr/
     (
-      \Q$tag\E(?:\Q$replace\E|\Q$cmnt\E)                   # Replace
+      \Q$tag\E(?:\Q$replace\E|\Q$cmnt\E)                         # Replace
     |
-      \Q$tag$expr\E(?:\Q$escp\E)?(?:\s*\Q$cpen\E(?!\w))?   # Expression
+      \Q$tag$expr\E(?:\Q$escp\E)?(?:\s*\Q$cpen\E(?!\w))?         # Expression
     |
-      \Q$tag\E(?:\s*\Q$cpen\E(?!\w))?                      # Code
+      \Q$tag\E(?:\s*\Q$cpen\E(?!\w))?                            # Code
     |
-      (?:(?<!\w)\Q$cpst\E\s*)?(?:\Q$trim\E)?\Q$end\E       # End
+      (?:(?<!\w)\Q$cpst\E\s*)?(?:\#.*?)?(?:\Q$trim\E)?\Q$end\E   # End
     )
   /x;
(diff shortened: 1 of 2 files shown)


karupanerura:
oops... please wait a moment... I'll fix to code.

karupanerura:
Umm... It's difficult to solve by me.

In this case, This patch break valid Perl expressions.
```
<% '#000000' %>
```

Other solutions:

1. Insert a line break after the expression.
  * Pros: Very simple.
  * Cons: Breaks many test.
2. Tokenize Perl expressions, and exclude token of comment, and rebuild the Perl expressions.
  * Pros: Can impl the correct behavior.
  * Cons: (Maybe) has a performance issue..
  * Cons: Difficult to maintenance.
3. Regexp follow the quote literal.
  * Pros: Nothing...
  * Cons: Difficult to maintenance.

IMO, `1` is best solution for this issue. (and for this project.)
But, I don't have a deep knowledge about `Mojo::Template`.
So I notice that this impl is difficult for me.
When I came up with a good solution, I'll pull-req again.
Thank you.

jberger:
I recently saw a strange failure when C<;> followed a begin command. I tried to investigate a similar solution and it turns out it is very hard. In actuality it is very important to realize that the expression forms are simply that, expressions. Comments (and even C<;>) are not expressions.

kraih:
Yes, the real tricky case is an expression line with `begin` at the end, followed by `;` or `#`, like `%= foo bar => begin   # something`. Because you're essentially dealing with `<%= foo bar => begin   # something %>`, which is not just `TAG PERL TAG`, but `TAG PERL TAG PERL TAG`, and therefore cannot be tokenized easily. I've also added a few more tests to make the problem more obvious. https://github.com/kraih/mojo/commit/ec40a509757784e08fcfccc620c160428f51e52f

kraih:
For the record, inline comments like your original test case `<%= "one" # evil =%><%= "two" ## evil ## %>  three` were never supposed to work.

karupanerura:
OK. I also agree with the idea.
Thank you for reviewing.

kraih:
The case of the `begin` keyword at the end of an expression line really bugs me, so i've expanded the documentation a bit to cover it. https://github.com/kraih/mojo/commit/4cd34c5a18536efe0a9348d377c5aadfb6a5f523

PR #753, opened 2015-03-03
Removed warning when using undef values in tablify

avkhozov:
Usefull with Mojo::Pg::Results::text and null values.


kraih:
Thanks, applied.

PR #754, opened 2015-03-03
Add form post example

marcusramberg:
I think this is a common case that should have an example directly in ojo.

### DIFF
--- a/lib/ojo.pm
+++ b/lib/ojo.pm
@@ -170,6 +170,7 @@ resulting L<Mojo::Message::Response> object.
 
   my $res = p('example.com');
   my $res = p('http://example.com' => {Accept => '*/*'} => 'Hi!');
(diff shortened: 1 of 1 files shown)


lib/ojo.pm
```diff
@@ -170,6 +170,7 @@ resulting L<Mojo::Message::Response> object.
 
   my $res = p('example.com');
   my $res = p('http://example.com' => {Accept => '*/*'} => 'Hi!');
+  my $res = p('http://example.com', form => { foo => 'bar' });
```
kraih: That comma should be a fat comma.

kraih:
Actually, wouldn't it make more sense to have consistent examples with `Mojo::UserAgent`?

marcusramberg:
Hmm, I think maybe it would. Not sure if it would bloat the docs too much, but it would probably be helpful.

PR #758, opened 2015-03-11
use true/false functions instead of variables to allow monkey-patched Mojo::JSON

Grinnz:
Fixes _decode_value() to use the true and false functions in case they have been monkey patched to return different boolean objects, e.g. by Mojo::JSON::MaybeXS.

### DIFF
--- a/lib/Mojo/JSON.pm
+++ b/lib/Mojo/JSON.pm
@@ -198,10 +198,10 @@ sub _decode_value {
     if /\G([-]?(?:0|[1-9][0-9]*)(?:\.[0-9]*)?(?:[eE][+-]?[0-9]+)?)/gc;
 
   # True
-  return $TRUE if /\Gtrue/gc;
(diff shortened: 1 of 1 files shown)


kraih:
What's the performance cost?

kraih:
There is a cost, but it's not that bad.

Grinnz:
Any cost is certainly mitigated by being able to use XS-based JSON encoders anyway ;)

kraih:
And reverted, since modules like `Mojo::JSON::MaybeXS` should always be monkey patching all functions exported by `Mojo::JSON`. https://github.com/kraih/mojo/commit/997873fb9f679dfc22ea88dbf55f098b7c4f7bff

kberov:
The other day I was considering to monkey patch Mojo::JSON. I think it would be very beneficial to make Mojo::JSON to use internally JSON::XS if installed?..

PR #760, opened 2015-03-12
Note about need to remove connection from the loop when you don't need it anymore

olegwtf:
This was non-obvious for me and I got this memory leak in my application.

### DIFF
--- a/lib/Mojo/IOLoop.pm
+++ b/lib/Mojo/IOLoop.pm
@@ -402,10 +402,13 @@ Open TCP connection with L<Mojo::IOLoop::Client>, takes the same arguments as
 L<Mojo::IOLoop::Client/"connect">.
 
   # Connect to 127.0.0.1 on port 3000
-  Mojo::IOLoop->client({port => 3000} => sub {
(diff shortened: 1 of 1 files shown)


kraih:
:-1: This is not quite correct, a connection that times out or gets closed is automatically removed.

olegwtf:
Ah, I see. Then may be add a note that loop will keep reference for a stream and it is not enough to forget received `$stream` to destroy it?

olegwtf:
And finally I found cause of memory leak. I used unsubscribe on the stream: https://github.com/olegwtf/p5-Mojo-SMTP-Client/blob/d9715e9b0d39abb90a7d059881263c66a3d76aa0/lib/Mojo/SMTP/Client.pm#L282..L284 But `Mojo::IOLoop` setups non-documented event handlers which should remove stream from the loop when it will be closed. So, I still think this things should be cleared somehow in the documentation.

kraih:
Keep i mind that `Mojo::IOLoop` is not an end user API, the documentation is intentionally vague so we can make changes for performance tuning and the like.

PR #766, opened 2015-03-19
fix tag description

stfn:


kraih:
I'm afraid this is incorrect.

stfn:
I mixed up Mojo::Template and Mojo Mojolicious::Guides::Rendering. Sorry
Shouldn't `<%= %>` provide escaped results as mentioned in Mojolicious::Guides::Rendering?

`<%= Perl expression, replaced with XML escaped result %>`
`<%== Perl expression, replaced with result %>`

> An additional equal sign can be used to disable escaping of the characters "<", ">", "&", "'" and """ in results from Perl expressions, which is the default to prevent XSS attacks against your application.

jberger:
@stfn Mojolicious::Plugin::EPRenderer provides the `ep` template handling for Mojolicious apps. It in turn wraps Mojo::Template and makes the default more secure. Mojo::Template is more general, and the default escaping might not always be preferred.

jberger:
Since this is not a bug, if further discussion is desired, please move it to the mailing list or irc. Thanks

PR #773, opened 2015-03-29
Prevent "Subroutine redefined" warnings

dex4er:
```
$ find lib -name '*.pm' | xargs -n1 perl -Ilib -c
lib/Mojo/Reactor/Poll.pm syntax OK
Subroutine acceptor redefined at lib/Mojo/IOLoop.pm line 32.
Subroutine client redefined at lib/Mojo/IOLoop.pm line 49.
Subroutine delay redefined at lib/Mojo/IOLoop.pm line 71.
```

Mojo shows more than 100 warnings about redefined subroutines. It makes more difficult to develop Mojo in IDE like Padre or Eclipse+EPIC.

### DIFF
--- a/lib/Mojo/ByteStream.pm
+++ b/lib/Mojo/ByteStream.pm
@@ -3,7 +3,7 @@ use Mojo::Base -strict;
 use overload bool => sub {1}, '""' => sub { ${$_[0]} }, fallback => 1;
 
 use Exporter 'import';
-use Mojo::Collection;
+require Mojo::Collection;
 use Mojo::Util;
 
 our @EXPORT_OK = ('b');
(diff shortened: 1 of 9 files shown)


dex4er:
It might be a problem with perl itself: -c option doesn't set %INC so then Perl tries to redefine symbols when there are cycled `use`s.

I discovered that Padre 1.00 overrides this issue (probably sets %INC by itself) and doesn't show so many warnings. Eclipse still uses "perl -c" syntax so still floods with these warnings :(

Another workaround might be:

```
package Mojo::Collection;
BEGIN { $INC{'Mojo/Collection.pm'} ||= __FILE__ }
```
instead `require` in place of `use` in some of packages.

kraih:
I'm afraid this pull request did not get the required votes. :-1:

PR #775, opened 2015-04-04
Added element_count_is method

zoffixznet:
This allows for testing for whether the page contains the expected number of elements, based on the provided selector. You can either test for the exact number, or more/less than the specified number.

### DIFF
--- a/lib/Test/Mojo.pm
+++ b/lib/Test/Mojo.pm
@@ -83,6 +83,13 @@ sub content_type_unlike {
 
 sub delete_ok { shift->_build_ok(DELETE => @_) }
 
+sub element_count_is {
+  my ($self, $selector, $wanted_count, $desc) = @_;
+  $desc ||= encode 'UTF-8', qq{element count for selector "$selector"};
+  my $count = $self->tx->res->dom->find($selector)->size;
+  return $self->_test('is', $count, $wanted_count, $desc);
+}
+
 sub element_exists {
   my ($self, $selector, $desc) = @_;
   $desc ||= encode 'UTF-8', qq{element for selector "$selector" exists};
@@ -617,6 +624,14 @@ Opposite of L</"content_type_like">.
 
 Perform a C<DELETE> request and check for transport errors, takes the same
 arguments as L<Mojo::UserAgent/"delete">, except for the callback.
+
+=head2 element_count_is
+
+  $t = $t->element_count_is('.product', 6, 'we have 6 elements');
+
+Check the count of elements specified by the selector. Second argument
+is the number of elements you expect to find. Third argument is optional,
+and specifies the description of the test.
 
 =head2 element_exists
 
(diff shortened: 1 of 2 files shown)


jberger:
Without commenting on the utility or implementation of this method, remember that "its not a feature without tests and documentation". I see documentation but not tests.

jberger:
Also, rather than parsing the test on your own and turning it into an `ok` test, why not use `cmp_ok`?

jhthorsen:
I don't like the name. ```element_count``` sounds like something that returns a number, instead of comparing. Maybe rather ```number_of_elements_is```. (Probably a bit too long though)

I'm no sure if i see why you would need ```cmp_ok```, instead of simply ```is```

Also, I'm still not sure if I'm positive to the addition. (I'm not negative, just haven't decided)

jberger:
`cmp_ok` lets you test `<=`, `==`, `>` or the like, and gives descriptive feedback on failure (rather than just pass/fail of `ok`)

zoffixznet:
How about `element_count_is` ?

jberger:
I would personally expect it to be something more like `selector_count_is`

jberger:
put it a different way, all of the `element_` tests are testing a specific single element. This test method would be testing the result of a query returning a collection, and indeed testing the collection not the element.

zoffixznet:
`selector_count_is` is very confusing to me. I'm certainly not counting selectors (CSS), but elements (HTML) my selector matches. The fact that I'm using CSS selectors to tell the test what to count is secondary to what the purpose of the test is.

I've updated the PR to include tests. Please check and offer suggestions.

jberger:
why parse out the `$operator` part of the test, why not just accept an `$operator` argument?

zoffixznet:
The `==`, `<`, and `>` operators cover all the use cases (so anything else `cmp_ok` accepts isn't needed). The `==` is a sensible default. Since test description is an optional argument, I don't see another good way to have an optional operator argument as well, other than simply parse it out.

jhthorsen:
I don't like ```selector_count_is```. I'm also unsure about using ```cmp_ok```, while having the test method end with "_is". In which case do you need the operator, which require ```cmp_ok```, instead of simply ```is```?

kraih:
I also don't like any of the proposed method names, perhaps something like `elements_exist`?

jhthorsen:
I think ```elements_exist``` sounds like "give this method multiple selectors, and only succeed if all exists". I rather like something like ```n_elements_exists", but "n_" is not very understandable.

kraih:
@jhthorsen Disagree, ` $t->elements_exist('.field-with-error', '>', 6)` looks fine to me.

kraih:
@jhthorsen But i guess you meant `$t->elements_exist('.field-with-error', '>6')`, which does indeed look too similar to another selector.

kraih:
There is a similar method in [Casper.js](http://casperjs.readthedocs.org/en/latest/modules/tester.html#assertelementcount), think i would be fine with the name `element_count_is` if there was no support for ranges.

zoffixznet:
Updated to remove ranges

jhthorsen:
:+1: from me now.

jberger:
I'm happy with the name `element_count_is`. While I was supportive of ranges via `cmp_ok` I think @kraih 's suggestion that there aren't many great cases for tests where an exact value isn't known but a range is. I'm now :+1: on implementation and leaning :+1: on inclusion.

kraih:
Thanks, applied. :+1:

PR #776, opened 2015-04-04
Loading of the controller class

dionys:
If default value of `$app->controller_class` has been changed then the error raises on the controller building: `Can't locate object method "new" via package "TestApp::Controller" (perhaps you forgot to load "TestApp::Controller"?)`. So, if we change `controller_class` to `TestApp::Controller` then we must add `use TestApp::Controller;` to the app package.

This patch brings the app to load the controller class before the app instantiate it. The test fails on current code.

### DIFF
--- a/lib/Mojolicious.pm
+++ b/lib/Mojolicious.pm
@@ -4,6 +4,7 @@ use Mojo::Base 'Mojo';
 # "Fry: Shut up and take my money!"
 use Carp ();
 use Mojo::Exception;
+use Mojo::Loader qw(load_class);
 use Mojo::Util;
 use Mojolicious::Commands;
 use Mojolicious::Controller;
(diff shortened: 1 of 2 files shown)


kraih:
:-1: from me, because this is a very rarely used feature and i don't think it is worth the performance cost (however small it might be).

jhthorsen:
I agree with @kraih.

:-1:

dionys:
Hm, you right @kraih. I see 2 ways here.

First:

    has _controller_class => sub { 'Mojolicious::Controller' };
    sub controller_class {
        # get or load and set _controller_class
        # fails on startup if class not found
    }

Second: document this behaviour.

jhthorsen:
@dionys: What's wrong with ```use Some::Other::Controller; app->controller_class("Some::Other::Controller");``` ?

dionys:
Nothing wrong. But:

```perl
$app->routes->namespaces(['App::Controller']);  # no use App::Controller::Foo;
$app->plugin('Foo');                            # no use Mojolicious::Plugin::Foo;
```

jhthorsen:
@dionys: It's very common in the perl world to load modules before using them. I'm still :-1: on what has come up in this discussion.

kraih:
I'm afraid this pull request did not get the required votes. :-1:

PR #778, opened 2015-04-05
Add ability to track name of the executable script by not resolving symlinks

cono:
Change Cwd::abs_path to File::Spec->rel2abs

### DIFF
--- a/lib/Mojo/Server.pm
+++ b/lib/Mojo/Server.pm
@@ -2,7 +2,7 @@ package Mojo::Server;
 use Mojo::Base 'Mojo::EventEmitter';
 
 use Carp 'croak';
-use Cwd 'abs_path';
+use File::Spec::Functions 'rel2abs';
 use Mojo::Loader 'load_class';
 use Mojo::Util 'md5_sum';
 use POSIX;
@@ -44,7 +44,7 @@ sub load_app {
 
   # Clean environment (reset FindBin defensively)
   {
-    local $0 = $path = abs_path $path;
+    local $0 = $path = rel2abs $path;
     require FindBin;
     FindBin->again;
     local $ENV{MOJO_APP_LOADER} = 1;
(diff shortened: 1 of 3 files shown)


cono:
Started conversation here:
https://groups.google.com/forum/#!topic/mojolicious/OQEH0Aisouo

Sebastian, would you mind to place label: work in progress
I'm willing to cover this change with some tests and add some words in documentation.
But its 2:47am over here :) So gonna do it a bit later.

kraih:
Please don't open unfinished pull requests.

cono:
Sorry, I thought its a common practice to make work visible earlier by labeling it "work in progress": #737

kraih:
For the record, i've actually tried this patch and it didn't pass all tests on OS X.

cono:
@kraih would you mind to post prove result?

kraih:
The 6.06 release actually used `rel2abs` instead of `abs_path` in some cases, and the bug reports have started coming in. https://github.com/kraih/mojo/issues/779

kraih:
A few more.
http://www.cpantesters.org/cpan/report/e9ce217a-6c9d-1014-bc1b-836f56a2bf19
http://www.cpantesters.org/cpan/report/cd867c2d-6bf6-1014-9b4e-4eb7ae217b3d
http://www.cpantesters.org/cpan/report/ada72587-6bf7-1014-95c5-84ab24fcbee8
http://www.cpantesters.org/cpan/report/c6dd4dd1-6bf3-1014-9398-3af5bc74cf01

cono:
@kraih my first intention was to fix daemon.t in a way:

+  rel2abs($path), 'right script name');

Which calls "canonpath" inside of it. But just removed it :(

As for the current issues, only daemon.t is failing and imho the issue as itself not critical.

p.s. its a pity that you are reverting it :(

kraih:
Failing installations for a large number of our users is a very critical issue!

kraih:
We have decided that this is not worth doing more experiments, and the change has therefore been reverted. https://github.com/kraih/mojo/commit/693a3cf2ba8288cca89a45781dc9b33406251bde

cono:
Got it.
Thanks for the try.

PR #781, opened 2015-04-09
pod_to_html helper now will remove spaces at the start of the line for code blocks

olegwtf:
For now Mojolicious removes this spaces only when it renders /perldoc. Will be good if `pod_to_html` helper will also remove this useless whitespaces.

See also https://rt.cpan.org/Public/Bug/Display.html?id=103389


kraih:
Thanks, applied.

jberger:
This has broken the ASCII art in the Growing Guide

PR #783, opened 2015-04-15
Added validation of morbo listen option.

heathtechnical:
Morbo doesn't validate the listen option, allowing this:

$ morbo examples/hello.pl -l foo
Use of uninitialized value in string eq at /usr/local/share/perl5/Mojo/Server/Daemon.pm line 160.
Use of uninitialized value in string eq at /usr/local/share/perl5/Mojo/Server/Daemon.pm line 184.
Server available at foo

Added basic scheme/port validation.


kraih:
Tests don't pass anymore.

PR #784, opened 2015-04-20
Fix IE bug with decoding json response without encoding.

kak-tus:
IE 7,8,9 has bug with decoding of json response without encoding.
See http://stackoverflow.com/questions/8481453/jquery-ajax-request-doesnt-success-with-json-on-ie

The default encoding of application/json is UTF-8.
http://www.ietf.org/rfc/rfc4627.txt

That means, we can specify encoding in response without any trouble.


kraih:
Tests don't pass anymore.

PR #785, opened 2015-04-20
Fix IE bug with decoding json response without encoding.

kak-tus:
IE 7,8,9 has bug with decoding of json response without encoding.
See http://stackoverflow.com/questions/8481453/jquery-ajax-request-doesnt-success-with-json-on-ie

The default encoding of application/json is UTF-8.
http://www.ietf.org/rfc/rfc4627.txt

That means, we can specify encoding in response without any trouble.

### DIFF
--- a/lib/Mojolicious/Types.pm
+++ b/lib/Mojolicious/Types.pm
@@ -15,7 +15,7 @@ has mapping => sub {
     jpeg     => ['image/jpeg'],
     jpg      => ['image/jpeg'],
     js       => ['application/javascript'],
-    json     => ['application/json'],
+    json     => ['application/json;charset=UTF-8'],
     mp3      => ['audio/mpeg'],
     mp4      => ['video/mp4'],
     ogg      => ['audio/ogg'],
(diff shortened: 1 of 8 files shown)


Grinnz:
+1, this is one of the the first things I set in most projects anyway.

jberger:
For what its worth, I'm happy that @kak-tus made the appropriate fixes for the tests and made a complete proposal. In isolation I would say sure, but I worry that this might need to be treated as a breaking change. My concern is for things like

    $t->header_is('Content Type', 'application/json')

Which will suddenly start failing for people.

kraih:
The list of content types in `Mojolicious::Types` has not been updated.

kak-tus:
@jberger Another variant is to do something like this for json renderer
```
$self->res->headers->content_encoding('utf-8');
```
May be it more safely.

kraih:
I'll give this a :+1:, since all commonly used REST APIs i've tested appear to be using `application/json;charset=UTF-8`.

jhthorsen:
:+1: I don't have any good reasons for my thumb up, other than serving json as anything else than UTF-8 isn't something I do.

PR #786, opened 2015-04-21
make url encoding compliant with rfc 3986 sec 2.2

rpcme:
The RFC Section is here http://tools.ietf.org/html/rfc3986#section-2.2

First paragraph of the section says any conflicting chars should be escaped:
If data for a URI component would conflict with a reserved character's purpose as a delimiter, **then the conflicting data must be percent-encoded before the URI is formed**.

In such case, the faulting delimiters were removed from the character class so the name and/or value of a given pair can be encoded according to section 2.2.

A test was changed to reflect this condition; it was the only test failing after making the change.  Interestingly, the change reflects the default character class for url_escape.  Further simplification is then probable.

### DIFF
--- a/t/mojo/parameters.t
+++ b/t/mojo/parameters.t
@@ -190,8 +190,8 @@ $params = Mojo::Parameters->new('!$\'()*,:@/foo?=!$\'()*,:@/?&bar=23');
 is $params->param('!$\'()*,:@/foo?'), '!$\'()*,:@/?', 'right value';
 is $params->param('bar'),             23,             'right value';
 is "$params",
-   '%21%24%27%28%29%2A%2C%3A%40%2Ffoo%3F=%21%24%27%28%29%2A%2C%3A%40%2F' .
-   '%3F&bar=23', 'right result';
+  '%21%24%27%28%29%2A%2C%3A%40%2Ffoo%3F=%21%24%27%28%29%2A%2C%3A%40%2F'
+  . '%3F&bar=23', 'right result';
 
 # No charset
 $params = Mojo::Parameters->new('%E5=%E4')->charset(undef);
--- a/t/mojo/parameters.t
+++ b/t/mojo/parameters.t
@@ -189,7 +189,9 @@ is "$params", '%25foo%25=%25', 'right result';
 $params = Mojo::Parameters->new('!$\'()*,:@/foo?=!$\'()*,:@/?&bar=23');
 is $params->param('!$\'()*,:@/foo?'), '!$\'()*,:@/?', 'right value';
 is $params->param('bar'),             23,             'right value';
-is "$params", '%21%24%27%28%29%2A%2C%3A%40%2Ffoo%3F=%21%24%27%28%29%2A%2C%3A%40%2F%3F&bar=23', 'right result';
+is "$params",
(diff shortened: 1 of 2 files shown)


t/mojo/parameters.t
```diff
@@ -189,7 +189,7 @@ is "$params", '%25foo%25=%25', 'right result';
 $params = Mojo::Parameters->new('!$\'()*,:@/foo?=!$\'()*,:@/?&bar=23');
 is $params->param('!$\'()*,:@/foo?'), '!$\'()*,:@/?', 'right value';
 is $params->param('bar'),             23,             'right value';
-is "$params", '!$\'()*,:@/foo?=!$\'()*,:@/?&bar=23', 'right result';
+is "$params", '%21%24%27%28%29%2A%2C%3A%40%2Ffoo%3F=%21%24%27%28%29%2A%2C%3A%40%2F%3F&bar=23', 'right result';
```
kraih: This line is longer than 80 characters.

rpcme:
by the way this issue came about by doing some google api work, where the colons in the post were not being encoded, which caused a 400

https://developers.google.com/identity/protocols/OAuth2ServiceAccount#authorizingrequests

t/mojo/parameters.t
rpcme: yeah ok I can fix that

rpcme:
okay i fixed the formatting.

Grinnz:
I made a test to see what jQuery get and post handlers do: http://grinnz.com:3000/
Source at http://paste.fedoraproject.org/214060/49494142/raw/
In chrome (with input string `A-Za-z0-9-._~!$\'()*,:@/?` it results in:

    GET:  A-Za-z0-9-._~!%24%5C%27()*%2C%3A%40%2F%3F
    POST: A-Za-z0-9-._~!%24%5C'()*%2C%3A%40%2F%3F

Grinnz:
Updated the test to include browser form submits. On chrome I get for the same input string:

    GET:  A-Za-z0-9-._~!$'()*,:@/?
    POST: A-Za-z0-9-._%7E%21%24%27%28%29*%2C%3A%40%2F%3F

New source: http://paste.fedoraproject.org/214093/55358142/raw/

kraih:
This appears to be a rather complicated issue, with browsers and specs contradicting each other, and so far i'm afraid it has not been sufficiently proven that this patch is the correct solution. The current escaping behavior has not been causing any trouble in the past, so i'll go with a conservative :-1: vote for now.

jhthorsen:
I really think this is very complicated as well. If i would make any change, it would be to be compatible with [encodeURIComponent](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent), but that's neither what is already implemented, nor what is suggested in this PR.

Grinnz:
encodeURIComponent seems to be what jquery is using, its description is consistent with my earlier jquery tests except ' is also encoded for GET.

Grinnz:
For the issue, I would agree with any change that would escape at least all of the characters escaped by encodeURIComponent. It would only break tests that are testing the encoded output string which end users should not need to do, these changes would not break compatibility with any URL decoder.

PR #787, opened 2015-04-26
Add 'push' method to Mojo::Collection

rpcme:
Easily add elements to an existing collection.

For me, the benefit is:

has foo => sub { c->new };

later...

$bar->foo->push('another');

I extended this class originally using 'add' but push is probably more perly.

### DIFF
--- a/lib/Mojo/Collection.pm
+++ b/lib/Mojo/Collection.pm
@@ -266,7 +266,7 @@ Construct a new array-based L<Mojo::Collection> object.
 
 =head2 push
 
-  $collection->push(6, 7, 8);
+  my $collection = $collection->push(6, 7, 8);
 
 Add elements to the existing L<Mojo::Collection> object.
 
--- a/lib/Mojo/Collection.pm
+++ b/lib/Mojo/Collection.pm
@@ -55,6 +55,12 @@ sub new {
   return bless [@_], ref $class || $class;
 }
 
+sub push {
+  my $self = shift;
+  push @$self, @_;
+  return $self;
+}
+
 sub reduce {
   my $self = shift;
   @_ = (@_, @$self);
@@ -257,6 +263,12 @@ passed to the callback and is also available as C<$_>.
   my $collection = Mojo::Collection->new(1, 2, 3);
 
 Construct a new array-based L<Mojo::Collection> object.
+
+=head2 push
+
+  $collection->push(6, 7, 8);
+
+Add elements to the existing L<Mojo::Collection> object.
 
 =head2 reduce
 
--- a/lib/Mojo/Collection.pm
+++ b/lib/Mojo/Collection.pm
@@ -101,6 +101,12 @@ sub uniq {
   return $_[0]->new(grep { !$seen{$_}++ } @{$_[0]});
 }
 
+sub unshift {
+  my $self = shift;
+  unshift @$self, @_;
+  return $self;
+}
+
 sub _flatten {
   map { _ref($_) ? _flatten(@$_) : $_ } @_;
 }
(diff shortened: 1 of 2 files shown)


jhthorsen:
I don't see the benefit over ```push @{$obj->col_attr}, @things```. I also think the PR is incomplete if any such method should reach Mojo::Collection: What I mean is that it should also contain pop, unshift, splice, ... and I think it's just too much.

:-1: from me.

Note: You can also use Mojo::Autobox or postfix dereferencing in perl 5.20.x:
```
use Mojo::Base -base;
use experimental 'postderef';
use Mojo::Collection;
has col => sub { Mojo::Collection->new('foo') };
push main->new->col->@*, 'bar';
```

rpcme:
But... The same argument could be made for join, grep, and reverse.  I think adding methods "as needed" is in the spirit of how Mojo::Collection currently operates.

lib/Mojo/Collection.pm
```diff
@@ -258,6 +264,12 @@ passed to the callback and is also available as C<$_>.
 
 Construct a new array-based L<Mojo::Collection> object.
 
+=head2 push
+
+  $collection->push(6, 7, 8);
```
kraih: The return value is missing.

kraih:
:-1: from me as well. Just adding a `push` method without `unshift` seems inconsistent to me. `Mojo::ByteStream` also doesn't have a `concat` method.

rpcme:
resolved unshift inconsistency

lib/Mojo/Collection.pm
```diff
@@ -258,6 +270,12 @@ passed to the callback and is also available as C<$_>.
 
 Construct a new array-based L<Mojo::Collection> object.
 
+=head2 push
+
+  my $collection = $collection->push(6, 7, 8);
```
kraih: It does not appear to create a new collection.

lib/Mojo/Collection.pm
```diff
@@ -331,6 +349,16 @@ Create a new collection without duplicate elements.
   # "foo bar baz"
   Mojo::Collection->new('foo', 'bar', 'bar', 'baz')->uniq->join(' ');
 
+=head2 unshift
+
+  my $collection = $collection->unshift(1, 2, 3);
```
kraih: Same here, not actually a new collection.

rpcme:
ugh I blindly added that based on your prior comment

push and unshift behavior is they modify the existing collection

which behavior do you want, modify existing or create new collection?  I would like modify existing, that is why the doc was originally that way.

rpcme:
it's actually method then probably confusion it's in the functions section

tempire:
:-1: due to not returning a new collection, which would be inconsistent with the rest of the module.

marcusramberg:
I think it doesn't make sense for push and unshift to return a new collection. I'm :+1: with the inconsistency resolved.

kraih:
There is still inconsistency with `Mojo::ByteStream`, but it appears that @rpcme is not interested in this functionality anymore, so i think we can just end the discussion.

jberger:
Sorry for jumping in so late, @rpcme if there was some demonstrable benefit to the core distribution I think this would get a more serious look. It took me some time to learn that about the mojolicious project, the core stays tight by not adding unnecessary features, even when they seem possibly useful. An alternative might be to create a subclass on CPAN?

rpcme:
jberger, thanks for your constructive comments.

The last week did show me that it's not in anyone's best interest for me to be involved, at least on this level.  I don't understand the limits and expectations of this project and I wasted a lot of time.  It is obvious I don't have the patience for this and it shows in a very bad way to the point of embarassment and led to things said that shouldn't have been said.

Also, I'm not interested in creating public modules that I see as functionality gap patching which can taint object model purity and understanding. Although it's consistent with how the Perl community behaves in general, it's not something I personally feel good about.

Thank you and everyone keep up the good work.

PR #788, opened 2015-04-26
Set all user groups of user in setuidgid

Grinnz:
I have inspected Apache 2.4.10 and nginx 1.6.3 and determined they both use the following behavior to set user and group of their worker processes.

Set user to specified user, and set groups to all groups of specified user, replacing the primary group with the specified group (default same name as user in nginx). Also add specified group as secondary group if not already listed.

This patch removes the ability to set the group without setting the user, or to set a user without a group change (defaulting to the user if unspecified), also matching the behavior of Apache and nginx.

The reason this is necessary is to allow the application to gain the secondary group permissions of the specified user, while still starting as root to bind to privileged ports like 80.

There are no tests or documentation included, but this feature currently has neither, and it would be impossible to test in an automated fashion. However, I have verified that this sets the groups correctly on Centos 6, in perl 5.10.1 and 5.20.2, using the following script (changing user/group/listen as needed).

    use Mojolicious::Lite;

    app->config(hypnotoad => {
        user => 'grinnz',
        group => 'nginx',
        listen => ['http://*:8081'],
    });

    get '/' => sub {
        my $c = shift;
        $c->render(json => {
                real_uid => $<,
                eff_uid => $>,
                real_gids => $(,
                eff_gids => $),
        });
    };

    app->start;

### DIFF
--- a/lib/Mojo/Server.pm
+++ b/lib/Mojo/Server.pm
@@ -79,22 +79,22 @@ sub setuidgid {
   $self->_error(qq{User "$user" does not exist})
     unless defined(my $uid = getpwnam $user);
 
-  # User groups
-  my @gids;
-  while (my (undef,undef,$id,$members) = getgrent()) {
-    push @gids, $id if grep { $_ eq $user } split ' ', $members;
-  }
-
   # Group
   my $group = $self->group // $user;
   $self->_error(qq{Group "$group" does not exist})
     unless defined(my $gid = getgrnam $group);
-  unshift @gids, $gid unless grep { $_ == $gid } @gids;
+
+  # User groups
+  my @gids = ($gid);
+  while (my (undef, undef, $id, $members) = getgrent()) {
+    next if $id == $gid;
+    push @gids, $id if grep { $_ eq $user } split ' ', $members;
+  }
 
   $( = $gid;
-  $self->_error(qq{Can't switch to group "$group": $!}) if $!;
+  $self->_error(qq{Can't switch to real group "$group": $!}) if $!;
   $) = join ' ', $gid, @gids;
-  $self->_error(qq{Can't switch to groups "$gid @gids": $!}) if $!;
+  $self->_error(qq{Can't switch to effective groups "$gid @gids": $!}) if $!;
   $self->_error(qq{Can't switch to user "$user": $!})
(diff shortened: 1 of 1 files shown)


Grinnz:
Security implications: Applications that set only the user configuration will be more secure with this change, as they will be required to change from the root groups. Security of setting both user and group will be as expected when changing user: gaining the privileges of the specified user and group only, but including the user's secondary group permissions.

Grinnz:
nginx implements this by calling initgroups(3), as seen here: https://github.com/nginx/nginx/blob/ae88e2338f6e27459ace8a23754afc5892c2c5be/src/os/unix/ngx_process_cycle.c#L864

Perl does not have the initgroups syscall, so the closest implementation is to find the groups of which the user is a member, and set those plus the specified group as the secondary groups.

kraih:
This is basically an untested security feature. So we need very clear code that has been reviewed very carefully by multiple parties. If it cannot be determined that the current solution is correct, i don't think it should stay, no matter the outcome of the vote. Personally, i'm still in favor of removing the feature completely and letting plugins handle it.

kraih:
@jberger @marcusramberg @tempire @jhthorsen @amenonsen What do you think?

Grinnz:
Apache also implements this functionality with initgroups, as on this page line 126:

http://svn.apache.org/viewvc/httpd/httpd/trunk/modules/arch/unix/mod_unixd.c?view=markup

On platforms that don't have initgroups, it reimplements them with the function on line 279 of this file:

http://svn.apache.org/viewvc/httpd/httpd/trunk/server/mpm_common.c?view=markup

It iterates through getgrent(), retrieves the groups which contain the user as a member (skipping the specified group), and sets those groups plus the specified group, which matches with the functionality in the patch.

kraih:
@Grinnz So, evidence points towards our current solution being incorrect.

jhthorsen:
:facepunch: Let's remove the whole thing. I've never used this myself. I rather use iptables to forward 80/443 to the server https://metacpan.org/pod/Toadfarm::Manual::RunningToadfarm#Listen-to-standard-HTTP-ports

Grinnz:
Closing with the release of https://metacpan.org/pod/Mojolicious::Plugin::SetUserGroup

PR #789, opened 2015-04-30
Added a test for to_string() caching

polettix:
This test exercises Mojo::Message::Response to check whether the `to_string()` method freezes the whole object and prevents further modifications, e.g. of the code. I consider this a bug (which may not be the case, of course).

The pull request itself is for the test only, so it will "break" the distribution in a TDD way. I don't have any idea of a patch.

### DIFF
--- a/t/mojo/request_to_string.t
+++ b/t/mojo/request_to_string.t
@@ -1,0 +1,27 @@
+use Mojo::Base -strict;
+
+use Test::More;
+use Mojo::Message::Response;
+
+my $res = Mojo::Message::Response->new;
+$res->code(204);
+is $res->code(), 204, 'setting the code';
+$res->body('whatever');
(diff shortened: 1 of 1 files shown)


jhthorsen:
Just out of curiosity: When do you need to rebuild the message?

polettix:
I was calling to_string in the middle of after_dispatch to print a log
while debugging and all of a sudden the headers I was adding after the call
to to_string disappeared from the response.
Il 30/apr/2015 16:06, "Jan Henning Thorsen" <notifications@github.com> ha
scritto:

kraih:
This is not a bug but a performance optimization.

jhthorsen:
@kraih: Could we clarify the behavior in the POD? (I'm not sure how the wording should be)

polettix:
I understand the view of the Owner, turning this into an Issue.

PR #790, opened 2015-05-03
Mojo::IOLoop->client with "proto" option and UDP support

dex4er:
Please, allow to use UDP protocol in ``Mojo::IOLoop::Client``. It makes possible to create ie. DNS client.

Proof of concept:

```perl
use Mojo::IOLoop;
use Net::DNS::Packet;

my $address = $ARGV[0] || 'mojolicio.us';

Mojo::IOLoop->client({address => '8.8.8.8', port => 53, proto => 'udp'} => sub {
    my ($loop, $err, $stream) = @_;
    $stream->on('read' => sub {
      my ($stream, $bytes) = @_;
      my $ans = Net::DNS::Packet->new(\$bytes, 1);
      $stream->close;
    });
    my $packet = Net::DNS::Packet->new($address);
    $stream->write($packet->data);
});

Mojo::IOLoop->start unless Mojo::IOLoop->is_running;
```

### DIFF
--- a/lib/Mojo/IOLoop/Client.pm
+++ b/lib/Mojo/IOLoop/Client.pm
@@ -83,6 +83,7 @@ sub _connect {
     $options{Proto} = $args->{proto};
     $options{Blocking} = 0;
     $options{LocalAddr} = $args->{local_address} if $args->{local_address};
+    $options{LocalPort} = $args->{local_port} if $args->{local_port};
     return $self->emit(error => "Can't connect: $@")
       unless $self->{handle} = $handle = IO::Socket::IP->new(%options);
   }
@@ -290,6 +291,12 @@ Use an already prepared handle.
   local_address => '127.0.0.1'
 
 Local address to bind to.
+
+=item local_port
+
+  local_address => 12345
+
+Local port to bind to.
 
 =item port
 
--- a/lib/Mojo/IOLoop/Client.pm
+++ b/lib/Mojo/IOLoop/Client.pm
@@ -41,6 +41,7 @@ sub connect {
     sub { $self->emit(error => 'Connect timeout') });
 
   # Blocking name resolution
+  $args->{proto} ||= 'tcp';
   $_ && s/[[\]]//g for @$args{qw(address socks_address)};
   my $address = $args->{socks_address} || ($args->{address} ||= '127.0.0.1');
   return $reactor->next_tick(sub { $self && $self->_connect($args) })
@@ -79,6 +80,7 @@ sub _connect {
   unless ($handle = $self->{handle} = $args->{handle}) {
     my %options = (PeerAddr => $address, PeerPort => _port($args));
     %options = (PeerAddrInfo => $args->{addr_info}) if $args->{addr_info};
+    $options{Proto} = $args->{proto};
     $options{Blocking} = 0;
     $options{LocalAddr} = $args->{local_address} if $args->{local_address};
     return $self->emit(error => "Can't connect: $@")
@@ -104,7 +106,8 @@ sub _ready {
   return $self->emit(error => $! || 'Not connected') unless $handle->connected;
 
   # Disable Nagle's algorithm
-  setsockopt $handle, IPPROTO_TCP, TCP_NODELAY, 1;
+  setsockopt $handle, IPPROTO_TCP, TCP_NODELAY, 1
+    if $args->{proto} eq 'tcp';
 
   $self->_try_socks($args);
 }
@@ -193,7 +196,7 @@ sub _try_tls {
 
 =head1 NAME
 
-Mojo::IOLoop::Client - Non-blocking TCP client
+Mojo::IOLoop::Client - Non-blocking TCP/UDP client
 
 =head1 SYNOPSIS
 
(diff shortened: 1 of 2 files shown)


jberger:
Since I'm no expert on the technical merits of this PR, I will say instead that this will have very little chance of inclusions without tests.

dex4er:
@jberger Ok, I'll hack ioloop_ipv6.t so it will test UDP protocol. Also I need to add UDP support for Mojo::IOLoop::Server.

jberger:
remember that I'm not in any way saying that doing so will guarantee
inclusion, simply that without tests it is nearly impossible.

dex4er:
I prepared `ioloop_client_udp.t` which tests client only because UDP server is de facto client with LocalAddr/LocalPort and without PeerAddr/listen/accept. It might be implemented by Mojo::IOLoop::Server but I'm afraid it is more changes than for ::Client.

jhthorsen:
I would be very surprised if this gets merged, since this it not core functionality.

:-1: from me, unless someone can convince me that this improves Mojolicious.

marcusramberg:
Can you expand what you mean by 'this is not core functionality'? "It can't be added to the core because it's not core functionality" sounds like circular reasoning to me. :confounded:  Do you mean 'since it's not needed for the core functionality'?

dex4er:
@jhthorsen It is difficult to move out this feature to separate MojoX module without copying a large portion of Mojo::IOLoop code and breaking further compatibility. See https://gist.github.com/dex4er/175471dd187a230b6077 which works with bare Mojolicious. Unfortunately this small example has a lot of boilerplate, overrides private methods of Mojo::IOLoop::Client and probably will break with further changes in Mojo API.

It could be a core functionality. This example uses Net::DNS and my next PoC would work with Net::DNS::Lite so it will be possible to use non-async pure-Perl DNS resolver in Mojo itself.

jberger:
While @jhthorsen is correct that typically we do not include code that isn't needed for Mojolicious core functionality, I think it is worth considering the fact that very little code is actually added for this PR; it simply exposes options that the underlying systems already provide. That leads me to tend :+1:.

jhthorsen:
@marcusramberg: Yes, "since it's not needed for the core functionality".

I also don't see how Mojo::IOLoop::Client can work with UDP: How does the "connect" method/event make sense in that context?

jberger:
@jhthorsen its that kind of thing that I can't comment on. I know 0 about UDP. I was assuming that @dex4er wouldn't be proposing this change if it didn't do anything.

harry-bix:
@jhthorsen: It is easy to write generic TCP clients and servers using only the Mojolicious core functionality.
The small changes in this PR enhance Mojolicious by making the same for generic UDP clients.

Mojolicious API may be incomplete if only client UDP functionality is available.

jberger:
To update my previous comment. I probably shouldn't vote because I know 0 about UDP. All I will say is that IF this functionality comes down to pass-through configuration, I don't think it should be ruled out because of issues of usefulness to the core. Note that that "IF" is very important though and that is why my knowledge gap fits.

dex4er:
@jhthorsen `strace` shows that `connect` syscall is used in this case:

```
socket(PF_INET, SOCK_DGRAM, IPPROTO_UDP) = 4
connect(4, {sa_family=AF_INET, sin_port=htons(54321), sin_addr=inet_addr("127.0.0.1")}, 16) = 0
poll([{fd=4, events=POLLIN|POLLPRI|POLLOUT}], 1, 0) = 1 ([{fd=4, revents=POLLOUT}])
write(4, "PING", 4)                     = 4
```

so `connect` event still has correct meaning. The UDP "server" has different semantic as far as don't use listen and accept syscalls (only bind). The UDP client is very similar to TCP client, at least for Linux system.

UDP support in Mojo will make possible to run pure-Perl DNS resolver. Also I think UPNP protocol runs on HTTPU protocol which is literally HTTP over UDP, so we'll gain new possibilities.

jhthorsen:
@dex4er: Can you give me a reference to what you're saying? From what I understand, connect makes little sense in UDP context.

http://man7.org/linux/man-pages/man7/udp.7.html

> When connect(2) is called on the socket, the default destination address
> is set and datagrams can now be sent using send(2) or write(2)
> without specifying a destination address.

dex4er:
@jhthorsen http://linux.die.net/man/2/connect

> If the socket sockfd is of type SOCK_DGRAM then addr is the address to which datagrams are sent by default, and the only address from which datagrams are received.
> Generally, connection-based protocol sockets may successfully connect() only once; connectionless protocol sockets may use connect() multiple times to change their association.

Well, I would define that connection is established when connect(2) function returns zero. Such definition makes sense for `connect` event for both TCP and UDP protocols.

jhthorsen:
@dex4er: That's interesting. Thank you.

I found some old conversations that might give some input on this discussion as well.
* http://irclog.perlgeek.de/mojo/2012-10-15#i_6064324
* http://irclog.perlgeek.de/mojo/2012-11-10#i_6142123
* http://irclog.perlgeek.de/mojo/2011-11-17#i_4720479

kraih:
:-1: I'm with @jhthorsen on this, the abstraction is all wrong, UDP is not a stream. And the addition of a new module like `Mojo::IOLoop::Datagram` would be hard to justify without any core usage.

kraih:

PR #791, opened 2015-05-05
Validation of Mojo::Server::Daemon listen parameter

heathtechnical:
Mojo::Server::Daemon currently appears to work with some rather strange listen strings:

    $ perl -I lib blib/script/morbo examples/hello.pl -l http://localhost:tproxy:3000
    Server available at http://localhost:tproxy:3000
    $ perl -I lib blib/script/morbo examples/hello.pl -l ftp://*:9000
    Server available at ftp://127.0.0.1:9000
    $ perl -I lib blib/script/morbo examples/hello.pl -l broken:9000
    Use of uninitialized value in string eq at lib/Mojo/Server/Daemon.pm line 160.
    Use of uninitialized value in string eq at lib/Mojo/Server/Daemon.pm line 184.
    Server available at broken:9000

To remedy, added basic listen string validation and more accurate printing of actual listen address/port.


marcusramberg:
This looks very nice to me. :+1:

lib/Mojo/Server/Daemon.pm
```diff
@@ -146,7 +147,17 @@ sub _finish {
 sub _listen {
   my ($self, $listen) = @_;
 
-  my $url     = Mojo::URL->new($listen);
+  my $url = Mojo::URL->new($listen);
+
+  croak(
+    "Scheme not specified or not one of http/https in listen location '$listen'"
+  ) if !$url->scheme || $url->scheme !~ /^https*$/;
```
jhthorsen:
```
croak("Scheme need to be http or https in listen location '$listen'") unless $url->scheme =~ /^https?$/;
```
Also note that "s*" == sssssss, while you probably don't want to match "httpssssss".

lib/Mojo/Server/Daemon.pm
```diff
@@ -146,7 +147,17 @@ sub _finish {
...
+
+  # Disallow symbolic port names
+  croak("Invalid address in listen location '$listen'")
+    if (!$url->host && !$url->port)
+    || $url->host =~ /:(?![^\[]*\])/;
```
jhthorsen: Can you give examples in the unit test on what this will and won't match?

lib/Mojo/Server/Daemon.pm
```diff
@@ -179,9 +190,12 @@ sub _listen {
   );
 
   return if $self->silent;
+  my $aport
+    = $self->ioloop->acceptor($self->acceptors->[$#{$self->acceptors}])->port;
+  $url->port($aport) if !$url->port;
```
jhthorsen: Why do you set port()?

lib/Mojo/Server/Daemon.pm
```diff
@@ -179,9 +190,12 @@ sub _listen {
...
+  $url->host("*") if !$url->host;
   $self->app->log->info(qq{Listening at "$url"});
   $query->pairs([]);
-  $url->host('127.0.0.1') if $url->host eq '*';
```
kraih: Why do you want beginners to see `Server available at http://*:3000` instead of `Server available at http://127.0.0.1:3000` when they start their morbo development server for the first time?

lib/Mojo/Server/Daemon.pm
```diff
@@ -179,9 +190,12 @@ sub _listen {
   );
 
   return if $self->silent;
+  my $aport
+    = $self->ioloop->acceptor($self->acceptors->[$#{$self->acceptors}])->port;
```
kraih: Why `[$#{$self->acceptors}]` instead of `[-1]`?

kraih:
Have to give this a :-1: for now i'm afraid, to balance out the vote from @marcusramberg. Most of the checks seem a little sketchy and not all are covered by tests.

jhthorsen:
I totally agree that the test need to be improved.

I do believe I'm +1 on the idea (since it's sometimes difficult to spot why the server fail to listen), but I'm :-1: on this PR as it stands now.

heathtechnical:
Thanks for the feedback.

Have improved checks and implemented more complete tests.

Let me know what you think.

lib/Mojo/Server/Daemon.pm
heathtechnical: Might it not be better to change the default listen location to http://127.0.0.1:3000 rather than faking the host like this?

jhthorsen: It's not for the default listen location, it's only used as information to screen.

marcusramberg: from a security perspective listening to localhost by default seems safer, but I guess it's also a breaking change that will surprise a lot of people.

heathtechnical: Yep agreed.  I realise this is only informational, but I would rather be told it's binding to all interfaces if that is what it really is doing.

kraih:

PR #796, opened 2015-05-09
determine the verbatim pod indent from least indented line

### LINKED ISSUE #795: ASCII art broken in pod
#781 removes leading whitespace in pod verbatim sections. This breaks the ASCII art in pages like the [Growing Guide](http://mojolicio.us/perldoc/Mojolicious/Guides/Growing#Model-View-Controller). Interestingly it doesn't seem to break all of them, like the end of this [one](http://mojolicio.us/perldoc/Mojolicious/Guides/Growing#Differences). 

Not sure if we should revert the patch or harden the ASCII art. My initial reaction is that for a general purpose pod renderer we should leave leading whitespace as-is for purposes such as these.

jberger:
This addesses ascii art broken by #781

fixes #795

### DIFF
--- a/lib/Mojolicious/Plugin/PODRenderer.pm
+++ b/lib/Mojolicious/Plugin/PODRenderer.pm
@@ -3,7 +3,6 @@ use Mojo::Base 'Mojolicious::Plugin';
 
 use Mojo::Asset::File;
 use Mojo::ByteStream 'b';
-use Mojo::Collection 'c';
 use Mojo::DOM;
 use Mojo::URL;
 use Mojo::Util 'slurp';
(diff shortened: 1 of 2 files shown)


jberger:
for the record, I would be as happy to unapply #781 as apply this patch

jberger:
A variant that only computes `length` once, but it rather more ugly: http://pastie.org/10178896

lib/Mojolicious/Plugin/PODRenderer.pm
```diff
@@ -96,7 +96,7 @@ sub _pod_to_html {
   my $parser = Pod::Simple::XHTML->new;
   $parser->perldoc_url_prefix('https://metacpan.org/pod/');
   $parser->$_('') for qw(html_header html_footer);
-  $parser->strip_verbatim_indent(sub { my ($i) = $_[0][0] =~ /^(\s+)/; $i });
+  $parser->strip_verbatim_indent(sub { (sort map /^(\s+)/, @{$_[0]})[0] });
```
kraih: Is `map` without a block good style now?

jberger: no, but without it perltidy kills it and makes it 4 lines

jberger: that was actually my complaint here: http://irclog.perlgeek.de/mojo/2015-05-09#i_10574016

jberger:
Note that @theory has actually adopted this logic in one of his modules. https://github.com/theory/text-markup/commit/6a8953d366389495ab519b2cc7bc142979ff8e08

Issue #792, opened 2015-05-05
to_string() freezes the state in Mojo::Message::Response


polettix:
Please consider documenting that "to_string()" freezes the object, and possibly include hints on how to get the string representation without being hit by the optimization.

This issue is a consequence of the discussion in #789.

tempire:
This is unlikely to happen in the core.

You are, of course, welcome to create a serialization plugin that fits your needs. You could handle this in the after_render hook.

polettix:
I'm only suggesting to document the current behavior, not to change it - this is why I closed PR #789 and opened this issue.

You can of course think that "to_string()" name and documentation are clear enough as to its semantics and effect on the object, and disregard this Issue. Just take into account that other people might to be biten in the future and you probably want to avoid bothering!

tempire:
Ah, you're right. I missed that part.

This is reasonable, considering several people have been tripped up on this issue.

kraih:
Closing this issue, since no actual proposals have been made yet for how this should be documented.

polettix:
I would suggest reconsidering this decision and label is issue at least as "future". An "excellent documentation" shouldn't leave space to undocumented behaviour like this.

I can try again with a (new) PR, although this trading performance for behaviour might happen somewhere else or in a wider context than the one I tried to pinpoint. In case, it would be useful to know that as well and this probably calls for a core developer to chime in.

Before wasting anyone's time, though, it would be interesting to know if this is of any interest. If you don't bother, probably there's nothing to bother about.

jberger:
I think that that is indeed what @kraih is asking for. Keep it brief, I
imagine it can be a sentence or less.

kraih:
@polettix Considering that nobody else has commented in over a week (aside from a misunderstanding), it may very well be that there is no interest.

kraih:
Perhaps i should mention that this is actually a lot more complicated than it may look, some `Mojo::Message::Request`/`Mojo::Message::Response` objects have dynamically generated content and can only be rendered once.

polettix:
@jberger I didn't see requests by @kraih. @tempire asked for some wording, but given how this goes into the inner working of the objects my Mojolicious-fu is too weak.
@kraih some people seemed to acknowledge that docs should be fixed. It surprises me that an issue like this can be closed without any change (considering we're talking about docs), maybe I'm just easy at being surprised.

I'll try to switch to "interview mode" and then propose a PR based on the answers:
[...]

PR #797, opened 2015-05-15
expand on to_string() documentation

polettix:
As anticipated on #792, a (minimal(istic)) proposal for a better documentation of to_string()

### DIFF
--- a/lib/Mojo/DOM.pm
+++ b/lib/Mojo/DOM.pm
@@ -73,7 +73,7 @@ sub find { $_[0]->_collect(@{$_[0]->_css->select($_[1])}) }
 sub following { _select($_[0]->_collect(@{$_[0]->_siblings(1)->[1]}), $_[1]) }
 sub following_nodes { $_[0]->_collect(@{$_[0]->_siblings->[1]}) }
 
-sub matches { $_[0]->_css->matches($_[1]) }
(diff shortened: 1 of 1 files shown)


lib/Mojo/Message.pm
```diff
@@ -630,7 +630,10 @@ L</"default_charset">.
 
   my $str = $msg->to_string;
 
-Render whole message.
+Render whole message. For performance optimization reasons, calling this
+method freezes the message representation, so that subsequent calls
+to other methods (e.g. to modify the headers) will have no effect on the
+message rendering.
```
kraih: This is incorrect, only certain parts of the message might get frozen, and we make no guarantees which. (It's an optimization after all, and we need to be able to keep changing it!)

kraih:
:-1: from me, because the description is incorrect.

tempire:
It might be a better idea to give a warning that to_string might freeze parts of the message, instead of documenting exactly what it does, so that the dev is prompted to do further research if they need it, or find another method to do what they want.

It would be a shame if optimizations change and the docs are forgotten or out of date for a period of time.

kraih:
And don't forget that it might not even be possible to render a message twice, we really should avoid making it look like using `to_string` is a good idea.

PR #798, opened 2015-05-16
to_string() might give surprises

polettix:
Documentation proposal related to #792.

### DIFF
--- a/lib/Mojo/Message.pm
+++ b/lib/Mojo/Message.pm
@@ -630,7 +630,9 @@ L</"default_charset">.
 
   my $str = $msg->to_string;
 
-Render whole message.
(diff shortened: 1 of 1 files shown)


kraih:
:-1: from me again, since the description still gives the impression that calling `to_string` is a good idea, even though it might destroy the content.

kraih:
I've added a different warning. https://github.com/kraih/mojo/commit/a66401d4cdf59ab218bbabedced11a3cd09894d8

polettix:
What's the difference with respect to a private method?
Il 16/mag/2015 15:02, "Sebastian Riedel" <notifications@github.com> ha
scritto:

PR #802, opened 2015-05-29
Minor grammar fixes for Guides

lindleyw:
Result of late night re-reading with blue pencil.

### DIFF
--- a/lib/Mojolicious/Guides/Cookbook.pod
+++ b/lib/Mojolicious/Guides/Cookbook.pod
@@ -212,7 +212,7 @@ C<mod_proxy>, the configuration looks quite similar to the Nginx one above.
 =head2 Apache/CGI
 
 C<CGI> is supported out of the box and your L<Mojolicious> application will
-automatically detect that it is executed as a C<CGI> script. It's use in
+automatically detect that it is executed as a C<CGI> script. Its use in
 production environments is discouraged though, because as a result of how
 C<CGI> works, it is very slow and many web servers are making it exceptionally
 hard to configure properly.
@@ -452,7 +452,7 @@ nested closures that often result from continuation-passing style.
 
 =head2 Timers
 
-Another primary feature of the event loop are timers, which are created with
+Another primary feature of the event loop is timers, which are created with
 L<Mojo::IOLoop/"timer"> and can for example be used to delay rendering of a
 response, and unlike C<sleep>, won't block any other requests that might be
 processed concurrently.
--- a/lib/Mojolicious/Guides/Rendering.pod
+++ b/lib/Mojolicious/Guides/Rendering.pod
@@ -443,8 +443,8 @@ L<Mojolicious::Plugin::DefaultHelpers/"accepts">.
 
 By now you've probably already encountered the built-in 404 (Not Found) and 500
 (Server Error) pages, that get rendered automatically when you make a mistake.
-Those are fallbacks for when your own exception handling fails, but especially
-during development they can also be a great help, you can render them manually
+Those are fallbacks for when your own exception handling fails. Especially
+during development, they can also be a great help: you can render them manually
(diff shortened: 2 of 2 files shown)


lib/Mojolicious/Guides/Cookbook.pod
```diff
@@ -212,7 +212,7 @@ C<mod_proxy>, the configuration looks quite similar to the Nginx one above.
 =head2 Apache/CGI
 
 C<CGI> is supported out of the box and your L<Mojolicious> application will
-automatically detect that it is executed as a C<CGI> script. It's use in
+automatically detect that it is executed as a C<CGI> script. Its use in
```
jberger: agree

lib/Mojolicious/Guides/Cookbook.pod
```diff
@@ -452,7 +452,7 @@ nested closures that often result from continuation-passing style.
 
 =head2 Timers
 
-Another primary feature of the event loop are timers, which are created with
+Another primary feature of the event loop is timers, which are created with
```
jberger: might be convinced

lib/Mojolicious/Guides/Rendering.pod
```diff
@@ -1128,7 +1128,7 @@ applications, plugins make that very simple.
 
   1;
 
-The C<register> method will be called when you load the plugin and to add your
+The C<register> method will be called when you load the plugin; to add your
```
jberger: disagree

lib/Mojolicious/Guides/Rendering.pod
```diff
@@ -443,8 +443,8 @@ L<Mojolicious::Plugin::DefaultHelpers/"accepts">.
 
 By now you've probably already encountered the built-in 404 (Not Found) and 500
 (Server Error) pages, that get rendered automatically when you make a mistake.
-Those are fallbacks for when your own exception handling fails, but especially
-during development they can also be a great help, you can render them manually
+Those are fallbacks for when your own exception handling fails. Especially
+during development, they can also be a great help: you can render them manually
```
jberger: Propose a third alternative

```
Those are fallbacks for when your own exception handling fails, which can be especially
helpful during development. You can also render them manually ...
```

that change would require rebalancing the lines of course

lib/Mojolicious/Guides/Rendering.pod
Grinnz: I would agree with the change, these are two separate statements.

lib/Mojolicious/Guides/Rendering.pod
Grinnz: +1 to jberger's suggestion

lib/Mojolicious/Guides/Cookbook.pod
Grinnz: This is a tricky one, "timers" is a single feature but "timers" is plural. Perhaps better to use timers as the subject:
    Timers, another primary feature of the event loop, are created with ...

lib/Mojolicious/Guides/Rendering.pod
jberger: oh, crap, I actually agree, I seem to have gone red/green colorblind on this one. to be clear, I prefer the semicolon. sorry for the confusion

lib/Mojolicious/Guides/Cookbook.pod
kraih: What @Grinnz said. :+1:

lib/Mojolicious/Guides/Rendering.pod
kraih: I like @jberger 's proposal too.

lib/Mojolicious/Guides/Rendering.pod
kraih: Perhaps two sentences?

PR #803, opened 2015-06-01
Using names for app and controller same as in documentation

avkhozov:


kraih:
No, in this context `$self` is correct.

avkhozov:
This names used only in lite app?

Grinnz:
These methods are within the app or controller object respectively thus the $app or $c object is $self in those methods.

PR #804, opened 2015-06-03
Tell GitHub our test files are Perl 5, not Perl 6

zoffixznet:
This fixes incorrect reporting by GitHub that our repo's test files are written in Perl 6:
perl6

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+*.t linguist-language=Perl
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 .*
 *~
+!.gitattributes
 !.gitignore
 !.perltidyrc
 !.travis.yml


kraih:
Not sure about this. I feel like this is really GitHub's problem and they should just be doing a better job with detecting languages.

kraih:
Has this actually been reported to GitHub as a bug yet?

zoffixznet:
I see two closed tickets [https://github.com/github/linguist/issues/2074])(2074) and [https://github.com/github/linguist/issues/2149](2149) that cover the issue.

Since `.t` can mean both Perl5 or Perl6, Linguist does some heuristics to figure out which language it is, and since the two languages are similar, in this case it's failing.

Adding a `perl` shebang will fix it, adding `use strict;` *might* fix it, and another solution provided in the (https://github.com/github/linguist/issues/2149#issuecomment-75452121)[answer to why the issue likely won't be fixed] is to use `.gitattributes`, which is what this PR has done.

kraih:
I've commented on one of the issues. https://github.com/github/linguist/issues/2149#issuecomment-108954064

kraih:
Clear :-1: from me. This is a problem that needs to be solved in Linguist (not just for us, but for the entire Perl community). Also, the workaround doesn't appear to work for syntax highlighting either.

zoffixznet:
You know what, the discussion here, on IRC, and reading the Linguist tickets actually changed my mind about this PR as well. Since this isn't the first time I'm forced to add `.gitattributes` ([here](https://github.com/scrottie/SOAP-WSDL/pull/4), [here](https://github.com/zoffixznet/Imager-Tiler/commit/e1de7bab15b078930ba712a3f8b0647f5bce8e8c), [and here](https://github.com/zoffixznet/Ignorant-Designer/commit/aa18b17f8e2d81c300fe189cbf8ba2d8f2286126)).

I agree popularity of a language should be factored into Linguist's heuristics and we shouldn't mindlessly work around a problem in the tools. The tools need to be fixed.

:-1:

PR #815, opened 2015-06-30
added remove_all in Mojo::Parameters

chenri:
When converting a cgi script to mojo I notice that I needed 'delete_all' equivalent in CGI. This change added a simple method 'remove_all' in Mojo::Parameters which does that. It is a very simple but convenient method to wipe out all input parameters.

### DIFF
--- a/lib/Mojo/Parameters.pm
+++ b/lib/Mojo/Parameters.pm
@@ -127,6 +127,13 @@ sub remove {
   return $self;
 }
 
+sub remove_all {
+  my ($self) = @_;
+  my $pairs = $self->pairs;
+  @$pairs=();
+  return $self;
+}
+
 sub to_hash {
   my $self = shift;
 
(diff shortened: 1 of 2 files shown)


kraih:
:-1: Just using `$params->pairs([])` requires less typing.

jberger:
I could also see just adding a snippet to the pairs documentation

chenri:
This is true for someone who really knows which method does
what. The name remove_all makes it easier for folks to
search for such functionality. One needs to know the innards
to see that $params->pairs([]) is the right one to use.

chenri:
Yes. Having verbiage relating remove_all to this pair method
will work just fine.

kraih:
Added to documentation. https://github.com/kraih/mojo/commit/fc39d9d4758eb7a48e008ccde8d41876dc34aa85

PR #820, opened 2015-07-12
doc: improved cookbook example

jwbargsten:


lib/Mojolicious/Guides/Cookbook.pod
```diff
@@ -610,7 +610,7 @@ L<Mojolicious::Controller/"send">.
     <head><title>Echo</title></head>
     <body>
       <script>
-        var ws = new WebSocket('<%= url_for('echo')->to_abs %>');
+        var ws = new WebSocket('<%= url_for('echo')->to_abs->scheme('ws') %>');
```
jhthorsen: I don't see why this is required. The websocket route defines the scheme as well.

jwbargsten:
True, works fine without. I originally used `url_for('/echo')`, which will not automatically detect the websocket route.

jberger:
does the example fail without the changes?

kraih:
Closing this pull request since there's no explanation for the changes.

jwbargsten:
Indeed, the changes lack an explanation, so I'll try to give one:
If you are new to websockets (as I am), one frequent approach is to take an example (such as the one from the cookbook) and modify it. However, this particular example easily leads (at least in my case) to confusion, because this:
```
    <script>
      var ws = new WebSocket('<%= url_for('echo')->to_abs %>');

      // Incoming messages
      ws.onmessage = function(event) {
        document.body.innerHTML += event.data + '<br/>';
      };

      // Outgoing messages
      ws.send('Started websocket!');
      window.setInterval(function () { ws.send('Hello Mojo!') }, 1000);
    </script>
```
does **NOT** work anymore. My idea was to make beginner's life easier and clarify that the `onopen` event needs to be used. I fully agree that the `->scheme('ws')` stuff is not necessary, but using the `onopen` method might save a lot of hassle. I hope this explanation is sufficient.

kraih:
@jwbargsten That would have been a very reasonable explanation, i've already changed that part a few hours ago for more consistency with our other WebSocket examples though (which already use `onopen`). https://github.com/kraih/mojo/commit/4d0563399ce932c122adfabb8030185f81cb349a

PR #824, opened 2015-08-05
Fixed skipping validation of optional fields for multiple values

kak-tus:
In case, when we have optional field with multiple values a validation is skipping, when one of this values is empty.

My opinion: if validation is optional for this field, it must skip empty values and check nonempty values.

I have fixed "required" method to its "original" functionality.

Fixed setup "output" of optional fields in case if there exists nonempty values.

And we skip check of empty values in case of "is_optional" flag.

Notice: we must persist order of values in optional field, so we store all values (even empty) in "output".

### DIFF
--- a/lib/Mojolicious/Validator/Validation.pm
+++ b/lib/Mojolicious/Validator/Validation.pm
@@ -26,7 +26,7 @@ sub check {
   my $cb          = $self->validator->checks->{$check};
   my $name        = $self->topic;
   my $input       = $self->input->{$name};
-  my $is_optional = $self->is_optional($name);
+  my $is_optional = $self->is_optional;
   for my $value (ref $input eq 'ARRAY' ? @$input : $input) {
     next if $is_optional && (!defined($value) || !length($value));
     next unless my $result = $self->$cb($name, $value, @_);
--- a/lib/Mojolicious/Validator/Validation.pm
+++ b/lib/Mojolicious/Validator/Validation.pm
@@ -72,7 +72,7 @@ sub optional {
   my @input = ref $input eq 'ARRAY' ? @$input : $input;
   $self->output->{$name} = $input if grep { defined($_) && length($_) } @input;
 
-  return $self->topic($name)->is_optional($name);
+  return $self->topic($name)->is_optional(1);
 }
 
 sub param { shift->every_param(shift)->[-1] }
@@ -160,7 +160,7 @@ L<Mojolicious::Validator> object this validation belongs to.
 =head2 is_optional
 
   my $is_optional = $validation->is_optional;
-  $validation     = $validation->is_optional('foo');
+  $validation     = $validation->is_optional(1);
 
 Flag, indicating optionality of field currently being validated.
 
--- a/lib/Mojolicious/Validator/Validation.pm
+++ b/lib/Mojolicious/Validator/Validation.pm
@@ -4,7 +4,7 @@ use Mojo::Base -base;
 use Carp         ();
 use Scalar::Util ();
 
-has [qw(csrf_token topic validator)];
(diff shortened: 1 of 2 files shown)


jberger:
I am not a frequent user of the validation framework so I will not comment on utility or implementation, but I know that this patch cannot be considered without tests. I will leave this open for now so that others may comment on the former, but without the latter this will likely be closed.

kak-tus:
If the community approves the patch as necessary at a whole, I will add the requested tests.

lib/Mojolicious/Validator/Validation.pm
```diff
@@ -23,10 +23,12 @@ sub check {
 
   return $self unless $self->is_valid;
 
-  my $cb    = $self->validator->checks->{$check};
-  my $name  = $self->topic;
-  my $input = $self->input->{$name};
+  my $cb          = $self->validator->checks->{$check};
+  my $name        = $self->topic;
+  my $input       = $self->input->{$name};
+  my $is_optional = $self->is_optional($name);
```
jhthorsen: ```$is_optional``` will always be true here. Not sure if you expect to fetch the old value or what's going on here :disappointed:

lib/Mojolicious/Validator/Validation.pm
```diff
@@ -148,6 +157,13 @@ Name of field currently being validated.
 
 L<Mojolicious::Validator> object this validation belongs to.
 
+=head2 is_optional
+
+  my $is_optional = $validation->is_optional;
+  $validation     = $validation->is_optional('foo');
+
+Flag, indicating optionality of field currently being validated.
```
jhthorsen: It's very confusing to read the code where you store ```$name```in is_optional(), the example code takes a string, while the documentation says it's a flag.

jhthorsen:
This pull request is very confusing to me. I don't understand what you're trying to accomplish. Need tests and example code for me to consider thumbs up. Are you trying to make a require() that skip values with length == 0...?

kak-tus:
@jhthorsen Main problem, that I want to fix: if validation is optional for field, it must skip empty values in this field and check nonempty values.

jhthorsen:
So what's wrong with optional() ?

http://mojolicio.us/perldoc/Mojolicious/Validator/Validation#optional
http://mojolicio.us/perldoc/Mojolicious/Guides/Rendering#Form-validation

kak-tus:
Check like
```
$validation->optional('foo')->in('baz')->has_error
```
don't produce error for values foo => [ '','bar' ].

jhthorsen:
Right! Thanks for the test and explanation. I've been looking at both the old and new code, and I see that it can be a bit confusing. Unfortunately, I don't like "is_optional" but I don't have a better alternative at the moment.

I think the output from this program is very confusing:

```
use Mojo::Base -strict;
use Test::More;
use Mojolicious::Validator;
my $validator = Mojolicious::Validator->new;

my @cases = (
  ['', 'bar'],     # not ok: optional '', 'bar'
  ['bar', 'foo'],  # ok
  'bar',           # ok
);

for my $c (@cases) {
  my $validation = $validator->validation;
  diag ref $c ? join ', ', map {qq('$_')} @$c : $c;
  $validation->input->{foo} = $c;
  ok $validation->optional('foo')->is_valid, "optional is_valid";
  ok !$validation->has_error, "optional has no error";
  ok $validation->in('baz')->has_error, "in has error";
}

done_testing;
```

kraih:
Maybe validation of optional values should work like this, but the implementation gets a definitive :-1: from me.

jhthorsen:
I agree with sri's comment.

kak-tus:
How can I do it better?

kraih:
It should work correctly now.

PR #832, opened 2015-08-20
Add shortcut for logger to controller, update examples and add pod

marcusramberg:


kraih:
Like i said on IRC, definitive :-1: from me. Aside from the documentation flaws and security issues, i do not agree that this is worth cluttering up `Mojolicious::Controller` further.

Grinnz:
:-1: as it is now. I would be :+1: if the security issue @kraih mentioned is fixed (hide method from the router), or if it was changed to a helper; and if https://github.com/kraih/mojo/blob/c_log/lib/Mojolicious/Controller.pm#L408 was changed to be a different example, since $c->app would no longer be needed for the log.

kraih:
@Grinnz It's not just examples, many descriptions need to be checked and updated as well. https://github.com/kraih/mojo/blob/c_log/lib/Mojolicious/Guides/Tutorial.pod#mode

lib/Mojolicious/Routes.pm
```diff
@@ -11,7 +11,7 @@ use Scalar::Util 'weaken';
 has base_classes => sub { [qw(Mojolicious::Controller Mojo)] };
 has cache        => sub { Mojo::Cache->new };
 has [qw(conditions shortcuts)] => sub { {} };
-has hidden     => sub { [qw(attr has new tap)] };
+has hidden     => sub { [qw(attr has new tap log)] };
```
kraih: Wrong place, this belongs in `Mojolicious.pm`.

kraih:
Looks like this pull request didn't pass the vote.

PR #835, opened 2015-08-25
update Makefile.PL to meta-spec v2 and set dynamic_config to 0

Grinnz:
`dynamic_config` should really be disabled so that CPAN tools know that Mojolicious has static prerequisities (as per https://metacpan.org/pod/CPAN::Meta::Spec#dynamic_config). To do this, the `META_MERGE` contents are updated to follow v2 of the META spec. The diff of the resulting META.json with the current META.json is linked below.

https://gist.github.com/Grinnz/ca3f85988463e34b8118

### DIFF
--- a/Makefile.PL
+++ b/Makefile.PL
@@ -15,15 +15,21 @@ WriteMakefile(
   AUTHOR       => 'Sebastian Riedel <sri@cpan.org>',
   LICENSE      => 'artistic_2',
   META_MERGE   => {
-    requires  => {perl => '5.010001'},
-    resources => {
-      license    => 'http://www.opensource.org/licenses/artistic-license-2.0',
+    dynamic_config => 0,
+    'meta-spec'    => {version => 2},
+    no_index       => {directory => ['examples', 't']},
+    prereqs        => {runtime => {requires => {perl => '5.010001'}}},
+    resources      => {
+      license    => ['http://www.opensource.org/licenses/artistic-license-2.0'],
       homepage   => 'http://mojolicio.us',
-      bugtracker => 'https://github.com/kraih/mojo/issues',
-      repository => 'https://github.com/kraih/mojo.git',
+      bugtracker => {web => 'https://github.com/kraih/mojo/issues'},
+      repository => {
+        type => 'git',
+        url  => 'https://github.com/kraih/mojo.git',
+        web  => 'https://github.com/kraih/mojo',
+      },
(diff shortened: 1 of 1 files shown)


kraih:
What do we gain by applying this patch?

Makefile.PL
```diff
@@ -15,15 +15,21 @@ WriteMakefile(
...
+    resources      => {
+      license    => ['http://www.opensource.org/licenses/artistic-license-2.0'],
       homepage   => 'http://mojolicio.us',
-      bugtracker => 'https://github.com/kraih/mojo/issues',
-      repository => 'https://github.com/kraih/mojo.git',
+      bugtracker => {web => 'https://github.com/kraih/mojo/issues'},
+      repository => {
+        type => 'git',
+        url  => 'https://github.com/kraih/mojo.git',
+        web  => 'https://github.com/kraih/mojo',
```
kraih: Why three settings? MetaCPAN seems to already have all the information it needs.

zoffixznet:
> dynamic_config should really be disabled so that CPAN tools know that Mojolicious has static prerequisities 

Does it have static prereqs? I thought it loaded a couple of modules for older perls.

Mojolicious wasn't working correctly when I installed it on one perl and then upgraded perl but not Mojolicious. So I think the dynamic aspect is certainly there.

kraih:
Looks like MetaCPAN uses the `dynamic_config` field to decide if it should display the message `and possibly others` in the dependency list. Considering our extra runtime dependencies for certain features, isn't that exactly what we want?

Makefile.PL
Grinnz: With only the one URL, MetaCPAN simply says "clone repository" instead of having separate "Repository" and "Clone" links. In this case I am just following the meta-spec v2.

Grinnz:
@zoffixznet upgrading perl in-place while having modules installed from CPAN that aren't upgraded will always result in breakage. Mojolicious has no dynamic dependencies.

Grinnz:
@kraih The message "and possibly others" is referring to dynamic dependencies, not optional dependencies. For CPAN tools to be aware of the optional dependencies of Mojolicious they would need to be added as "recommends" or "suggests"; however optional dependencies are not relevant to the `dynamic_config` metadata.

jberger:
From my understanding it would be `suggests` since they shouldn't be
attempted for installation.

PR #836, opened 2015-08-25
More helpful diagnostic on failed connection

pjlsergeant:
Consider:

    my $t = Test::Mojo->new('MyApp');
    $t->get_ok('admin_pages/my_page');

This fails on my Mac with the somewhat unhelpful error message:

    Can't connect: nodename nor servname provided, or not known

This change makes it clearer to the implementor that the problem is 'admin_pages' being treated as the hostname of a remote machine.

### DIFF
--- a/lib/Mojo/UserAgent.pm
+++ b/lib/Mojo/UserAgent.pm
@@ -116,7 +116,11 @@ sub _connect {
 
       # Connection error
       return unless $self;
-      return $self->_error($id, $err) if $err;
+      if ( $err ) {
+        return $self->_error( $id,
+          sprintf( 'Error connecting to %s://%s:%s/: %s',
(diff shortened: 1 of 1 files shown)


jhthorsen:
I can't see how this is better than "Can't connect".

:-1:  from me, but I have no strong feelings about it.

pjlsergeant:
@jhthorsen The difference being it makes it clear that an external connection was attempted to the first part of the URL, which will be surprising and more informative than a "nodename nor servname provided" in a situation when the user didn't think either was needed.

kraih:
Afraid i don't see the gain either, and the patch lacks tests. :-1:

kraih:
Closing this pull request since it cannot pass the vote.

PR #838, opened 2015-09-03
Optional dependency on Unicode::UTF8

Grinnz:
Halves UTF-8 encoding roundtrip time in my naive benchmark:

```
perl -Ilib -Mojo -MMojo::Util=decode,encode -E'my $str = "\x{f1}"; n { decode "UTF-8", encode "UTF-8", $str } 1000000'
```

### DIFF
--- a/lib/Mojo/Util.pm
+++ b/lib/Mojo/Util.pm
@@ -566,7 +566,10 @@ Convert C<CamelCase> string to C<snake_case> and replace C<::> with C<->.
 
   my $chars = decode 'UTF-8', $bytes;
 
-Decode bytes to characters and return C<undef> if decoding failed.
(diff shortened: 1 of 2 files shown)


Grinnz:
I'm not sure if this should be added anywhere else in the documentation. It's not really applicable in Mojo::IOLoop::Client, where the rest of the optional dependencies are documented.

kraih:
:-1: from me i'm afraid, since i don't see this making much of a difference in real-world applications.

jhthorsen:
I'm -1-ish... I don't want more modules in the "mojo version" output. I have no strong feelings about the changes in Mojo::Util.

Grinnz:
Closing with the release of Mojo::Unicode::UTF8.

PR #847, opened 2015-09-25
add callback to uniq method of collection

CandyAngel:
This change adds a callback to the uniq method in Mojo::Collection, allowing users to define what constitutes a unique element. This permits collections of hashes and arrays to be de-duplicated, rather than just elements which stringify reasonably (strings, numbers, overloaded objects).

### DIFF
--- a/lib/Mojo/Collection.pm
+++ b/lib/Mojo/Collection.pm
@@ -93,8 +93,11 @@ sub tap { shift->Mojo::Base::tap(@_) }
 sub to_array { [@{shift()}] }
 
 sub uniq {
+  my ($self, $cb) = @_;
   my %seen;
-  return $_[0]->new(grep { !$seen{$_}++ } @{$_[0]});
+  return $self->new(grep { !$seen{ $cb->($_) }++ } @$self)
+    if ref $cb eq 'CODE';
+  return $self->new(grep { !$seen{$_}++ } @$self);
 }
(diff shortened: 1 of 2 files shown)


lib/Mojo/Collection.pm
```diff
@@ -93,8 +93,11 @@ sub tap { shift->Mojo::Base::tap(@_) }
 sub to_array { [@{shift()}] }
 
 sub uniq {
+  my ($self, $cb) = @_;
   my %seen;
-  return $_[0]->new(grep { !$seen{$_}++ } @{$_[0]});
+  return $self->new(map { (!$seen{ $cb->($_) }++) ? $_ : () } @$self)
```
jhthorsen: Why not use grep instead of map?

jhthorsen:
I'm neutral, leaning toward +1.

zoffixznet:
I'd suggest using a better example in the docs. I like the method call example in [List::UtilsBy::uniq_by](https://metacpan.org/pod/List::UtilsBy#uniq_by)

kraih:
:+1: from me.

lib/Mojo/Collection.pm
CandyAngel: Just a (bad) habit, I use map *far* more than grep, so I use it by default. Fixed.

PR #848, opened 2015-09-26
add attr name to accessor

vovka667:
I need an attribute name in accessor.


kraih:
I'm afraid this pull request has to be rejected right away, because it is incomplete.

jberger:
There would need to be an extremely good use case for this. I'm guessing that this is a shortcut for you making many attributes in a loop, which would not be enough for me. Can you please elaborate? 

Further this cannot be considered without tests and documentation, as with all features. http://mojolicio.us/perldoc/Mojolicious/Guides/Contributing#CONTRIBUTING-CODE

PR #851, opened 2015-10-13
JSON pointer: fix '/foo/'

dolmen:
`/foo/` must return `42` in `{'foo' => { '' => 42}}`.
The bug is due to an edge case of Perl's [`split`](http://perldoc.perl.org/perlfunc.html#split).

### DIFF
--- a/lib/Mojo/JSON/Pointer.pm
+++ b/lib/Mojo/JSON/Pointer.pm
@@ -13,7 +13,7 @@ sub _pointer {
 
   my $data = $self->data;
   return $data unless $pointer =~ s!^/!!;
-  for my $p ($pointer eq '' ? ($pointer) : (split '/', $pointer)) {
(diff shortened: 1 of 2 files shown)


jberger:
TIL:

>If LIMIT is negative, it is treated as if it were instead arbitrarily large; as many fields as possible are produced.
If LIMIT is omitted (or, equivalently, zero), then it is usually treated as if it were instead negative but with the exception that trailing empty fields are stripped (empty leading fields are always preserved); if all fields are empty, then all fields are considered to be trailing (and are thus stripped in this case). 

-- perldoc -f split

dolmen:
I dream of sending a PR on the RFC to add the `/foo/` test case...

dolmen:
@kraih That's not fair to remove original authorship from git commits...

kraih:
@dolmen That was not at all my intention, if a pull request is sensible but doesn't replicate the style of the surrounding code, i just fix it and commit it myself with proper attribution.

PR #854, opened 2015-10-14
Allow test descriptions in 1-arg version of json_is and json_message_is

tagg:
The '1-arg' version of json_is and json_message_is does not accept a description, due to how the params are pulled off of `@_`.

So what would seem intuitive:

`$t->json_is( { foo => [1, 2, 3] }, 'a simple test' );`

does not actually work.

This patch fixes that. All existing tests pass.

For this small change, additional tests may be unnecessary?

### DIFF
--- a/lib/Test/Mojo.pm
+++ b/lib/Test/Mojo.pm
@@ -164,7 +164,7 @@ sub json_hasnt {
 
 sub json_is {
   my $self = shift;
-  my ($p, $data) = @_ > 1 ? (shift, shift) : ('', shift);
+  my ($p, $data) = ref $_[0] ? ('', shift) : (shift, shift);
   my $desc = encode 'UTF-8', shift || qq{exact match for JSON Pointer "$p"};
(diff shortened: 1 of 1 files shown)


jberger:
Certainly it would need to be used in the test suite and need to be documented. Regarding functionality, I'm of two minds here. Originally json documents had to be either a hash or an array and so to compare the top level structure it must indeed be a reference.  That said, in the recent spec json documents may be any type and of course you can just use the empty pointer to fill up the slots.

kraih:
:-1: Have to agree with @jberger on all points.

tagg:
The test and docs points I could address, but if you don't agree with the functionality, I'll just close the request again.

kraih:
A `ref` check just couldn't work with RFC 7159 compliant JSON.

tagg:
I see what you mean. The json response could well be unstructured. Thanks.

kraih:
In case someone else stumbles over this pull request in the future, what you can do is `$t->json_is('' => {foo => [1, 2, 3]}, 'right data structure')`.

PR #858, opened 2015-10-26
Improve punctuation

s1037989:


zoffixznet:
:+1:

kraih:
:-1: i'm afraid, because you've changed `application wide` to `application-wide` only in one place, which is inconsistent.

kraih:
I've fixed the typo everywhere, since the punctuation changes need to be discussed separately. https://github.com/kraih/mojo/commit/8ea4618cf9a6c7c717b4062d282b353f6d2af086

kraih:
We've talked about this on IRC, and there's many different opinions on how punctuation should look in that sentence. Not sure how to settle this.

s1037989:
No problem!  Usually, FWIW, that's a sign (different opinions) that reducing the sentence to shorter and more sentences is in order.  Longer sentences can often be more efficient and clever, but readability (I think specifically in the case of this project) is likely the number one goal.  That might be just something worth keeping in mind, not necessarily striving for.

For uninterrupted clear readability, I do believe that "for example" should be offset with commas.  I think this is a case of "misplaced modifier".

"it allows you for example to customize" -> "it allows you to, for example, customize"

PR #865, opened 2015-10-30
fixed multi-line attribute behavior

jamadam:
Since mojo v6.26, Mojo::DOM behavior has changed and now it doesn't handle
multi-line attributes.

    <div test="line1
                line2" />

According to some specs HTML5 doesn't allow such attributes, on the other hand
XML seems to do.

I referred to the following document, hoping it's the right one for me.

http://www.w3.org/TR/xml/


kraih:
Thanks, applied.

PR #871, opened 2015-11-11
Fixed url_for to handle route named against package namespaces

jjehannet:
My application uses route names generated by the package namespaces (REST framework).

My problem is I can't use `url_for` in controllers because the regex url considers colon presence as an absolute URL. The patch changes now the regex to refuse two consecutive colons in the URL detection.

Are you interested in merging this pull request ?

I tried to add some tests in the codebase but I fear unnecessary code.

### DIFF
--- a/Changes
+++ b/Changes
@@ -1,5 +1,6 @@
 
 6.30  2015-11-04
+  - Fixed url_for to handle route named against package namespaces.
 
 6.29  2015-11-03
   - Fixed a few bugs in built-in templates. (Zoffix, sri)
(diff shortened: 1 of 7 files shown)


kraih:
I don't think this would be correct, `foo::bar` should be a valid absolute URL.

jjehannet:
Mojo::URL parses `foo::bar` as scheme="foo" and path=":bar" but I didn't think It was an acceptable URL.
Feel free to delete the pull request. I can still use this snippet:

    my $route =  $c->app->routes->lookup(__PACKAGE__);
    $c->url_for($route->render($params));

PR #872, opened 2015-11-12
Add items_in helper

zoffixznet:
This PR introduces the `items_in` Default Helper. I wasn't sure where the tests for it should be placed; let me know and I'll add them.

### Use Case

The use case is to make your templates code more concise and easier to read, when dealing with arrayrefs (potentially stored in the `stash`). Here are some of the examples that demonstrate the code with and without the proposed `items_in` helper:

Without:
```
% if ( @{stash('entries') || []} ) {
  % for ( @{stash('entries') || []} ) {
    ...
  % }
% }
```
With:
```
% if ( items_in 'entries' ) {
  % for ( items_in 'entries' ) {
    ...
  % }
% }
```

Without:
```
% for ( @{flash('entries') || []} ) {
  ...
% }
```

With:
```
% for ( items_in flash 'entries' ) {
  ...
% }
```

Without:
```
% for ( @{$some->{var} || []} ) {
  ...
% }
```

With:
```
% for ( items_in $some->{var} ) {
  ...
% }
```

### DIFF
--- a/lib/Mojolicious/Plugin/DefaultHelpers.pm
+++ b/lib/Mojolicious/Plugin/DefaultHelpers.pm
@@ -34,6 +34,14 @@ sub register {
 
   $app->helper(dumper => sub { shift; dumper @_ });
   $app->helper(include => sub { shift->render_to_string(@_) });
+  $app->helper(
+    items_in => sub {
+      my ($c, $what) = @_;
+      return unless defined $what;
+      $what = $c->stash($what) // [] unless ref $what;
+      return @$what;
+    }
+  );
 
   $app->helper("reply.$_" => $self->can("_$_")) for qw(asset static);
 
@@ -374,6 +382,15 @@ response headers with L<Mojolicious::Static/"is_fresh">.
   $c->is_fresh(etag => 'abc', last_modified => 1424985708)
     ? $c->rendered(304)
     : $c->render(text => 'I ♥ Mojolicious!');
+
+=head2 items_in
+
+  % if  ( items_in 'stash_variable' ) { ... }
+  % for ( items_in $array_ref ) { ... }
+
+Dereference an array reference. If the passed argument is not a reference,
+it is given to L<Mojolicious::Controller/"stash"> and the return value
+is dereferenced instead.
 
 =head2 layout
 


jberger:
Might be nice to do keys of a hashref too

zoffixznet:
> Might be nice to do keys of a hashref too

I'm a bit :-1: on that. I wouldn't call the keys the "items" of a hashref. And the utility of the helper in this situation diminishes, since to access the values you'd still have to make extra calls to `stash` to fetch the value.

nnutter:
I'm not a Mojo dev but this seems like it overlaps heavily with `each`, `keys`, and `values` already provided by Perl when using automatic dereferencing.

Grinnz:
FYI: https://metacpan.org/pod/release/WOLFSAGE/perl-5.23.1/pod/perldelta.pod#The-autoderef-feature-has-been-removed

jhthorsen:
I think this is useful for templates only..? If so, wouldn't it be better with something like:

    %= items_in "entries", begin
      <li>item <%= $_ %></li>
    % end

kraih:
@jhthorsen Or even two helpers. :wink:
```
%= items_in entries => begin
  <li><%= item %></li>
% end
```

jhthorsen:
I'm :-1: on the implementation in this PR. I don't see that it has enough value. The reason for this is very subjective: I always extract my data in the controller code and the just do ```for(@$foo)``` in my templates.

I haven't decided if I like my or @kraih's suggestion.

zoffixznet:
@jnthorsen didn't you say in August that that method would have to duplicate all the data ( for the item helper) or am I misremembering?

jhthorsen:
No, I think you remember correctly. I don't know how the "item" helper could access an aliased value. The simple solution would be ```helper item=>sub{$_}```, but that doesn't really help much imo.

kraih:
@jhthorsen The `items_in` helper would loop over the items and store them in the `item` stash value (`local $stash->{item} = $item`). And the `item` helper would just be a wrapper around that stash value, similar to how the `layout` and `title` helpers work.

jhthorsen:
@kraih: Will local() alias the values? I thought it made a copy... Maybe Perl is clever enough to do copy on modification or something?

What I mean is that if $item contains a 10MB scalar: Will the application use 20MB after ```local $stash->{item} = $item```?

kraih:
@jhthorsen You're overthinking it. Why would we want to optimize for 20MB blobs in templates?

jhthorsen:
Hehe... Guess you're right :smile_cat:

zoffixznet:
I fished out @jhthorsen's implementation of this: https://ssl.thorsen.pm/paste/f7cf6d717dfe

It uses the `every` for the name helper instead, which might be more appropriate for looping, but the helper should also function inside an `if(){}`...

So, to sum up progress so far:
1) The original implementation needs to be improved to use `begin` => `end` blocks, e.g. like in the link above.
2) If begin/end are not provided, the helper should return a list of items instead of looping over them
3) The name of the helper stays as `items_in`

Thoughts?

jhthorsen:
I think this is better (if you replace the global value) https://ssl.thorsen.pm/paste/cce766a3ff5f

zoffixznet:
I now think this PR should be closed, based on yesterday's conversation [http://irclog.perlgeek.de/mojo/2015-11-12#i_11526666] regarding optional stash vars being an anti-pattern that creates multiple templates, and trying to write docs explaining how the `every` works (they ended up convoluted).

The most common use for me is stash vars, but if we get rid of optional stash vars, the first example code above becomes:
#### Without helper
```perl
% if ( @$entries ) {
  % for ( @$entries ) {
    <%= $_ =>
  % }
% }
```

#### With helper
```perl
% if ( items_in 'entries' ) {
  %= items_in entries => begin
    <%= item =>
  % end
% }
```
Thus, by avoiding the helper, the code is more consice and is  pure Perl, which is easier to undertand than a helper.

PR #874, opened 2015-11-17
Css parser fix

jamadam:
This PR improves Mojo::DOM::CSS behavior with some test cases. It's not an atomic PR (sorry) so partial merger is also welcome.

* Reduces following Perl errors when :nth-* equation is '0'. Believe it or not, nth-child(0) is valid CSS.

> Use of uninitialized value in multiplication (*) at /Users/sugamakeita/Documents/dev/mojo-fork/lib/Mojo/DOM/CSS.pm line 174, <DATA> line 2125.
> Use of uninitialized value in addition (+) at /Users/sugamakeita/Documents/dev/mojo-fork/lib/Mojo/DOM/CSS.pm line 174, <DATA> line 2125.

* Fixes a bug where :not pseude-class name have been unexpectedly case-sensitive.
* Slightly faster css matching.


kraih:
Thanks, applied. (and cleaned up)

PR #885, opened 2015-12-12
fix for both ? and form for get

fayland:


kraih:
This pull request breaks existing tests and doesn't add new ones.

PR #887, opened 2015-12-15
Allow developers to override url_escape() chars

DabeDotCom:
This is to "fix" a problem with URLs that included %28/%29 parentheses.  For example:
```
  $self->redirect_to("http://example.org/foo%20%28bar%29.txt");
```
would always send back:
```
  302 Found
  Location: http://example.org/foo%20(bar).txt
```
Although that's not *strictly* invalid, I was having trouble downstream, where a third-party app [over which I have no control] was cutting it off, thusly:
```
  BAD:  <a href="http://example.org/foo%20">foo </a>(bar).txt

  GOOD: <a href="http://example.org/foo%20%28bar%29.txt">foo (bar).txt</a>
```
Rather than hard-coding a different set of escape characters globally, which has the potential to break something, this patch allows developers to remove those characters -- or add other ones -- if they choose:
```
  $app->hook( after_dispatch => sub {
      $Mojo::Path::url_escape_chars =~ s/[()]//g;
  } );
```

### DIFF
--- a/lib/Mojo/Path.pm
+++ b/lib/Mojo/Path.pm
@@ -10,6 +10,9 @@ use Mojo::Util qw(decode encode url_escape url_unescape);
 
 has charset => 'UTF-8';
 
+# Allow developers to override which characters AREN'T url_escape()-ed
+our $url_escape_chars = '^A-Za-z0-9\-._~!$&\'*+,;=:@()';
+
 sub canonicalize {
   my $self = shift;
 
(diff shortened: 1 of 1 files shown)


jberger:
This patch contains no new tests nor documentation and therefore by policy cannot be considered, that said, using a package variable is out of style for the Mojolicious project. If nothing else it would be an attribute on the path object.

DabeDotCom:
(Sorry for the missing tests; I've since added a couple, but I won't bother submitting a new PR until the actual implementation Best Practices are sorted out, below.)

I had initially attempted to use an attribute, but I don't actually have easy access to the `Mojo::Path` *instance* from within my app...

All my app does is call `Mojolicious::Controller::redirect_to("http://example.org/foo%28bar%29.txt")`.  After that, `M::C::redirect_to()` calls `M::C::url_for()`, which instantiates a new `Mojo::URL` object, whose `Mojo::URL::parse()` method calls `Mojo::URL::path()`, which finally instantiates the new `Mojo::Path` object, which is where said attribute would ultimately live.

That's **WAY** farther down the rabbit hole than I wanted to venture...  :-}

I also considered sticking something in the stash — `mojo.path_escape_chars` perhaps? — but I wasn't sure if adding another reserved stash value was a good idea (and besides, `Mojo::*` packages shouldn't have access to controller-level data, anyway...)

Although it *is* "spooky action at a distance", I figured a package variable (à la `$Data::Dumper::Maxdepth`) was the 'least ugly' solution.

But I'm certainly open to other suggestions, hehe...  Cheers!

kraih:
There's also consistency issues, since `Mojo::URL` and `Mojo::Parameters` also percent encode parts of the URL.

PR #894, opened 2016-01-09
include Mojo::Channel, don't include docs yet

jberger:


jhthorsen:
I think you should just rebase this over "my" single change manually @jberger. Go crazy :+1:

jhthorsen:
I do find it a tad humorous that the tests fail because of "Mojo::Channel: couldn't find pod", hehe

PR #900, opened 2016-01-21
fix log message bug in Mojo::Server::Daemon ( issue 899)

sjdy521:
https://github.com/kraih/mojo/issues/899


kraih:
https://github.com/kraih/mojo/issues/899#issuecomment-173573179

PR #902, opened 2016-01-30
Allow to include templates which are extending same template

KES777:
```
$ cat t1.html.ep
% content test => begin
   BASE
% end
$ cat t2.html.ep
% extends 't1';
% content test => begin
  T2
%
$ cat t3.html.ep
% extends 't1';
% content test => begin
  T3
%
$ cat t4.thml.ep
%= include 't2';
%= include 't3';
```
Trying ot rendre 't4' will result:
T2
T2

Insteand of expected:
T2
T3

I have not change, but you should pay attention to Renderer.pm:136 and Plugin/DefaultHelpers.pm:58 lines. This is critical. 
If template we are including (T2) will define 'mojo.content' ( % content test in example ) this will have global effect and will prevent 'content test' definition in template (T3) in same way as it prevent for extending tempalte (T1)

Same effect exists also for 'layout', 'extends' so you should check Renderer.pm:90-91 lines.

### DIFF
--- a/lib/Mojolicious/Renderer.pm
+++ b/lib/Mojolicious/Renderer.pm
@@ -89,6 +89,8 @@ sub render {
   my $stash = $c->stash;
   local $stash->{layout}  = $stash->{layout}  if exists $stash->{layout};
   local $stash->{extends} = $stash->{extends} if exists $stash->{extends};
+  local $stash->{ 'mojo.content' } ||=  {};
+
 
   # Rendering to string
   local @{$stash}{keys %$args} if my $string = delete $args->{'mojo.string'};


jberger:
Thanks for the report. I certainly agree that I would expect that your example would work as you suggest. Your proposed patch breaks the pattern of setting `content_for` before rendering and would need to be at least `local $stash->{ 'mojo.content' } = $stash->{ 'mojo.content' }` if not `local $stash->{ 'mojo.content' } = $stash->{ 'mojo.content' } if exists $stash->{ 'mojo.content' };` like the lines above it. Further, to be accepted it would need to include tests which could basically be your example turned into test form.

jberger:
I have investigated further and here are my results:

First this is the failing test you might have written (don't worry about the `layout undef`, that has to do with the vagueries of this particular test)

```diff
diff --git a/t/mojolicious/layouted_lite_app.t b/t/mojolicious/layouted_lite_app.t
index 7e65ea6..d6976c1 100644
--- a/t/mojolicious/layouted_lite_app.t
+++ b/t/mojolicious/layouted_lite_app.t
@@ -51,6 +51,8 @@ get '/double_inheritance' =>
 
 get '/triple_inheritance';
 
+get '/base_reuse_both';
+
 get '/mixed_inheritance/first' => {template => 'first'};
 
 get '/mixed_inheritance/second' => {template => 'second', layout => 'green'};
@@ -201,6 +203,9 @@ $t->get_ok('/triple_inheritance')->status_is(200)
   ->content_is("<title>Works!</title>\n<br>\nSidebar too!\n"
     . "New <content>.\nShared content!\n\nDefault footer!\n");
 
+$t->get_ok('/base_reuse_both')->status_is(200)
+  ->content_is("Base always\n\n  One main\n\nBase always\n\n  Two main\n\n");
+
 # Mixed inheritance (with layout)
 $t->get_ok('/mixed_inheritance/first')->status_is(200)
   ->header_is(Server => 'Mojolicious (Perl)')
@@ -392,6 +397,29 @@ Sidebar too!
 New <content>.
 %= content_for 'stuff'
 
+@@ base_reuse_base.html.ep
+Base always
+%= content main => begin
+  Base main
+% end
+
+@@ base_reuse_one.html.ep
+% extends 'base_reuse_base';
+%= content main => begin
+  One main
+% end
+
+@@ base_reuse_two.html.ep
+% extends 'base_reuse_base';
+%= content main => begin
+  Two main
+% end
+
+@@ base_reuse_both.html.ep
+% layout undef;
+%= include 'base_reuse_one'
+%= include 'base_reuse_two'
+
 @@ layouts/plugin_with_template.html.ep
 layout_with_template
 <%= content %>
```

The problem with your proposal is it breaks another pattern, wherein before renderering you can pre-set content buffers like so: `$c->content_for(some_buffer => 'thingy')`. This is used even within the mojolicious core and indeed were the tests you were breaking your original PR.

There are some ways you can mitigate this behavior using existing mojolicious functionality, first you could use `content_with` in your child templates which more aggressively replaces existing content buffers
```perl
@@ base_reuse_one.html.ep                                                                 
% extends 'base_reuse_base';                                                              
%= content_with main => begin
  One main
% end                                                                                     
                                                                                           
@@ base_reuse_two.html.ep                                                                 
% extends 'base_reuse_base';                                                              
%= content_with main => begin
  Two main
% end  
```
or you could create a new "compilation unit" (really a new stash) by doing
```perl
@@ base_reuse_both.html.ep
% layout undef;
%= app->build_controller->include('base_reuse_one')
%= app->build_controller->include('base_reuse_two')
```

Going forward, while I cannot support your proposal, I might be amenable to allowing an explicit buffer reset ala `content_with main => undef` or else a new helper (which we always are loath to add) which would locally replace the existing buffers `with_new_buffers { include 'base_reuse_two' }`

As such, I'm going to close this PR as it cannot be accepted as-is and any other proposal would be a new one.

PR #903, opened 2016-01-31
Do not affect content of including template

KES777:
When we include some template we want its content. So $stash->{'mojo.content'}
shoould be localized.

Without that when our main template **occasionally** has content with same name
this will prevent the template we are including to generate its own
Even if we include two templates which have contents with same name the first
one will prevent following to generate right content.
In large system things goes too wired and complex to debug

https://github.com/kraih/mojo/pull/902

### DIFF
--- a/lib/Mojolicious/Renderer.pm
+++ b/lib/Mojolicious/Renderer.pm
@@ -89,6 +89,8 @@ sub render {
   my $stash = $c->stash;
   local $stash->{layout}  = $stash->{layout}  if exists $stash->{layout};
   local $stash->{extends} = $stash->{extends} if exists $stash->{extends};
+  local $stash->{'mojo.content'} ||= {} if $args->{'mojo.string'};
+
 
   # Rendering to string
   local @{$stash}{keys %$args} if my $string = delete $args->{'mojo.string'};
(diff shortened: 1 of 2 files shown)


kraih:
I've added more tests that break this patch. https://github.com/kraih/mojo/commit/fad35ff0765ec01d21bbbd05083aa21d33297b71

jberger:
In addressing #902 I had also considered treating `render_to_string` differently to reset the content buffers as you mention. In the end the problem is that there is no reason that this should be true. In the same way that you might want to prepopulate content buffers for rendering, you might want to do so when rendering to a string (same as include).

As I suggested in that PR, to address this, we would need to propose some new way to indicate that the user WANTS to reset a buffer or all buffers. I don't see any way to proceed in the direction of trying to guess at intentions.

jberger:
I've given this some more thought and again I cannot get around the idea that there is no way to divine the user's intent when including templates. Content buffers are per-request not per-render and have been this way for the entire history of their existence. Changing that now is too painful. There would have to be some mechanism as I describe [here](https://github.com/kraih/mojo/pull/902#issuecomment-177259442) to disambiguate.

KES777:
How about `%= standalone 'template';` ?

PR #904, opened 2016-02-10
Mojo::Pointer should throw an error when JSON Pointer is missing a s…

kivilahtio:
…tarting '/' (against RFC 6901)

Hi!

I have a JSON object, which had a bug:

`
{
  "type": "object",
  "properties": {
    "listContentNumber": { "$ref": "../definitions.json#listContentNumber" },
    "biblionumber": { "$ref": "../definitions.json#/biblionumber" },
    "itemnumber": { "$ref": "../definitions.json#/itemnumber" },
    "borrowernumber": { "$ref" : "../definitions.json#/borrowernumber" }
  }
}
`

can you spot it?
Mojo::Pointer couldn't, and I hope you can accept my solution to improve it.

Looking into the ietf specification for JSON Pointer:
    http://tools.ietf.org/html/rfc6901#section-3
if the pointer has a 'reference-token', it MUST be prepended by a '/'-character.

  **>   json-pointer = *( "/" reference-token )   <**
  * any number of "/" + reference-token -tokens are allowed, but if the reference-token is defined,
  * the prepending "/" is mandatory and the JSON Pointer implementation SHOULD throw an error.

for example this is invalid:
    { "$ref": "../definitions.json#listContentNumber" }
whereas this is valid:
    { "$ref": "../definitions.json#/listContentNumber" }

If the reference-token doesn't 'start' with '/' then it returns the whole data-object!
It should check for pointer validity and die instead.

### DIFF
--- a/lib/Mojo/JSON/Pointer.pm
+++ b/lib/Mojo/JSON/Pointer.pm
@@ -1,5 +1,6 @@
 package Mojo::JSON::Pointer;
 use Mojo::Base -base;
+use Mojo::Exception;
 
 has 'data';
 
@@ -10,6 +11,8 @@ sub new { @_ > 1 ? shift->SUPER::new(data => shift) : shift->SUPER::new }
 
 sub _pointer {
   my ($self, $contains, $pointer) = @_;
+
+  Mojo::Exception->throw(__PACKAGE__.": Invalid JSON Pointer '$pointer'. Pointer must start with '/' or be undefined") if ($pointer && $pointer =~ /^[^\/].+/);
 
   my $data = $self->data;
   return $contains ? 1 : $data unless $pointer =~ s!^/!!;
--- a/t/mojo/json_pointer.t
+++ b/t/mojo/json_pointer.t
@@ -102,4 +102,10 @@ is $pointer->get('/k"l'),   6,     '"/k\\"l" is 6';
 is $pointer->get('/ '),     7,     '"/ " is 7';
 is $pointer->get('/m~0n'),  8,     '"/m~0n" is 8';
 
+# catch invalid pointer reference-tokens
+eval {
+  $pointer->get('no_slash');
+};
+ok $@ =~ /Invalid JSON Pointer 'no_slash'/, 'reference-token must start with "/"';
+
 done_testing();


lib/Mojo/JSON/Pointer.pm
```diff
@@ -11,6 +12,8 @@ sub new { @_ > 1 ? shift->SUPER::new(data => shift) : shift->SUPER::new }
 sub _pointer {
   my ($self, $contains, $pointer) = @_;
 
+  Mojo::Exception->throw(__PACKAGE__.": Invalid JSON Pointer '$pointer'. Pointer must start with '/' or be undefined") if ($pointer && $pointer =~ /^[^\/].+/);
```
kraih: Why do you use `Mojo::Exception` here? The tests don't even use any of its features.

kraih:
The proposed behavior seems inconsistent with how we handle other DSLs, like CSS selectors in `Mojo::DOM`.

jhthorsen:
:-1: from me. I don't see why mojo should check the input. Seems like a corner case to me.

jberger:
I would suggest leaving the door open to having stuff before a /, specifically if we ever supported JSON References more tightly in the Mojo::JSON ecosystem: https://tools.ietf.org/html/draft-pbryan-zyp-json-ref-03

Of course the APIs for that would have to be worked out then but I don't want to close the door on it now by promising that we die in that case.

kraih:
Thanks, i think we have reached consensus on not throwing an exception for invalid JSON Pointers.

lib/Mojo/JSON/Pointer.pm
kivilahtio: Because I think throwing exceptions is better than dying.

kivilahtio:
Thank you for reviewing my change request.
Why I think this was a welcome fix, was due to the package actually claiming it is a JSON::Pointer, even referencing the official standard in the doc (however relaxed).
If you reference a standard you should follow and enforce it.
If you need a more lax pointer somewhere, there should be another implementation for it, but that might take more work.

However it is not too difficult to validate the JSON first using a json-schema v4 validator to alleviate any future related issues.

kraih:
Personally, i'm a strong believer of the rubustness principle "Be conservative in what you send, be liberal in what you accept".

PR #905, opened 2016-02-10
make csrf_protect handle multiple tokens

kristoff3r:
Currently csrf_protect will always fail if more than one input called csrf_token is present. It would be more intuitive if it always checks the last value like param, which I have implemented in this pull request.

You could argue that it should be an error to have multiple csrf_token's, but then it should still be explicitly checked for in my opinion.

### DIFF
--- a/lib/Mojolicious/Validator/Validation.pm
+++ b/lib/Mojolicious/Validator/Validation.pm
@@ -37,6 +37,7 @@ sub check {
 sub csrf_protect {
   my $self  = shift;
   my $token = $self->input->{csrf_token};
+  $token = ref $token eq 'ARRAY' ? $token->[-1] : $token;
   $self->error(csrf_token => ['csrf_protect'])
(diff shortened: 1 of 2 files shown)


kraih:
What's the use case for this? Having it fail with multiple tokens almost sounds like a feature to me.

jberger:
I don't think that this PR is how I'd go

> You could argue that it should be an error to have multiple csrf_token's, but then it should still be explicitly checked for in my opinion.

That seems much more sensible

jhthorsen:
I can't make sense of the test: Why are you overriding the hash value with another hash value (multiple keys are merged)?
https://github.com/kraih/mojo/pull/905/files#diff-af7f52660125846348aa0b9dd5a72bf1R230

I don't think we should accept a list of csrf tokens. (I guess I'm arguing that multiple csrf_token's is an error)

kraih:
Thanks, i think we have reached consensus on rejecting multiple CSRF tokens.

PR #910, opened 2016-02-17
Added clear info about the validation error

KES777:
That is not explicit from where the $result, @args appeared in the error


lib/Mojolicious/Validator/Validation.pm
```diff
@@ -174,7 +174,7 @@ Validate C<csrf_token> and protect from cross-site request forgery.
 =head2 error
 
   my $err     = $validation->error('foo');
-  $validation = $validation->error(foo => ['custom_check']);
+  $validation = $validation->error(foo => ['custom_check_name','check_result','arg1','arg2']);
```
kraih: This line is too long and doesn't replicate the style of the surrounding documentation.

PR #916, opened 2016-02-25
Return right default value for checkbox/radiobutton

KES777:
As stated at: https://www.w3.org/TR/html5/forms.html#input-type-attr-summary
for checkbox/radiobutton to IDL attribute 'value' should be applied 'default/on'
method: https://www.w3.org/TR/html5/forms.html#dom-input-value-default-on

On getting, if the element has a value attribute, it must return that
attribute's value; otherwise, it must return the string "on"

### DIFF
--- a/t/mojolicious/tag_helper_lite_app.t
+++ b/t/mojolicious/tag_helper_lite_app.t
@@ -24,6 +24,8 @@ get '/basicform';
 
 post '/text';
 
+post '/check_box_no_value';
+
 get '/multibox';
 
 get 'form/:test' => 'form';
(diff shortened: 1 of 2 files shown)


t/mojolicious/tag_helper_lite_app.t
```diff
@@ -209,6 +211,13 @@ $t->post_ok(
 EOF
 
 # Checkboxes
+$t->post_ok('/check_box_no_value' => form => { foo => 'on' }
+)->status_is(200)->content_is(<<EOF);
```
kraih: You didn't `perltidy` the code.

kraih:
That spec is for the browser API, and i don't think it applies here.

kraih:
To be more precise, i don't think it applies for generating input elements.

KES777:
The bug is when you have checked the checkbox and post data then the browser sends value 'on' for this <input>

Then when displaying this cechekbox you compare 'on' to '' so the checkbox displayed as unchecked but it should be checked. Have you looked at my testcase?

KES777:
Oh. Have not see your commit. Thanks for applying my notice.

PR #921, opened 2016-02-26
Preserve explicit 'content' in template

KES777:
if template which extends define its own content explicitly:
```
% extends 'base';




% content content => begin
TEST
%end
```

the content block will be unconditionally replaced by empty lines from this template.
Some output from debugger:
```
DBG>e $content
{ content => sub { ... } }
DBG>n
./lib/Mojolicious/Renderer.pm
105:   local $content->{content} = $output if $stash->{extends} || $stash->{layout};
DBG>e $content
{ content => "\n\n\n\n\n\n\n" }
```

https://github.com/kraih/mojo/commit/0f3f6d8b

### DIFF
--- a/lib/Mojolicious/Renderer.pm
+++ b/lib/Mojolicious/Renderer.pm
@@ -132,7 +132,8 @@ sub render {
 
   # Inheritance
   my $content = $stash->{'mojo.content'} ||= {};
-  local $content->{content} = $output if $stash->{extends} || $stash->{layout};
+  local $content->{content} = $content->{content} // $output
+    if $stash->{extends} || $stash->{layout};
   while ((my $next = _next($stash)) && !defined $inline) {
     @$options{qw(handler template)} = ($stash->{handler}, $next);
     $options->{format} = $stash->{format} || $self->default_format;
--- a/t/mojolicious/layouted_lite_app.t
+++ b/t/mojolicious/layouted_lite_app.t
@@ -51,6 +51,8 @@ get '/double_inheritance' =>
 
 get '/triple_inheritance';
 
+get '/triple_inheritance_explicit_content';
+
 get '/mixed_inheritance/first' => {template => 'first'};
 
 get '/mixed_inheritance/second' => {template => 'second', layout => 'green'};
@@ -197,6 +199,12 @@ $t->get_ok('/double_inheritance')->status_is(200)
 
 # Triple inheritance
 $t->get_ok('/triple_inheritance')->status_is(200)
+  ->header_is(Server => 'Mojolicious (Perl)')
+  ->content_is("<title>Works!</title>\n<br>\nSidebar too!\n"
+    . "New <content>.\nShared content!\n\nDefault footer!\n");
+
+# Triple inheritance with explicit content
+$t->get_ok('/triple_inheritance_explicit_content')->status_is(200)
   ->header_is(Server => 'Mojolicious (Perl)')
   ->content_is("<title>Works!</title>\n<br>\nSidebar too!\n"
     . "New <content>.\nShared content!\n\nDefault footer!\n");
@@ -391,6 +399,13 @@ Sidebar too!
 % extends 'double_inheritance';
 New <content>.
 %= content_for 'stuff'
+
+@@ triple_inheritance_explicit_content.html.ep
+% extends 'double_inheritance';
+% content content => begin
+New <content>.
+%= content_for 'stuff'
+% end
 
 @@ layouts/plugin_with_template.html.ep
 layout_with_template


kraih:
What is the use case for this?

KES777:
if not applied it violates ['New content will be ignored if the named buffer is already in use'](http://mojolicious.org/perldoc/Mojolicious/Plugin/DefaultHelpers#content)

The new content of empty lines replaces content of named buffer which is already in use.
This patch prevents that.

Defining 'content' buffer explicitly prevents flood of empty lines in result document.

in example above result will be
```
TEST
```
But if I define 'content' block implicitly
```
% extends 'base';




TEST
```
the generated document will be:
```





TEST
```

KES777:
Also in double inheritance without this patch the base content is not rendered if main template does not define 'content' (by default accidentally it will be flood of empty lines)
```
@@ base.html.ep
Hello

@@ main.html.ep
% extends 'base';



```
The rendering result will be just empty lines 
```




```
instead of expected 
```
Hello
```

kraih:
The code looks fine, but without a use case i have to vote :-1: i'm afraid.

KES777:
You mean documentation?

kraih:
We've talked a bit about this pull request on IRC, and it appears that nobody so far has been able to understand it. So i'm inclined to close it, unless you can provide a better description.

kraih:
Also, [this comment](https://github.com/kraih/mojo/pull/921#issuecomment-189539473) is incorrect, the result is `Hello`.

KES777:
hm.. yes. last comment is my fault(.
may be this example will be too synthetic, but I will try.
You may write templates strictly (like 'use strict' in perl) by defining your named blocks explicitly.
Here I mean in templates you use '% content content => begin... % end'

Then when you once decide to use 'extends' you should rewrite your templates without 'content' blocks because they stop to work.

KES777:
>sri: maybe there's big picture stuff i'm missing, but he didn't even try to explain it

Maybe this is too far away problem and is not applied for this simple inheritance that exists at mojolicious. The problem I have faced in when start to use inheritance:

```
@@ list.html.ep
% content header => begin
% end

% content footer => begin
% end

% content list => begin
% for my $value ( @{ stash->{ list } } ) {
% stash->{ _value } = $value;
  %= content 'value'
% }
% end

%= content 'header'
%= content 'list'
%= content 'footer'


@@ main.html.ep
% extends 'list';

% content 'value' => begin
%= stash->{ _value }
% end
```
So I created this patch https://github.com/kraih/mojo/pull/875 to resolve raised issues (see below).
I split my changes by many commits to show you what is coming on and why I do this or that change, but you recommend to squash history ((( to throw out useful info. Even more nobody vote because inheritance is 'hard' and no one interested in just using '%= content ... %= include'.

I think you should left this for later and do not force events because this feature not so popular but will bring complexity.

Issues inheritance has:
1. Content is rendered when you meet it
2. Content block does not support arguments
3. Content is rendered even when you just define block
4. Because of 1 you can not define a block at descendant and call it from parent

The real life code:
```
@@ short_info.html.ep
% extends 'table';

% content hpanel =>  begin
%= rec 'name';
<br>
% end

% content trow =>  begin
<tr><td>
%== '<br>' .rec( 'name' ) .': &euro;' .rec( 'price' )
% end

@@ table.htmle.ep
% extends 'list';



% content js =>  begin
% end



% content tid => begin
    trows
% end



% content table => begin
% my $rows =  shift;
<script type="text/javascript">
function addRow( id ){
    var table = document.getElementById(id);
    var newrow = table.insertRow(-1);
    newrow.innerHTML = table.rows[1].innerHTML;

    %= content 'js_body';
}
</script>
%= content 'js';

<table class="accounts" cellspacing="0" cellpadding="0" id=<%= content 'tid' %> >
    <thead>
        <tr>
            %= content 'thead';
        </tr>
    </thead>
    <tbody>
            %= content 'trows', $rows;
            %= content 'trow_last';
    </tbody>
</table>
% end

% content content =>  begin
    <div class="white_form border"><div class="make_order custumer_area_edit">
        % local stash->{ _block } =  'hpanel';
        %= content 'hpanel'
        %= content table => stash->{ rows }
        %= content 'fpanel'
    </div></div>
% end

@@ list.html.ep
% content hpanel =>  begin
% end

% content trow_last =>  begin
% end

% content trows =>  begin
% my $rows =  shift;
% $rows //=  $self->_query_rows;
% local stash->{ _block };
% while( my $row =  $rows->next ) {
    % stash->{ _block } =  { $row->get_columns };
    %= content 'trow', $row;
% }
%= content 'trow_last'
% end



% content fpanel =>  begin
% end


% content content =>  begin
    % local stash->{ _block } =  'hpanel';
    %= content 'hpanel'
    % local stash->{ _block } =  'trows';
    %= content 'trows'
    %= content 'fpanel'
% end

```
When first render this I was surprised to have empty page. The implicit content 'content' of short_info.html.ep replace the 'table.html.ep' content 'content'.
ok. I have fixed this issue by patch as in this PR. I have surprised second time because I can MUST create content 'content' explicitly. because if I left in 'table.html.ep' code as:
```
    <div class="white_form border"><div class="make_order custumer_area_edit">
        % local stash->{ _block } =  'hpanel';
        %= content 'hpanel'
        %= content table => stash->{ rows }
        %= content 'fpanel'
    </div></div>
```

the 'table' block will be called at compilation time of 'table.html.ep' that in turn will call 'trows',
But at this step 'trows' is not defined at all, because the 'list.html.ep' is not compiled yet.
So when I explicitly define `content 'content'` for block above I get two things:
1. I postpone rendering of 'table' until `%= content` expression 
2. I feel ground under foot ( I hate [this](https://github.com/kraih/mojo/blob/0f3f6d8bf779e8768e82920ea3d72863dfd8d442/lib/Mojolicious/Renderer.pm#L112)) brain breaking line. 
Instead of that hug better do not render empty lines at all and use simple checking:
$content->{ content } //= $output

kraih:
Afraid i don't think we can settle this without a real discussion, which would be really hard to have because of the language barrier. :cry: So, for now we might have to accept that we disagree on what template inheritance should be and move on.

KES777:
> What is the use case for this?

https://github.com/kraih/mojo/pull/922 
I think the "\s+" content should be fixed by '' (or even better by `undef`) so you will be not required to do this [check](https://github.com/kraih/mojo/blob/0f3f6d8bf779e8768e82920ea3d72863dfd8d442/lib/Mojolicious/Renderer.pm#L112) in the loop wasting the time

>  which would be really hard to have because of the language barrier

Stop repeating that. If you really want to fix the issue you will find a way or just ignore if don't.

PR #922, opened 2016-02-28
Test cases for 'content' data when templates are empty

KES777:
Test fail because of bug: different content is returned.

Please pay attention to tests when '' or "\n" is returned.


KES777:
First template: we have comment and one empty line but we got empty template ''. Strange but not critical, move on.

Second template have just comment and no lines in it so generated html should be ''. Success!

In third template I remove comment and expect '' as in first test case. But got 404 error 

In forth I just want to render empty page. I create file with no content and expect '' But also got 404 error

PR #924, opened 2016-03-01
Clarify K combinator

lindleyw:
Clarify that tap here means not "wiretap; eavesdrop; infiltrate" but "insert" and that "K Combinator" is a term in the relatively new Combinatory logic. Explain exactly what is going on for those new to these principles.

### DIFF
--- a/lib/Mojo/Base.pm
+++ b/lib/Mojo/Base.pm
@@ -225,9 +225,13 @@ pass it either a hash or a hash reference with attribute values.
   $object = $object->tap($method);
   $object = $object->tap($method, @args);
 
-K combinator, tap into a method chain to perform operations on an object within
-the chain. The object will be the first argument passed to the callback, and is
-also available as C<$_>.
+Tap into a method chain to perform operations on an object within the
+chain (also known as a K combinator). The object will be the first
+argument passed to the callback and is also available as C<$_>. The
+callback's return value will be ignored; instead, the object (the
+callback's first argument) will be the return value. In this way,
+arbitrary code can be used within (i.e., spliced or tapped into) a
+chained set of object method calls, which is a commonly used pattern.
 
   # Longer version
   $object = $object->tap(sub { $_->$method(@args) });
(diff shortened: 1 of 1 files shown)


s1037989:
:+1: Thank you!!  I have been meaning to pose this question for over a year because I have had *no* idea what `tap` means or what it's use is.  The synopsis did not help me either.

I created [this gist](https://gist.github.com/s1037989/e834d72f42a4dd82e6d0) just a few weeks ago to help me to understand it better, but your explanation for the docs is much appreciated!

s1037989:
Could you link "K combinator" to an online resource?  I remember searching the term over a year ago and came up with nothing.

Grinnz:
http://raganwald.com/JQuery-Combinators/ is the most relevant resource I can find, but it's not really generic enough to link to.

Grinnz:
Here is another, more ruby specific: http://combinators.info/#kestrels

lindleyw:
[Wikipedia's article](https://en.wikipedia.org/wiki/Combinatory_logic#Examples_of_combinators) seems quite typical in assuming quite a bit of knowledge of "lambda calculus," and other beyond-doctorate-level advanced mathematics. It's all very off-putting, even for those of us with degrees in Computer Science.

lib/Mojo/Base.pm
```diff
@@ -225,9 +225,13 @@ pass it either a hash or a hash reference with attribute values.
...
-K combinator, tap into a method chain to perform operations on an object within
-the chain. The object will be the first argument passed to the callback, and is
-also available as C<$_>.
+Tap into a method chain to perform operations on an object within the
+chain (also known as a K combinator). The object will be the first
+argument passed to the callback and is also available as C<$_>. The
+callback's return value will be ignored; instead, the object (the
+callback's first argument) will be the return value. In this way,
+arbitrary code can be used within (i.e., spliced or tapped into) a
+chained set of object method calls, which is a commonly used pattern.
```
kraih: The formatting is a little off, we use 80 characters per line.

kraih:
Note that we do not link to online resources for other concepts and patterns in the documentation. So, for consistency, a lot more would have to be updated as well.

kraih:
Otherwise :+1:.

lib/Mojo/Base.pm
kraih: This has come up on IRC, the "...which is a commonly used pattern" at the end, does not add much value to an otherwise good description.

PR #927, opened 2016-03-05
create named variable within Mojo::Template

s1037989:
This leaves much to be desired, but I wanted to run my technique by you and get any comments from you.  Please let me know what I can do to make this PR acceptable.  Thanks!!

### DIFF
--- a/lib/Mojo/Template.pm
+++ b/lib/Mojo/Template.pm
@@ -97,7 +97,7 @@ sub compile {
 }
 
 sub interpret {
-  my $self = shift;
+  my $self = $_[0];
 
   # Stack trace
   local $SIG{__DIE__} = sub {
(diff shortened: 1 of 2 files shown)


kraih:
Making it work in `Mojo::Template` is not the hard part, you have to make it work for `Mojolicious::Plugin::EPRenderer` too.

s1037989:
I'll look at that next if this is more or less acceptable?

kraih:
I don't think there should be a stash in `Mojo::Template`, an attribute accessor containing a hash does the job just as well.

kraih:
We generally don't do work in progress pull requests, if you want to discuss working branches you can do so on IRC or the mailing-list.

s1037989:
OK no problem. How should I best share my code for discussion? Just link
you to my fork?

kraih:
Yes.

kraih:
I would expect something like this. https://gist.github.com/anonymous/f1066c1027e965dd4e5a

PR #928, opened 2016-03-05
fix typo

s1037989:


kraih:
Thanks, but i think a comma is where we have to draw the line for pull requests. :wink:

s1037989:
Woops, sorry! Normally I just edit it online via Github when it's just a
typo. Doesn't that just create a PR too? Cuz you're not saying you don't
want simply typo fixes, correct?

kraih:
If it's a bunch of typo fixes that's ok, but we can't do pull requests for every comma.

s1037989:
No problem! I'll batch them up in the future! Which is somewhat awkward
because there are so few remaining...
:)

On Mar 4, 2016 11:38 PM, "Sebastian Riedel" <notifications@github.com>
wrote:
>
> If it's a bunch of typo fixes that's ok, but we can't do pull requests
for every comma.

PR #929, opened 2016-03-06
add context feature to Mojo::Template

s1037989:
I am jealous of your skills.  What I did worked but was not even close to the right approach!!!  Ugh!  Oh well, this helps me learn.  Thank you!

### DIFF
--- a/t/mojo/template.t
+++ b/t/mojo/template.t
@@ -26,6 +26,11 @@ my $mt     = Mojo::Template->new;
 my $output = $mt->render('');
 is $output, '', 'empty string';
 
+# Provide context
+$mt     = Mojo::Template->new;
+$output = $mt->context({test => 'TEST'})->render('<%= $test %>');
+is $output, "TEST\n", 'empty string';
+
 # Named template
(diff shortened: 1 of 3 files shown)


t/mojo/template.t
```diff
@@ -26,6 +26,11 @@ my $mt     = Mojo::Template->new;
 my $output = $mt->render('');
 is $output, '', 'empty string';
 
+# Provide context
+$mt     = Mojo::Template->new;
+$output = $mt->context({test => 'TEST'})->render('<%= $test %>');
+is $output, "TEST\n", 'empty string';
```
kraih: That is not en empty string. :wink:

kraih:
Not sure i like how the value of the `context` attribute is handled. There are a few quirks the documentation and tests do not cover. Like that it needs to be present when the template is first compiled, or that it needs to be reset every time the template is rendered. The whole thing is not very intuitive.

kraih:
This feature has been implemented a little differently. https://github.com/kraih/mojo/commit/af1c5e8aef965e6f634ca914c4b79d162067ecbe

PR #930, opened 2016-03-07
Handle multiple changed files at once for restart

nugged:
If you change numerous files at once or one-by-one in second(s), Morbo saw first file changed and tried to restart right away (though meanwhile restart process started, it can catch few files more on next "sleep 1"), but then after restart Morbo finds next one or next few changed files, not updated in cache, and restarts again until all cached files iterated.

That causes Morbo to restart many times if you, for example, do checkout for git branch where many files different or upload/sync pack of files through sftp at once, etc.

Changes in code: now changed files names packed in array in one pass, not returning after first met, and only returning that array. That updates cache for all files, then calls restart.

Also notification message printed to console with one file name if only one changed, that is for our usual work mode (when we uploaded one file from editor), or give number of changed files only if we do mass change (branch checkout, mass sync).

### DIFF
--- a/lib/Mojo/Server/Morbo.pm
+++ b/lib/Mojo/Server/Morbo.pm
@@ -13,8 +13,13 @@ has watch  => sub { [qw(lib templates)] };
 
 sub check {
   my $self = shift;
-  $self->_check($_) and return $_
-    for map { -f $_ && -r _ ? $_ : files $_ } @{$self->watch};
+  # pack all changed files names in one pass in array
+  # to prevent constant mojo "one-by-one" restart:
+  my @files_changed;
+  for (map { -f $_ && -r _ ? $_ : files $_ } @{$self->watch}) {
+    push @files_changed, $_ if $self->_check($_);
+  }
+  return \@files_changed if @files_changed;
   return undef;
 }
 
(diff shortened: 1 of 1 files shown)


kraih:
Good idea, but the patch has a few problems. You forgot to update the documentation, add tests, and broke a few existing tests in `t/mojo/morbo.t`.

kraih:
Not to mention that this is a breaking change, what are your plans for the deprecation process?

kraih:
Same effect could be achieved with a change that doesn't break backwards compatibility.
```diff
diff --git a/lib/Mojo/Server/Morbo.pm b/lib/Mojo/Server/Morbo.pm
index f06f804..7da4cc5 100644
--- a/lib/Mojo/Server/Morbo.pm
+++ b/lib/Mojo/Server/Morbo.pm
@@ -13,9 +13,9 @@ has watch  => sub { [qw(lib templates)] };
 
 sub check {
   my $self = shift;
-  $self->_check($_) and return $_
-    for map { -f $_ && -r _ ? $_ : files $_ } @{$self->watch};
-  return undef;
+  my @files = grep { $self->_check($_) }
+    map { -f $_ && -r _ ? $_ : files $_ } @{$self->watch};
+  return $files[0];
 }
 
 sub run {
```
But now the method becomes hard to test and we increase technical debt. Which is very bad for a module that's already hard to test.

leejo:
There's some tweaking here to address the above issues: https://github.com/kraih/mojo/compare/master...leejo:gh/930_handle_multiple_files_morbo_restart_SQUASHED

kraih:
Thanks, resolved by #931.

PR #940, opened 2016-03-24
Fix two test names to match the test result

andk:
Found two tests that had inverted names, as if there was a copy and paste and later the test was inverted or something.


kraih:
Thanks, applied.

PR #942, opened 2016-03-29
Add support for SSL_ca_path

cziebuhr:


lib/Mojo/IOLoop/Client.pm
```diff
@@ -171,14 +171,16 @@ sub _try_tls {
...
     SSL_cert_file  => $args->{tls_cert},
     SSL_error_trap => sub { $self->emit(error => $_[1]) },
     SSL_hostname   => IO::Socket::SSL->can_client_sni ? $args->{address} : '',
     SSL_key_file   => $args->{tls_key},
     SSL_startHandshake  => 0,
-    SSL_verify_mode     => $args->{tls_ca} ? 0x01 : 0x00,
+    SSL_verify_mode     => ($args->{tls_ca} || $args->{tls_ca_path}) ? 0x01 : 0x00,
     SSL_verifycn_name   => $args->{address},
-    SSL_verifycn_scheme => $args->{tls_ca} ? 'http' : undef
+    SSL_verifycn_scheme => ($args->{tls_ca} || $args->{tls_ca_path}) ? 'http' : undef
```
jhthorsen: Does it make sense to do the ```-d``` test above, when SSL_verify_mode and SSL_verifycn_name does not?

kraih:
A few things, **a)** You submitted this pull request without a description, so the motivation remains a mystery and even start a discussion, **b)** The patch has no tests, **c)** The attribute name `ca_path` is inconsistent with `ca`, `cert` and `key`.

cziebuhr:
> c) The attribute name ca_path is inconsistent with ca, cert and key.

Do you have a better name for it? I tried to be consistent with the SSL_ca_path option.

kraih:
@cziebuhr No, perhaps all related attributes should be renamed if there's going to be multiple variants of some. Now that you mention it, there's another inconsistency, `Mojo::Server::Daemon` would also need something like `ca_path` to be consistent with `Mojo::UserAgent`.

PR #945, opened 2016-04-08
address issue 944 Mojo::Loader::load_class()

kiwiroy:
Might be a better way, but I can't see it at the moment.


kraih:
Thanks, applied with a few changes.

PR #950, opened 2016-05-02
Fix a warning from Mojo::Exception->throw when $e->message is undef

autarch:
This is pathological, but it managed to occur in some $WORK code, and it's
worth avoiding warnings from the exception system.

### DIFF
--- a/Changes
+++ b/Changes
@@ -1,5 +1,7 @@
 
 6.62  2016-05-02
+  - Fixed a warning from Mojo::Exception->throw that could occur in
+    pathological cases where the exception message was undefined. (autarch)
 
 6.61  2016-04-30
   - Improved Mojo::Server::Daemon to no longer log when a connection has been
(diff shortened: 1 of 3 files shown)


jberger:
Can I ask how you end up with `undef` as your exception?

autarch:
I think we had a `$SIG{__DIE__}` handler catching something from some other code and trying to turn it into a Mojo::Exception. Definitely a bizarro corner case.

autarch:
Arguably, this is a problem with the class's constructor. Maybe it shouldn't allow `undef` as the message?

jhthorsen:
I would like to see that warning. It makes me wonder what i messed up on my side.

👎  from me.

kraih:
:-1: Think i have to agree with @jhthorsen, the warning actually showed a bizarro corner case, that's something positive in my book.

autarch:
I think in that case it would be preferable to have `Mojo::Exception`'s constructor die (or warn at least). As it is, figuring out what caused the `undef` message is very difficult. The current warning is incredibly unhelpful.

marcusramberg:
Guess I'm siding with @jhthorsen and @kraih on this one.. 👎

kraih:

PR #953, opened 2016-05-06
Few update lib/Mojolicious/resources/templates/mojo/debug.html.ep

mche:
highlight error row

### DIFF
--- a/lib/Mojolicious/resources/templates/mojo/debug.html.ep
+++ b/lib/Mojolicious/resources/templates/mojo/debug.html.ep
@@ -59,7 +59,11 @@
         color: #eee;
         text-shadow: #333 0 1px 0;
       }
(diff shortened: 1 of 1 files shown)


kraih:
I'm afraid without before/after screenshots i don't see how anyone would vote on this.

valchonedelchev:
Before:
https://dl.dropboxusercontent.com/u/97661837/Screen%20Shot%202016-06-08%20at%209.17.45%20AM.png

After:
https://dl.dropboxusercontent.com/u/97661837/Screen%20Shot%202016-06-08%20at%209.16.41%20AM.png

jhthorsen:
I think the red colour "pops" too much. Doesn't look it's in the same scheme as the rest. I wouldn't mind if the change was made, but I think a PR won't be the right way to get the change through. I think joining #mojo and discussing what you want to accomplish is a better way.

PR #955, opened 2016-05-09
Do not load FindBin until it is needed in Mojo::Home

autarch:
FindBin will die if $0 is not a file on disk. If, for example, you set $0 in
your tests to make debugging hanging tests easier, then even loading
Mojo::Home in the test suite causes an error, even if Mojo::Home would be able
to work without FindBin.

This was a real issue at $work and I can confirm that this PR fixes the problem. We load a Mojo app in our tests but we don't ever reach the code path that would need FindBin.

### DIFF
--- a/lib/Mojo/Home.pm
+++ b/lib/Mojo/Home.pm
@@ -5,7 +5,6 @@ use overload bool => sub {1}, '""' => sub { shift->to_string }, fallback => 1;
 use Cwd 'abs_path';
 use File::Basename 'dirname';
 use File::Spec::Functions qw(abs2rel catdir catfile splitdir);
-use FindBin;
 use Mojo::Util qw(class_to_path files);
 
 has parts => sub { [] };
@@ -27,6 +26,12 @@ sub detect {
     # Turn into absolute path
     return $self->parts([splitdir abs_path catdir(@home) || '.']);
   }
+
+  # We don't want to load this unconditionally. When FindBin is loaded it will
+  # die if $0 does not point to an actual file. Since we may not need FindBin
+  # at all, we can make using Mojo::Home safer by not loading it unless it's
+  # absolutely necessary.
+  require FindBin;
 
   # FindBin fallback
   return $self->parts([split '/', $FindBin::Bin]);
--- a/t/mojo/home_findbin.t
+++ b/t/mojo/home_findbin.t
@@ -1,0 +1,10 @@
+use Mojo::Base -strict;
+
+use Test::More;
+
+local $0 = 'not a path to a file';
+
+eval { require Mojo::Home };
+is $@, '', 'no exception loading Mojo::Home when $0 is not a file';
+
+done_testing();


jberger:
Sorry, can I get a little more detail? When would `$0` not be a file on disk? Is this a bug in FindBin? Also, by deferring loading FindBin to a point where it doesn't hurt you in this case does it not just become a heisenbug for someone for whom the deferred load does not fix the bug?

kraih:
👎 Have to agree with @jberger, deferring loading of `FindBin` seems like it would result in even harder to find bugs.

autarch:
It can be useful to set `$0` to something while code is running. In our case we have test code that runs a bunch of test classes. We've seen code hang while tests are running and it's difficult to figure out why it hung. I've now made Test::Class::Moose optionally update $0 as it goes to include test class names and test methods. That way if something hangs we can shell into the box running the tests and get a better idea of what exactly is hanging.

It's not really a bug in FindBin, although FindBin could make this less of an issue if it had an API beyond "set some package vars" by deferring the error to only when you ask FindBin for the info.

Contrary to what @kraih suggests, this will not result in harder to find bugs. Either way you'll get exactly the same error from FindBin. The only difference is that with my change there are many cases where you never get the error at all because FindBin isn't needed.

As an aside, while I appreciate the quick responses, the super quick _negative_ responses make for a somewhat unpleasant experience in contributing to this project.

jberger:
Could you set `$0` to a temporary file?

autarch:
@jberger I think that wouldn't work the purposes of Mojo::Home anyway, unless the temporary file was in the same directory as the Mojo script, right?

Setting the process name is a totally legit Unixy thing to do. FindBin is doing the best it can, but it's really not a great tool for use outside of scripts. Using it in a module like Mojo::Home is a bit dubious.

There might be a better way to get at this information, but I'm not 100% clear on how Mojo::Home is used. If the primary use case is for other Mojo modules to find the home directory, then perhaps the main entry point for Mojo apps could set something based on `cwd()` as early as possible?

autarch:
I should replace "home directory" with "project root", I guess.

jberger:
> If the primary use case is for other Mojo modules to find the home directory, then perhaps the main entry point for Mojo apps could set something based on cwd() as early as possible?

I like this notion generally. It would certainly take some thinking about.

jhthorsen:
I think the usecase is a bit odd, but the patch looks so simple. I'm giving it 👍

kraih:
👎 This patch is the wrong solution.

kraih:
The detection code in `Mojo::Home` was not designed, it (and related code in `Mojo::Server`) has simply grown over time and went through many iterations. It might not even be clear if `FindBin` support is needed at all anymore. So, i'd like to see some real research into the topic, and maybe some sort of home detection spec that explains all our use cases and makes sure we have them all tested.

kraih:
So, i'm afraid the patch as proposed here will not be applied.

oalders:
Could someone update this with a link to the new issue or PR once it's created?

jberger:
A new issue has been opened as #958 . Since this behavior has not been reported before and is in very old code we think we have some time to solve it right by investigating a better solution.

PR #956, opened 2016-05-11
Make Mojo::Template support Devel::Cover

ksmadsen:
Add support for Devel::Cover in Mojo::Template, by monkey patching the compiled
template into its own namespace. This way the previously-anonymous sub exists
in the symbol table, and therefore Devel::Cover can see it.

As suggested by @haarg in pjcj/Devel--Cover#31

A path attribute is also added, to provide the path to the template instead of
using the name, which doesn't identify the template file.

### DIFF
--- a/lib/Mojo/Template.pm
+++ b/lib/Mojo/Template.pm
@@ -18,6 +18,7 @@ has escape        => sub { \&Mojo::Util::xml_escape };
 has [qw(escape_mark expression_mark trim_mark)] => '=';
 has [qw(line_start replace_mark)] => '%';
 has name      => 'template';
+has path      => '';
 has namespace => 'Mojo::Template::SandBox';
 has tag_start => '<%';
 has tag_end   => '%>';
@@ -167,6 +168,7 @@ sub process {
     return Mojo::Exception->new($@)->inspect($self->unparsed, $code)
       ->trace->verbose(1)
       unless $compiled = eval $self->_wrap($code, @_);
+    monkey_patch $self->namespace, '_compiled', $compiled;
     $self->compiled($compiled);
   }
 
@@ -186,6 +188,7 @@ sub render { shift->parse(shift)->process(@_) }
 sub render_file {
   my ($self, $path) = (shift, shift);
 
+  $self->path($path) unless defined $self->{path};
   $self->name($path) unless defined $self->{name};
   my $template = slurp $path;
   my $encoding = $self->encoding;
@@ -256,9 +259,10 @@ sub _compile {
 }
 
 sub _line {
-  my $name = shift->name;
-  $name =~ y/"//d;
-  return qq{#line @{[shift]} "$name"};
+  my $self = shift;
+  my $path = $self->path || $self->name;
+  $path =~ y/"//d;
+  return qq{#line @{[shift]} "$path"};
 }
 
 sub _trim {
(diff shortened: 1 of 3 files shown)


jberger:
Generally I'm glad to see this work. As to this patch, I have a few questions. 

Is the path attribute necessary? Maybe I'm just not understanding but I don't see how that is related. (I also just woke up :p).

I would like to see the monkey patch protected by a compile time constant set from an environment variable. Putting all those subs in the namespace is eventually akin to a memory leak. However if I wanted I could set `MOJO_TEMPLATE_COVER=1` and only pay for it if I wanted.

ksmadsen:
I've added the path attribute, because the filename in the line directives before my patch are of the form:

```
#line 1 "template path/to/template.html.ep"
```

IOW, Devel::Cover will look for a file named `template path/to/template.html.ep`, which of course doesn't exist. (NYTProf also makes some noise about this).

We debated whether or not to just use the name attribute, and change the name generation in `Mojolicious::Plugin::EPLRenderer` instead. We chose to introduce a new attribute, in case the name was used for something we hadn't considered.

Making the monkey patching dependant on an environment variable is fine by me.

I'll happily update the patch with the environment variable stuff, and just use name as the path, if you wish me to.

kraih:
👎 The `path` attribute makes no sense, because the `name` attribute already exists.

ksmadsen:
Ok. I'll prepare a new patch, which eliminates the path attribute, and only monkey patches when the `MOJO_TEMPLATE_COVER` environment variable is true.

I'm not sure I understand how it doesn't take EP templates into account? It should enable `Devel::Cover` support for anything that uses `Mojo::Template`?

`DATA` section templates won't have the correct `#line` directive, but for those we'll need both the name of the file they come from and the line number in order where they start, for it to work correctly.

kraih:
If you make `Devel::Cover` support a feature, then i think `DATA` templates should be included.

ksmadsen:
Fair enough. However getting the line-number for the `DATA` templates seems to require substantial changes to `Mojo::Loader::_all`. I'll be happy to give it a shot, but I'm far from certain that I can match your coding style on what will basically need to be a rewrite of `_all`.

lib/Mojo/Template.pm
```diff
@@ -167,6 +168,7 @@ sub process {
     return Mojo::Exception->new($@)->inspect($self->unparsed, $code)
       ->trace->verbose(1)
       unless $compiled = eval $self->_wrap($code, @_);
+    monkey_patch $self->namespace, '_compiled', $compiled;
```
kraih: This is horrible code. Every time a new template gets compiled, the result is monkey patched into the shared namespace, redefining whatever was there previously. I'm assuming it is meant to assign a name to the otherwise anonymous coderef. That does not actually work reliably, because `monkey_patch` relies for this on `Sub::Util::set_subname`, which is only an optional dependency and not always available.

jberger:
Some background from @haarg on IRC #p5p: 

Devel::Cover do not require that subs have their sub name set. They need to exist in some form that Devel::Cover can find them.  this means either in the symbol table or in the optree (of something in the symbol table). That second case is why normal anon subs get coverage. A normal anon sub is linked from the optree, and then will be cloned when the code is run (assuming it's a closure). However because Mojo::Template compiles the sub on the fly in an eval it isn't in any optree and thus has no way to be found.

kraih:
So, there's a lot of open questions remaining. Perhaps this should be discussed on IRC or in a new issue (as a feature request perhaps?).

PR #959, opened 2016-05-12
callback shortcut

mche:
Hi!

### DIFF
--- a/lib/Mojolicious/Routes/Route.pm
+++ b/lib/Mojolicious/Routes/Route.pm
@@ -132,8 +132,13 @@ sub to {
 
   if ($shortcut) {
 
+    # callback
+    if (ref($shortcut) && ref($shortcut) eq 'CODE') {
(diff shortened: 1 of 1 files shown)


marcusramberg:
This pull request needs an actual description to be considered.

mche:
> an actual description

`$r->to(sub {...}) `

is same as

`$r->to(cb=>sub {...}) `

>  tests

Ok

mche:
huge amount tests in /t!
suggest me where to place test callback action.

jberger:
Why wouldn't you just use the hybrid route approach at this point? `$r->any('/path' => sub { ... })`

kraih:
👎 from me, since we already have enough alternatives.

jberger:
without a compelling reason not to use the hybrid routes I mentioned, I think I'm a 👎 as well, for the same reason as @kraih

mche:
Okay, not so important/significant

PR #961, opened 2016-05-17
Patch to disable layout for mojo/templates

mche:
### Hello

in cases
*  ` $app->defaults(layout=>'foo',)`
*   or `$c->render(..., layout=>...)`

i think we needs prevent rendering mojo/templates trought layout


kraih:
This pull request is very incomplete. http://mojolicious.org/perldoc/Mojolicious/Guides/Contributing#CONTRIBUTING-CODE

PR #964, opened 2016-05-26
fix substr outside of string in Mojo::Asset::Memory

leejo:
We updated to v6.62 (from v6.42) this morning and are seeing some odd behaviour, after a restart we sometimes get pages rendered as the raw HTML as if the headers are being sent through twice. We're still investigating this.

The one thing i do notice is `substr outside of string at lib/perl5/site_perl/5.16.3/Mojo/Asset/Memory.pm line 41` warnings since the update. I've included a PR here to fix this issue. The important point is that a substr outside of a string will return undef and if that is being fed into `length` then unexpected things might be happening, observe:

```
> perl -E 'say length(substr("foo",4))'

> perl -E 'say length(substr("foo",3))'
0
```

### DIFF
--- a/lib/Mojo/Asset/Memory.pm
+++ b/lib/Mojo/Asset/Memory.pm
@@ -38,7 +38,8 @@ sub get_chunk {
     $max = $end + 1 - $offset if ($offset + $max) > $end;
   }
 
-  return substr shift->{content} // '', $offset, $max;
+  my $len = length($self->{content} //= '');
+  return substr $self->{content}, $offset > $len ? $len : $offset, $max;
 }
 
 sub move_to {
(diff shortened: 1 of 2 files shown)


kraih:
Thanks for your effort, but i'm afraid this pull request did not pass the vote.

mattp-:
seeing this as well, unsure what the cause is. if there is a reason this was rejected, can you help explain how to stop this warning from occurring?

kraih:
@mattp- This was submitted as a new feature, instead of a bug fix. Which means it needs a certain number of votes to be accepted, and that it did not reach.

mattp-:
ah ok. I actually just realized what was causing this for any future people who stumble upon this kind of issue:
I was rendering twice by calling $c->reply->not_found, but not actually returning in the controller method itself, so a second render was occurring with null/undef being returned. thanks for the quick response.

PR #978, opened 2016-07-01
add an example of a json service

jberger:
### Summary
Interestingly, the Request's json method does not appear in any of the guides. This PR adds an example that uses it and other common json service functionality

### Motivation
There does appear to be some confusion as to how to get the json representation of a request as evidenced by the google search results for "Mojolicious parse JSON body". Having a small example in the Tutorial seems like a simple place to start. I have chosen to place it after the GET and POST parameters section and the HTTP section since you need the concept of input parameters and also knowledge of the `->req` and `->res` methods to establish these concepts. The `json` stash value doesn't seem to be mentioned directly either but is rather introduced obliquely later, this also gives a nod in that direction.

### References
https://groups.google.com/forum/#!topic/mojolicious/_r-tzqKKtGI
https://github.com/kraih/mojo/issues/522#issuecomment-229823654

### DIFF
--- a/lib/Mojolicious/Guides/Tutorial.pod
+++ b/lib/Mojolicious/Guides/Tutorial.pod
@@ -168,6 +168,33 @@ L<Mojolicious::Command::get>.
 
   $ ./myapp.pl get -v -M POST -c 'test' /echo
 
+=head2 JSON Services
+
+JSON is a very common interchange format for web services. Mojolicious loves
+JSON and makes it very easy to use everywhere. In this example you can fetch
+module data from L<MetaCPAN|http://metacpan.org>.
+
+  use Mojolicious::Lite;
+
+  post '/module_info' => sub {
+    my $c = shift;
+    my $module = $c->req->json->{module};
+    my $info = $c->ua->get("http://api.metacpan.org/module/$module")->res->json;
+    $c->render(json => $info);
+  };
+
+  app->start;
+
+Now let's see the current distribution that contains L<Mojo>.
+
+  $ ./myapp.pl get -M POST -c '{"module":"Mojo"}' /module_info /release
+
+That trailing path is a L<"JSON pointer"|Mojo::JSON::Pointer> which helps dive
+into JSON documents quickly. The example above might have used one to get the
+module from the request.
+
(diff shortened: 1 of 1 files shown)


lib/Mojolicious/Guides/Tutorial.pod
```diff
@@ -168,6 +168,33 @@ L<Mojolicious::Command::get>.
 
   $ ./myapp.pl get -v -M POST -c 'test' /echo
 
+=head2 JSON Services
+
+JSON is a very common interchange format for web services. Mojolicious loves
+JSON and makes it very easy to both build and use JSON services. In this example
+you can fetch module data from L<MetaCPAN|http://metacpan.org>.
```
marcusramberg: Maybe it would be better to write 'In this example we fetch module data from ...'?

kraih:
Afraid i have to give this a 👎 , since the example is out of place and introduces features that are meant to be introduced much later in the tutorial.

dancingfrog:
Thank you @jberger, I agree that:
1) Methods that deal with parsing the body of requests (especially JSON) should be appended to the current Mojolicious documentation and
2) Perhaps the example is out of place given the current structure of the tutorial. However, I don't think there's anything sacred about the tutorial or the general flow of the documentation, which I actually find kind of difficult to navigate from the the perspective of a reference guide.
 
Sorry for the apparent (yet, unintended) "snark" in previous comments.

jberger:
The tutorial is not intended as a reference guide. Indeed the problem is that the request's json methos is only described in the class reference documentation. I'm trying to correct the fact that it does not appear in the Guides which is where we want people to be introduced to the concepts.

kraih:
Closing this pull request for now. While most of us still believe that an example similar to this one should exist in the guides, @dancingfrog's behavior here and on other communication channels has resulted in an exodus of volunteers willing to work on it.

jberger:
For the record, a similar tutorial was added in https://github.com/kraih/mojo/compare/cebbec8c18ae8dd1d9803f92e0d63616f0244e46...56bb6015d80b8d221630ef99af0dbac1ad363013

PR #979, opened 2016-07-05
Attribute Selector fails for defined attributes with wrong values in Mojo::DOM::CSS

Akron:
Given a tag with a defined but value-less attribute like

``` html
<input type="text" name="example" value>
```

(for example generated using the taghelper plugin with ```$c->text_field('example', value => undef)```),
the CSS selector

``` CSS
input[value="myname"]
```

will match regardless of the non-fitting value.
I tested the behaviour in Firefox and assume the selector should not match.
I added a test and a patch (although I doubt the patch is optimal).

Thank you!


kraih:
Thanks, applied.

PR #986, opened 2016-08-10
websocket_proxy.t: fix warning on perl 5.10.1

dolmen:
### Summary
Fix an issue in the testsuite (t/mojo/websocket_proxy.t): the `$server` variable contains either an object or `undef`. Using `defined` instead of `length` is the right thing (whatever perl version).

### Motivation
The original code provokes a warning when running the testsuite on perl 5.10.1.

### DIFF
--- a/examples/connect-proxy.pl
+++ b/examples/connect-proxy.pl
@@ -14,7 +14,7 @@ Mojo::IOLoop->server(
 
         # Write chunk from client to server
         my $server = $buffer{$id}{connection};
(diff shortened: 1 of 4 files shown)


kraih:
Agree with the change, but that line of code exists in multiple places.

dolmen:
OK. I'll update the PR with the fixes for the other tests.

kraih:
Think an example script is affected too.

dolmen:
Commit updated.

jberger:
thanks!

PR #989, opened 2016-08-15
close DATA to prevent strange addendum in subsequent error messages

jberger:
### Summary
As reported on IRC (and indeed I've seen it too, just not had the inclination to look) there is a message appended to error messages saying ", at DATA line 2231".

### Motivation
Closing the DATA handle after using it in Mojo::Util suppresses this addendum.

### References
http://irclog.perlgeek.de/mojo/2016-08-15#i_13025043


kraih:
Doesn't the `DATA` section code in `Mojo::Loader` still cause the same kind of warnings?

PR #990, opened 2016-08-16
reset $. removes addendum

kiwiroy:
### Summary
Further to the data-line-in-error branch, adding a solution and test.

### Motivation
In preparing this a small issue with `Symbol::delete_package` (called by `Mojo::Util::_teardown`) was discovered, though it appears to be already known as an overzealous approach to deleting everything, modules included, which leads to a few extra line changes in this patch.

### References
#989


kraih:
Thanks, using `$.` was a great idea.

PR #994, opened 2016-08-25
Do not pass globals as arguments. They may be clobbered

KES777:
### Summary
Error is clobbered before we process it

### Motivation
Arguments to functions are just aliases. When we process error we should copy $@ as soon as possible. When we pass $@ as argument we may implicitly loose its value

### References

sub test {
	$@ =  'Oops';
	my( $x ) =  @_;
	print $x; # Oops
}
$@ =  'Exception';
test( $@ );


KES777:
sorry, miss my repository

Issue #995, opened 2016-09-08
Test::Mojo::json_is does not work correctly with the $desc in Unicode


asan13:
* Mojolicious version: 5.54
* Perl version: 5.20.2
* Operating system: Debian 8.5 (x86_64) 

### Steps to reproduce the behavior
```perl
use ut8;
use open qw(:std :utf8);

my $t = Test::Mojo->new('MyApp');
   
$t  -> get_ok('foo')
    -> json_is('/xxx' => 'yyy', 'Комментарий');
```
_Output_
```
    ok 1 - GET http://localhost:36491/test/foo
    ok 2 - ÐÐ¾Ð¼Ð¼ÐµÐ½ÑÐ°ÑÐ¸Ð¹
```
### Expected behavior
```
    ok 1 - GET http://localhost:36491/test/foo
    ok 2 - 'Комментарий'
```
### Actual behavior
Breaks description of the test, if it is in Unicode.

The reason is that the 'json_is' _always_ makes 'encode' for the parametr '$desc'. And it breaks the output if used _use open qw(:std :utf8);_
```perl
sub json_is {
  my $self = shift;
  my ($p, $data) = @_ > 1 ? (shift, shift) : ('', shift);
  my $desc = encode 'UTF-8', shift || qq{exact match for JSON Pointer "$p"}; # <<<==========
  return $self->_test('is_deeply', $self->tx->res->json($p), $data, $desc);
}
```
In similar functions done differently, and there is no problem. For example, json_like:
```perl
sub json_like {
  my ($self, $p, $regex, $desc) = @_;
  $desc ||= encode 'UTF-8', qq{similar match for JSON Pointer "$p"}; # <<<========
```

kraih:
How to handle Unicode output on `STDERR` and `STDOUT` in Mojolicious has been an open question for years now. So far it seemed like the safes choice not to mess with Perl defaults, therefore our expectation is that you don't do either. Of course we are open to proposals for changing this, but they would have to take the whole framework into account.

kraih:
Also, please feel free to discuss this on IRC, a few of the many problems with possible proposals have already been brought up there today.

asan13:
It is just that the implementation of the 'json_is' did not differ from the rest 'json_ | content_' etc. It seems enough:

```perl
sub json_is {
  my $self = shift;
  my ($p, $data) = @_ > 1 ? (shift, shift) : ('', shift);
[...]

PR #997, opened 2016-09-14
Fix utf8 comments for test

rshadow:
### Summary
Fix utf8 comments for test

### Motivation
Compromise for `use utf8; use open ':std :utf8'` users and mojo.
Everybody gets what they want.

### References
#995
8fdf806ff56520bd79dfe0fe8fc24315918bfd37

### DIFF
--- a/lib/Test/Mojo.pm
+++ b/lib/Test/Mojo.pm
@@ -323,7 +323,7 @@ sub _build_ok {
   return $self->_request_ok($self->ua->build_tx($method, $url, @_), $url);
 }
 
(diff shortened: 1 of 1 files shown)


kraih:
Definitive 👎 from me, because this is what running the tests would look like. https://travis-ci.org/kraih/mojo/jobs/159861250

rshadow:
You right. We need a big patch for migration mojo to utf8 support.

kraih:
Please try to understand the problem completely before submitting more pull requests. It is very complicated and there will be no simple solution.

rshadow:
OK. I go to IRC.

PR #998, opened 2016-09-15
Update Morbo.pm

Arseny-N:
Fixed typo.

### DIFF
--- a/lib/Mojo/Server/Morbo.pm
+++ b/lib/Mojo/Server/Morbo.pm
@@ -16,7 +16,7 @@ sub modified_files {
 
   my $cache = $self->{cache} ||= {};
   my @files;
-  for my $file (map { -f $_ && -r _ ? $_ : files $_ } @{$self->watch}) {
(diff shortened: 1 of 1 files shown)


kraih:
I'm afraid that is not a typo.

jhthorsen:
Would changing it to ```map { -f -r $_ ? $_ : files $_ }``` make it less confusing?

@Arseny-N: You can read more about ```_``` here: https://metacpan.org/pod/distribution/perl/pod/perlfunc.pod#X-FILEHANDLE (a couple of paragraphs down)

jberger:
Funny I was about to reply that you needed a newer perl for that but ...
https://metacpan.org/pod/perl5100delta#Stacked-filetest-operators

Arseny-N:
Oh! Thanks for the links.   @jhthorsen, yes it would be less confusing since it will not look like a typo, but if it will introduce some dependency on the perl version, then I think a  comment, containing a googlable phrase  would be good enough. 
After a second thought it seems stupid to comment a language feature, so ...

Issue #999, opened 2016-09-18
Mojo::UserAgent::Transactor should redirect 303 to GET for all request methods


jberger:
* Mojolicious version: 7.06
* Perl version: ALL
* Operating system: ALL

### Steps to reproduce the behavior
Build a simple app that redirects from a `DELETE` (or any non-`POST`) back to a `GET`. Be sure to use a 303 response code.
``` bash
$ perl -Mojo -E 'any "/" => { text => "hello" }; del "/delete" => sub { $_->res->code(303); $_->redirect_to("/") }; app->start' get -r -v -M DELETE /delete
```

### Expected behavior
Per https://tools.ietf.org/html/rfc7231#section-6.4.4 (a more plain reading version is on wikipedia https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#3xx_Redirection) seems to suggest that the `GET` method type used for getting the new location mentioned by a 303 response. 

### Actual behavior
Mojo::UserAgent does in fact redirect to the new resource but it continues to use a `DELETE` method.

jberger:
Note that I have confirmed that Chrome handles it as `DELETE` -> 303 -> `GET` as well, so there is browser precedent here too.

PR #1000, opened 2016-09-18
handle all 303 redirects with a GET method

jberger:
As detailed in #999, 303 redirects should be handled with a `GET` method. This PR implements that behavior.


jberger:
I had some success at golfing this down to 0 lines but I needed to make `$method` just a bit shorter :-P. If it comes down to that, let's discuss.

jhthorsen:
LGTM 👍

PR #1009, opened 2016-10-17
fix for multiple Location: headers in response

jkramer:
### Summary
Fix/workaround for servers that return multiple Location headers.

### Motivation
Make Mojo::UA work with servers that return multiple Location headers.

### References
When a server returns multiple Location headers (which it shouldn't, but some do), Mojo::UserAgent redirects to a URL that is made of the multiple URLs in the headers concatenated. Example:

MOJO_USERAGENT_DEBUG=1 perl -MMojo::UserAgent -e 'Mojo::UserAgent->new->max_redirects(2)->get("http://www.lvz.de/Leipzig/Polizeiticker/Polizeiticker-Leipzig/Amokdrohungen-gegen-vier-Leipziger-Schulen-Polizei-sichert-Gebaeude")'

This fix simply uses the first URL for the redirect.

### DIFF
--- a/lib/Mojo/UserAgent/Transactor.pm
+++ b/lib/Mojo/UserAgent/Transactor.pm
@@ -76,7 +76,7 @@ sub redirect {
   return undef if uc $req->method eq 'CONNECT';
 
   # Fix location without authority and/or scheme
-  return undef unless my $location = $res->headers->location;
+  return undef unless my $location = $res->headers->to_hash(1)->{Location}->[0];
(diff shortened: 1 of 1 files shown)


jhthorsen:
Thanks for the PR.

For this PR to be considered, there must at least be a test, and preferably some documentation.

kraih:
Thanks, fixed. https://github.com/kraih/mojo/commit/01a86c9f2bace0a92c71f4b1c8cf0da8272176e3

PR #1018, opened 2016-11-28
Fix unexpected websocket handshake failure degrading into infinite loop

CandyAngel:
### Summary
The test suite (reasonably) presumes that websocket handshaking will succeed during testing. When it does not, the test suite gets (vocally) stuck in Mojo::IOLoop indefinitely. This change handles this edge case correctly, just resulting in failed tests.

The only known instance of this failure is when proxied HTTP connections are intercepted by local network security software and cause the handshaking to fail. This makes the behaviour unlikely to occur and may not be worth the cost to the codebase.

### Motivation
The test suite failing into an infinite loop would appear to be unacceptable behaviour outside of unmitigatable failure such as malfunctioning RAM.

### References
This PR has an [alternative PR](https://github.com/kraih/mojo/pull/1019) which implements the same fix in a different way (minimises line count).

### DIFF
--- a/t/mojo/websocket_proxy.t
+++ b/t/mojo/websocket_proxy.t
@@ -160,7 +160,7 @@ is $result, 'http://example.com/proxy', 'right content';
 $ua->websocket(
   "ws://127.0.0.1:$port/test" => sub {
     my ($ua, $tx) = @_;
-    Mojo::IOLoop->stop and return unless $tx->is_websocket;
+    return Mojo::IOLoop->stop unless $tx->is_websocket;
     $kept_alive = $tx->kept_alive;
     $tx->on(finish => sub { Mojo::IOLoop->stop });
     $tx->on(message => sub { shift->finish; $result = shift });
@@ -183,7 +183,7 @@ $result = undef;
 $ua->websocket(
   "ws://127.0.0.1:$port/test" => sub {
     my ($ua, $tx) = @_;
-    Mojo::IOLoop->stop and return unless $tx->is_websocket;
+    return Mojo::IOLoop->stop unless $tx->is_websocket;
     $tx->on(finish => sub { Mojo::IOLoop->stop });
     $tx->on(message => sub { shift->finish; $result = shift });
     $tx->send('test1');
@@ -201,10 +201,9 @@ my ($success, $leak, $err);
 $ua->websocket(
   "ws://127.0.0.1:$dummy/test" => sub {
     my ($ua, $tx) = @_;
-    Mojo::IOLoop->stop and return unless $tx->is_websocket;
     $success = $tx->success;
-    $leak    = !!Mojo::IOLoop->stream($tx->previous->connection);
     $err     = $tx->error;
+    $leak    = !!Mojo::IOLoop->stream($tx->previous->connection) if $tx->is_websocket;
     Mojo::IOLoop->stop;
   }
 );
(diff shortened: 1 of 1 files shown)


t/mojo/websocket_proxy.t
```diff
@@ -160,6 +160,7 @@ is $result, 'http://example.com/proxy', 'right content';
 $ua->websocket(
   "ws://127.0.0.1:$port/test" => sub {
     my ($ua, $tx) = @_;
+    Mojo::IOLoop->stop and return unless $tx->is_websocket;
```
kraih: The return value of `Mojo::IOLoop->stop` is undefined.

kraih:
Afraid i have to vote 👎, the described scenario is just too much of a fringe case.

jberger:
Would there be any reasonable way to bail out of the test early if the OS breaks websockets? Some quick check near the top of the test script?

CandyAngel:
You could put in a test that checks that websocket tx's are being returned for proxied requests at the start and BAIL_OUT if not.

```
# Test websocket handshakes over proxy
$ua->proxy->http("http://127.0.0.1:$port");
my $got_websocket = undef;
$ua->get(
  'http://example.com/proxy' => sub {
    my ($ua, $tx) = @_;
    $got_websocket = $tx->is_websocket;
    Mojo::IOLoop->stop;
  }
);
Mojo::IOLoop->start;
ok $got_websocket, 'Websocket gotten' or BAIL_OUT('Proxied websocket handshaking failed');
```

As the concern is getting stuck in IOLoop, could IOLoop can have a timeout instead? This would have the benefit of protecting against other cases where you can get stuck in IOLoop, but would also require all instances of Mojo::IOLoop->stop to be replaced with Mojo::IOLoop->reset? Something like this:

```
# Proxy WebSocket
$ua = Mojo::UserAgent->new;
$ua->proxy->http("http://127.0.0.1:$proxy");
my $result = undef;
$ua->websocket(
  "ws://127.0.0.1:$port/test" => sub {
    my ($ua, $tx) = @_;
    $tx->on(finish => sub { Mojo::IOLoop->reset });
    $tx->on(message => sub { shift->finish; $result = shift });
    $tx->send('test1');
  }
);
_ioloop_timeout(5);
is $connected, "127.0.0.1:$port", 'connected';
is $result,    'test1test2',      'right result';
ok $read > 25, 'read enough';
ok $sent > 25, 'sent enough';

sub _ioloop_timeout {
  Mojo::IOLoop->timer(5 => sub { shift->reset });
  Mojo::IOLoop->start;
}
```
Or if it would be useful elsewhere, IOLoop could track the timer id and remove it whenever the loop is stopped, so you could just do this:
```
Mojo::IOLoop->start(timeout => 5);
```
But running the IOLoop for (up to) a certain amount of seconds seems to be pretty specific to preventing getting stuck in testing.

CandyAngel:
It may be worth mentioning that one of the rules governing Mojolicious is this:

> The installation process should be as fast and painless as possible. (Less than a minute on most common hardware is a good rule of thumb)

As the installation process usually involves running the test suite _which may fail_ (after all, that is the point of running it before installing), a failing test suite should also be as fast and painless as possible. An infinite loop caused by the lack of a check would seem to violate this rule.

With this in mind, should I raise an issue to deal with this in a general sense, so it isn't tied to this specific PR?

kraih:
Afraid it looks like this pull request did not pass the vote.

PR #1019, opened 2016-11-28
Fix unexpected websocket handshake failure degrading into infinite loop

CandyAngel:
### References
This PR has an [alternative PR](https://github.com/kraih/mojo/pull/1018) which implements the same fix in a different way (less risk of side effects, clearer code [in my opinion]).

### DIFF
--- a/t/mojo/websocket_proxy.t
+++ b/t/mojo/websocket_proxy.t
@@ -163,7 +163,7 @@ $ua->websocket(
     $kept_alive = $tx->kept_alive;
     $tx->on(finish => sub { Mojo::IOLoop->stop });
     $tx->on(message => sub { shift->finish; $result = shift });
-    $tx->send('test1');
+    ($tx->is_websocket) ? $tx->send('test1') : Mojo::IOLoop->stop;
   }
 );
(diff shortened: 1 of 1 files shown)


t/mojo/websocket_proxy.t
```diff
@@ -199,6 +199,7 @@ my ($success, $leak, $err);
 $ua->websocket(
   "ws://127.0.0.1:$dummy/test" => sub {
     my ($ua, $tx) = @_;
+    Mojo::IOLoop->stop and return unless $tx->is_websocket;
```
kraih: The return value of `Mojo::IOLoop->stop` is undefined.

kraih:
Same as #1018, i have to vote 👎.

kraih:
Maybe worth mentioning too, i find the fact that there are two competing pull requests very confusing, and the descriptions don't make a clear case for why either one should be picked.

kraih:
If you are not confident enough to pick one solution, perhaps it was too early to send a pull request.

CandyAngel:
Closing this as I think #1018 is the better implementation.

PR #1021, opened 2016-12-01
Add extra documentation re: placement of layouts.

lloy0076:
### Summary
Add documentation making it obvious "layouts" by default must lie underneath the configured template paths.

### Motivation
Assuming that "layouts" by default lived in MyApp/layouts as opposed to MyApp/templates/layouts it took me a few goes to figure out where to put the layouts.

### DIFF
--- a/lib/Mojolicious/Guides/Rendering.pod
+++ b/lib/Mojolicious/Guides/Rendering.pod
@@ -42,6 +42,10 @@ All templates should be in the C<templates> directories of the application,
 which can be customized with L<Mojolicious::Renderer/"paths">, or one of the
 the C<DATA> sections from L<Mojolicious::Renderer/"classes">.
 
+Any layouts should be directly under one of the C<templates> directories of the
+application (for example C<templates/layouts/example.layout.html.ep>) or one of
+the C<DATA> sections from L<Mojolicious::Rendered/"classes">.
+
   __DATA__
 
   @@ time.html.ep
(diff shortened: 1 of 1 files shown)


lib/Mojolicious/Guides/Rendering.pod
```diff
@@ -42,6 +42,10 @@ All templates should be in the C<templates> directories of the application,
 which can be customized with L<Mojolicious::Renderer/"paths">, or one of the
 the C<DATA> sections from L<Mojolicious::Renderer/"classes">.
 
+Any layouts should be directly under one of the C<templates> directories of the
+application (for example C<templates/layouts/example.layout.html.ep>) or one of
+the C<DATA> sections from L<Mojolicious::Rendered/"classes">.
```
jberger: I was about to say the same thing. Also `example.layout.html.ep` is a bad name because it makes it look like `.layout.` is magical somehow. Later examples use `mylayout.html.ep`

lib/Mojolicious/Guides/Rendering.pod
```diff
@@ -571,6 +575,10 @@ value).
 
   my $html = $c->render_to_string('reminder', layout => 'mail');
 
+Note that when searching for a file based layout, the C<layout> helper will
```
jberger: I'd say something like 'when a layout is set' since you can set it via a helper or a stash key.

lib/Mojolicious/Guides/Rendering.pod
lloy0076: I wasn't quite sure where to put this and put it in Rendering.pod mostly for lack of a better place.

As another project owner commented, "moving documentation to the right place is the easy part" - it's the writing that is not.

I'm wondering if anyone had a better suggested place to put something like this?

jberger: I think I'm saying that the second note (below) is sufficient.

kraih:
Have to vote 👎 on this pull request, since the problems brought up have not been addressed.

kraih:
I'm afraid this didn't pass the vote.

Issue #1022, opened 2016-12-19
Mojo::DOM method text should work on text-nodes


wollmers:
* Mojolicious version: 7.11, Doughnut
* Perl version: v5.20.1, darwin
* Operating system: OS X El Capitan Version 10.11.6

### Steps to reproduce the behavior
```
$ perl -MMojo::DOM -E 'say q{"},Mojo::DOM->new("&amp;")->child_nodes->[0]->text,q{"};'
""
```

### Expected behavior
```
$ perl -MMojo::DOM -E 'say q{"},Mojo::DOM->new("&amp;")->child_nodes->[0]->tree->[1],q{"};'
"&"
```

Of course the work-around `->tree->[1]`can be used, but that's not an official API and can break in future.

### Actual behavior
The method `text` on nodes of type `text`, `raw`and `cdata` returns an empty string instead of the text.

The method `to_string`works but is entity-encoded and thus does not allow direct processing as plain text.

jberger:
The underlying cause of the confusion here is that `content` is documented to work on nodes while `text` is documented to work on elements. There are several ways to proceed here:

* close, saying that the documentation already indicates this
* go slightly further and cross link the documentation for `text` to `content` since people may look in the wrong place for this ("to extract text from a node, see L</content>")
* change text to DTRT on a node element

kraih:
If there was enough interest i guess i'd support making `->text` work on all nodes, otherwise i'd say keep everything the way it is.

wollmers:
@jberger Thx for the work-around using `content`.

As I understand the docs, `content` is intended for extracting and setting XML/HTML fragments, as the docs state:

```
Return this node's content or replace it with HTML/XML fragment (for root and tag nodes) or raw content.
```

BTW I wouldn't expect this difference:

```
$ perl -MMojo::DOM -E 'say q{"},Mojo::DOM->new("&amp;")->child_nodes->[0]->content,q{"};'
"&"

$ perl -MMojo::DOM -E 'say q{"},Mojo::DOM->new("<a>&amp;</a>")->at("a")->content,q{"};'
"&amp;"
```

Can we agree that it would be more intuitive to use `->text` to extract text from nodes of type `text`, `cdata`and `raw`? Or should we agree to disagree?

In other parsers like XML::Twig `->text` works on all nodes as expected, i. e. returning plain Perl strings.

wollmers:
@kraih IMHO returning an empty string is wrong in any case.

I can provide a pull request this weekend.

[...]

PR #1024, opened 2016-12-25
text method in Mojo::Dom works now also on nodes of type text, raw and cdata

wollmers:
### Summary
Allows to process the text of nodes directly.

### Motivation
In traversing a parsed tree the text nodes also need processing as text.

### References
#1022

### DIFF
--- a/t/mojo/dom.t
+++ b/t/mojo/dom.t
@@ -293,7 +293,7 @@ ok !$dom->child_nodes->first->matches('*'), 'no match';
 is_deeply $dom->child_nodes->first->attr, {}, 'no attributes';
 is $dom->child_nodes->first->namespace, undef, 'no namespace';
 is $dom->child_nodes->first->tag,       undef, 'no tag';
-is $dom->child_nodes->first->text,      '',    'no text';
+is $dom->child_nodes->first->text,      'foo',    'text';
 is $dom->child_nodes->first->all_text,  '',    'no text';
 
 # Class and ID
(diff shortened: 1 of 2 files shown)


jberger:
The code can likely be golfed a little, but assuming this is the direction we want to go (allow `text` to act on nodes as well as elements) then 👍

kraih:
I'll have to give this a 👎 i'm afraid, because of the inconsistency between `text` and `all_text`, as well as the size of the code.

t/mojo/dom.t
```diff
@@ -2520,4 +2520,24 @@ $dom = Mojo::DOM->new($huge);
 is $dom->all_text, 'works', 'right text';
 is "$dom", $huge, 'right result';
 
+# text
```
kraih: This comment is not consistent with the rest.

t/mojo/dom.t
```diff
@@ -293,7 +293,7 @@ ok !$dom->child_nodes->first->matches('*'), 'no match';
 is_deeply $dom->child_nodes->first->attr, {}, 'no attributes';
 is $dom->child_nodes->first->namespace, undef, 'no namespace';
 is $dom->child_nodes->first->tag,       undef, 'no tag';
-is $dom->child_nodes->first->text,      '',    'no text';
+is $dom->child_nodes->first->text,      'foo',    'text';
```
kraih: This test no longer belongs here.

jhthorsen:
All the tests are still OK with this:

```
 sub text {
   my $tree = shift->tree;
-  my $type = $tree->[0];
-  if ($type eq 'text' || $type eq 'cdata' || $type eq 'raw') {
-    return $tree->[1];
-  }
-  _text([_nodes($tree)], 0);
+  _text([$tree, _nodes($tree)], 0);
 }
```

I guess the same change could be applied to `all_text()`, but with unit tests added.

kraih:
Afraid it looks like this pull request did not pass the vote.

PR #1025, opened 2016-12-27
prevent warnings whenever morbo checks broken symlinks

Grinnz:
### Summary
Currently, if broken symlinks exist in the directories watched by morbo, `stat` will return undefined values for that file causing unnecessary caching and warnings. Anything which does not return defined values can simply be skipped by morbo.

### Motivation
Reduce computation and avoid warnings if watched directories have broken symlinks.

### References
https://irclog.perlgeek.de/mojo/2016-12-27#i_13808575


jberger:
👍

kraih:
Not sure if we need this, but the code looks good.

jhthorsen:
I've experienced the same thing. I'm 👍 on this change.

PR #1028, opened 2017-01-07
add a Mojo::File module

kraih:
### Summary
This patch adds a `Mojo::File` module to encapsulate most of the code we use to interact with file systems on different operating systems. Its scope is currently limited to things we need and use in Mojolicious itself.

While i was a bit sceptical at first about the value it will add to the project. After seeing the improvements it brought to tests and documentation, i think it has the potential to make a lot of code look quite a bit cleaner.

There is a small performance cost, due to a lot of `Mojo::File` objects getting instantiated and method call overhead. But all tests so far suggest that this is negligible.

### Motivation
There's a lot of `File::*` modules we use in Mojolicious (`File::Basename`, `File::Spec`, `File::Find`, `File::Copy`, `File::Path`...). Their use is very decentralized at the moment, even though we've already written our own little extensions, like `files()`, `slurp()` and `spurt()`, which we had to put into `Mojo::Util`. So i think it might make sense to re-evaluate this approach.

### References
https://groups.google.com/d/msg/mojolicious/8ld7jmH-qbg/wO2OCkCsEQAJ

### DIFF
--- a/lib/Mojo/File.pm
+++ b/lib/Mojo/File.pm
@@ -1,0 +1,380 @@
+package Mojo::File;
+use Mojo::Base -strict;
+use overload
+  '@{}'    => sub { shift->to_array },
+  bool     => sub {1},
+  '""'     => sub { ${$_[0]} },
+  fallback => 1;
+
+use Carp 'croak';
+use Cwd qw(abs_path getcwd);
+use Exporter 'import';
+use File::Basename ();
+use File::Copy     ();
+use File::Find     ();
+use File::Path     ();
+use File::Spec;
+use File::Temp ();
+use Mojo::Collection;
+use Scalar::Util 'blessed';
+
+our @EXPORT_OK = ('path', 'tempdir');
+
+sub basename { scalar File::Basename::basename ${$_[0]}, @_ }
+
+sub child {
+  my $self = shift;
+  return $self->new($self, @_);
+}
+
+sub dirname { $_[0]->new(scalar File::Basename::dirname ${$_[0]}) }
+
+sub is_abs { File::Spec->file_name_is_absolute(${shift()}) }
+
+sub list {
+  my ($self, $options) = (shift, shift // {});
+
+  return Mojo::Collection->new unless -d $$self;
+  opendir(my $dir, $$self) or croak qq{Can't open directory "$$self": $!};
+  my @files = grep { $_ ne '.' && $_ ne '..' } readdir $dir;
+  @files = grep { !/^\./ } @files unless $options->{hidden};
+  @files = map { File::Spec->catfile($$self, $_) } @files;
+  @files = grep { !-d } @files unless $options->{dir};
+
+  return Mojo::Collection->new(map { $self->new($_) } sort @files);
+}
+
+sub list_tree {
+  my ($self, $options) = (shift, shift // {});
+
+  # This may break in the future, but is worth it for performance
+  local $File::Find::skip_pattern = qr/^\./ unless $options->{hidden};
+
+  my %files;
+  my $want = sub { $files{$File::Find::name}++ };
+  my $post = sub { delete $files{$File::Find::dir} };
+  File::Find::find {wanted => $want, postprocess => $post, no_chdir => 1},
+    $$self
+    if -d $$self;
+
+  return Mojo::Collection->new(map { $self->new($_) } sort keys %files);
+}
+
+sub make_path {
+  my $self = shift;
+  File::Path::make_path $$self, @_
+    or croak qq{Can't make directory "$$self": $!};
+  return $self;
+}
+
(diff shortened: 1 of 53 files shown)


jberger:
I just reread. I really like how this new class unifies a lot of previously spread functionality. :+1:

hertzhaft:
Something I missed in Perl for a long time. Makes Mojo an ever more useful toolkit.

lib/Mojo/Util.pm
```diff
@@ -56,13 +55,16 @@ my %CACHE;
 
 our @EXPORT_OK = (
   qw(b64_decode b64_encode camelize class_to_file class_to_path decamelize),
-  qw(decode deprecated dumper encode files getopt hmac_sha1_sum html_unescape),
+  qw(decode deprecated dumper encode getopt hmac_sha1_sum html_unescape),
   qw(md5_bytes md5_sum monkey_patch punycode_decode punycode_encode quote),
-  qw(secure_compare sha1_bytes sha1_sum slurp split_cookie_header),
-  qw(split_header spurt steady_time tablify term_escape trim unindent unquote),
-  qw(url_escape url_unescape xml_escape xor_encode)
```
jhthorsen: Should the functions be removed from EXPORT_OK, before they are removed from this module?

kraih: Look a few lines down. 😉

jhthorsen: Oh! I missed out on that. Thanks 👍

scottw:
+1; I would make use of Mojo::File in the same way I make use of Mojo::Util, Mojo::JSON, and Mojo::UserAgent even in non-Mojolicious (server) projects.

kraih:
Quick update, i've added a rather big change, turning `Mojo::Home` into a subclass of `Mojo::File`. https://github.com/kraih/mojo/commit/f67d11cfe475ec09139e6cfc9544ed42074d6244

lammel:
After review of the changes I must say the code looks a lot cleaner and reads easier (seems to have more Mojo). 
Although the overhead introduced is a downside (through "method" calls on the Mojo::File object), I really think this issue needs to be tackled by Perl itself.

So... +1

lib/Mojo/File.pm
```diff
@@ -0,0 +1,377 @@
...
+Include hidden files and directories.
+
+=back
+
+=head2 make_path
+
+  $path = $path->make_path;
+
+Create the directories if they don't already exist with L<File::Path>.
+
```
oetiker: looking at the code, it seems that 

```perl
$path->make_path({opts})
```

according to http://perldoc.perl.org/File/Path.html would also be supported ... 

maybe mention this ...

kraih: That's intentional, i'm not sure yet how we'll support those options. So it will stay unofficial for now.

kraih: A clean proposal with documentation and tests could convince me to support it though.

kraih:
I've also tried to benchmark and optimize a few common uses from the Mojolicious codebase, and used `Path::Tiny` for reference. So far it looks pretty good, but maybe some of you have more ideas.
```
$ perl -Ilib -MMojo::File -MPath::Tiny -Mojo -E 'n { Path::Tiny::path("/foo/bar") } 100000; n { Mojo::File::path("/foo/bar") } 100000'
0.166467 wallclock secs ( 0.17 usr +  0.00 sys =  0.17 CPU) @ 588235.29/s (n=100000)
0.0755291 wallclock secs ( 0.08 usr +  0.00 sys =  0.08 CPU) @ 1250000.00/s (n=100000)
```
```
perl -Ilib -MMojo::File -MPath::Tiny -Mojo -E 'n { Path::Tiny::path("/foo/bar")->child("baz", "yada") } 100000; n { Mojo::File::path("/foo/bar")->child("baz", "yada") } 100000'
0.475247 wallclock secs ( 0.48 usr +  0.00 sys =  0.48 CPU) @ 208333.33/s (n=100000)
0.412924 wallclock secs ( 0.41 usr +  0.00 sys =  0.41 CPU) @ 243902.44/s (n=100000)
```
```
$ perl -Ilib -MMojo::File -MPath::Tiny -Mojo -E 'n { Path::Tiny::path("/Users/sri/.vimrc")->slurp_raw } 100000; n { Mojo::File::path("/Users/sri/.vimrc")->slurp } 100000'
2.36471 wallclock secs ( 1.73 usr +  0.63 sys =  2.36 CPU) @ 42372.88/s (n=100000)
1.22802 wallclock secs ( 0.70 usr +  0.53 sys =  1.23 CPU) @ 81300.81/s (n=100000)
```

kraih:
Thanks everyone, the pull request passed the vote. 😄

PR #1036, opened 2017-01-13
fix "split to @_ is deprecated" warning

preaction:
On Perl 5.10, perl is incorrectly interpreting any split not assigned to
an array as a split intended to replace @_. Since this behavior was
deprecated, Perl is warning about it. This warning is causing tests in
downstream modules (like my Mercury application) which check for
warnings to fail.

I can't find when this deprecation was added (before 5.6), but the
feature was removed from Perl in 5.12, and the warning fixed to only
apply to split called in void context.

The only syntax that prevents this warning is assigning the result of
split to an array. Disabling the category of this warning ("syntax", as
reported in perldiag) does not seem to prevent this warning, but
disabling all warnings does.

### DIFF
--- a/lib/Mojo/Home.pm
+++ b/lib/Mojo/Home.pm
@@ -13,7 +13,8 @@ sub detect {
   # Location of the application class (Windows mixes backslash and slash)
   elsif ($class && (my $path = $INC{my $file = class_to_path $class})) {
     $home = Mojo::File->new($path)->to_array;
-    splice @$home, split('/', $file) * -1;
+    my @file_parts = split '/', $file;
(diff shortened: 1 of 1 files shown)


kraih:
Testing for warnings unconditionally is a very bad idea imo, and definitely not reason enough to add workarounds to Mojolicious. We generally don't specifically optimize for 5.10.1 either, deprecation warnings (even false ones) are a bit of an odd problem, so i'm not sure if the current workaround will stay. I'd hate for this to result in a wave of pull requests with more workarounds for old versions of Perl.

PR #1045, opened 2017-01-30
Expose dom as an attribute on Test::Mojo

marcusramberg:
### Summary
Expose dom as an attribute on Test::Mojo to allow testing Mojo::DOM objects without involving a user agent. Also makes some tests in t/mojo/mojo-dom.t a bit simpler to read.

### Motivation
This makes it easy to use Test::Mojo to test the contents of HTML/XML strings.

### DIFF
--- a/lib/Test/Mojo.pm
+++ b/lib/Test/Mojo.pm
@@ -17,7 +17,8 @@ use Mojo::Util qw(decode encode);
 use Test::More ();
 
 has [qw(message success tx)];
-has ua => sub { Mojo::UserAgent->new->ioloop(Mojo::IOLoop->singleton) };
+has ua  => sub { Mojo::UserAgent->new->ioloop(Mojo::IOLoop->singleton) };
+has dom => sub { shift->tx->res->dom };
 
 # Silent or loud tests
 $ENV{MOJO_LOG_LEVEL} ||= $ENV{HARNESS_IS_VERBOSE} ? 'debug' : 'fatal';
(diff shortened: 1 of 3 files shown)


jberger:
I've wanted this on the off occasion where there are mixed api return types, like a JSON api for which one of the values is HTML.

jberger:
I'm less sure about changing how the `t/mojo/dom.t` tests test Mojo::DOM (now via Test::Mojo) but the rest I'm mildly 👍  on

lib/Test/Mojo.pm
```diff
@@ -17,7 +17,8 @@ use Mojo::Util qw(decode encode);
 use Test::More ();
 
 has [qw(message success tx)];
-has ua => sub { Mojo::UserAgent->new->ioloop(Mojo::IOLoop->singleton) };
+has ua  => sub { Mojo::UserAgent->new->ioloop(Mojo::IOLoop->singleton) };
+has dom => sub { shift->tx->res->dom };
```
kraih: Is an attribute really needed here? How much faster is it than a method?

jberger: I don't think it is intended as a shortcut but rather as an assignable DOM.

marcusramberg: Yeah, the idea is that you can use Test::Mojo dom methods without doing a request by assigning a Mojo::DOM object, as illustrated by the changes in t/mojo/dom.t

kraih:
👎 Really don't like the changes in `dom.t`. If you want better test coverage for `Test::Mojo`, then i believe the right way to do it would be a new test file with tests for pass and fail. Not sure how to do it best, but i'm sure other test modules have tests too.

marcusramberg:
Ok, I'm definitively not attached to the t/mojo/dom.t changes, will add a
few tests to t/test-mojo.t or so instead. Closing this pull request and
will reopen when I'm done.

M

jberger:
FYI, once Perl ships a major (minor, sigh) version with Test2, I intend to rewrite the Test::Mojo internals and test them better (with approval of course).

PR #1046, opened 2017-01-30
Testing Mojo::DOM with Test::Mojo

marcusramberg:
### Summary
Expose dom as an attribute on Test::Mojo to allow testing Mojo::DOM objects without involving a user agent. Revised version of #1045 with new tests instead of changing t/mojo/dom.t

```
my $t=Test::Mojo->new();
$t->dom(Mojo::DOM->new('<foo>bar</foo>'));
$t->is_text('foo','bar');
```

### Motivation
This makes it easy to use Test::Mojo to test the contents of HTML/XML strings.

### DIFF
--- a/lib/Test/Mojo.pm
+++ b/lib/Test/Mojo.pm
@@ -16,6 +16,7 @@ use Mojo::UserAgent;
 use Mojo::Util qw(decode encode);
 use Test::More ();
 
+has dom => sub { shift->tx->res->dom };
 has [qw(message success tx)];
 has ua => sub { Mojo::UserAgent->new->ioloop(Mojo::IOLoop->singleton) };
 
@@ -85,7 +86,7 @@ sub delete_ok { shift->_build_ok(DELETE => @_) }
 
 sub element_count_is {
   my ($self, $selector, $count, $desc) = @_;
-  my $size = $self->tx->res->dom->find($selector)->size;
+  my $size = $self->dom->find($selector)->size;
   return $self->_test('is', $size, $count,
     _desc($desc, qq{element count for selector "$selector"}));
 }
@@ -93,13 +94,13 @@ sub element_count_is {
 sub element_exists {
   my ($self, $selector, $desc) = @_;
   $desc = _desc($desc, qq{element for selector "$selector" exists});
-  return $self->_test('ok', $self->tx->res->dom->at($selector), $desc);
+  return $self->_test('ok', $self->dom->at($selector), $desc);
 }
 
 sub element_exists_not {
   my ($self, $selector, $desc) = @_;
   $desc = _desc($desc, qq{no element for selector "$selector"});
-  return $self->_test('ok', !$self->tx->res->dom->at($selector), $desc);
+  return $self->_test('ok', !$self->dom->at($selector), $desc);
 }
 
 sub finish_ok {
@@ -371,6 +372,7 @@ sub _request_ok {
     my $desc = _desc("WebSocket handshake with $url");
     return $self->_test('ok', $self->tx->is_websocket, $desc);
   }
+  delete $self->{dom};
 
   # Perform request
   $self->tx($self->ua->start($tx));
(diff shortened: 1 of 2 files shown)


t/mojolicious/lib/TestApp.pm
```diff
@@ -0,0 +1,10 @@
+package TestApp;
```
kraih: This class doesn't really add much value, does it? Couldn't you just as well have used `Mojolicious::Lite` in the test file?

kraih:
Looks like your pull request is borked, it contains my last commit.

marcusramberg:
Fixed the extra commit, and moved the test app into the t/test-mojo.t file - Will rebase and force push if the pull request is accepted.

kraih:
Move it into `t/mojo/test.t`.

Changes
```diff
@@ -1,5 +1,6 @@
 
 7.24  2017-01-30
+  - Expose dom as an attribute on Test::Mojo.
```
kraih: No need for a `Changes` entry, i'll rewrite it anyway. 😉

lib/Test/Mojo.pm
```diff
@@ -17,7 +17,8 @@ use Mojo::Util qw(decode encode);
 use Test::More ();
 
 has [qw(message success tx)];
-has ua => sub { Mojo::UserAgent->new->ioloop(Mojo::IOLoop->singleton) };
+has ua  => sub { Mojo::UserAgent->new->ioloop(Mojo::IOLoop->singleton) };
+has dom => sub { shift->tx->res->dom };
```
kraih: Order is not alphabetical.

lib/Test/Mojo.pm
```diff
@@ -450,6 +452,13 @@ C<HARNESS_IS_VERBOSE> environment variable.
 
 L<Test::Mojo> implements the following attributes.
 
+=head2 dom
+
+  my $dom = $t->dom;
+  $t     = $t->dom(Mojo::DOM->new);
```
kraih: Indentation is borked here.

t/test-mojo.t
```diff
@@ -0,0 +1,23 @@
+
+use Mojo::Base -strict;
+
+BEGIN {
+  $ENV{MOJO_MODE}    = 'testing';
```
kraih: The mode does not appear to be used anywhere in the test.

t/test-mojo.t
```diff
@@ -0,0 +1,23 @@
...
+  $ENV{MOJO_REACTOR} = 'Mojo::Reactor::Poll';
+}
+use FindBin;
```
kraih: `FindBin` does not appear to used anywhere.

nicomen:
Long sought-after and clean enough. Good work ;)

kraih:
I'd really like to hear more about how you're planning to use this.

jberger:
My use has usually been with mixed type apis. So if there's say a json api that has a key who's value is HTML. This happens rather commonly with backend-rendered content for a single page front-end.

```perl
$t->get('/some/resource/with/html.json')
  ... # json tests

$t->dom(Mojo::DOM->new($t->tx->res->json('/contains/html')));
$t->... # html tests
```

at least, that's how I've wanted it in the past.

kraih:
@jhthorsen, @amenonsen and @tempire still have to review.

jhthorsen:
I'm neutral. There are times when I want to poke into the "res" for manual testing or debugging, but I don't think the "dom" attribute is special for my cases.

kraih:
@jhthorsen Please do a formal review, so we can see the result "Reviewers" list at the top.

kraih:
Guess i'm slightly 👎 on this as well, something just doesn't feel right about it.

kraih:
I'm afraid this pull request did not pass the vote as it ended in a tie, which means i had to make the final decision. What ultimately made me decide against it, was the fact that the feature can be easily implemented as a `Test::Mojo::Role`, and that there was only a single rather uncommon use case presented.

Issue #1048, opened 2017-02-02
Make Mojo::File::make_path/remove_tree with arguments a feature


kraih:
Just opening this issue as a reminder that we still need documentation and tests for `Mojo::File::make_path` and `Mojo::File::remove_tree` with arguments (which are passed through to `File::Path::make_path` and `File::Path::remove_tree`).

kraih:
Resolved by #1049.

PR #1049, opened 2017-02-03
add doc on passing arguments and tests to check pass though to File::…

kiwiroy:
### Summary
Documentation and tests for arguments passed through from `Mojo::File::make_path` and `Mojo::File::remove_tree` to `File::Path`.

### Motivation
🇳🇿 contribution

### References
#1048

### DIFF
--- a/lib/Mojo/File.pm
+++ b/lib/Mojo/File.pm
@@ -281,8 +281,10 @@ Include hidden files and directories.
 =head2 make_path
 
   $path = $path->make_path;
+  $path = $path->make_path({mode => 0711});
 
-Create the directories if they don't already exist with L<File::Path>.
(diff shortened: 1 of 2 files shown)


lib/Mojo/File.pm
```diff
@@ -281,8 +281,10 @@ Include hidden files and directories.
 =head2 make_path
 
   $path = $path->make_path;
+  $path = $path->make_path({mode => 0711});
 
-Create the directories if they don't already exist with L<File::Path>.
+Create the directories if they don't already exist with L<File::Path>. Any
+additional arguments are passed through unmodified to L<File::Path::make_path>.
```
kraih: The `L<File::Path::make_path>` doesn't look right to me, does that actually lead anywhere?

PR #1051, opened 2017-02-08
Avoid utf8 warnings for non-utf8 files in Mojo::Exception::inspect

pmorch:
Instead source files with illegal utf8 chars are ignored as if they
were unreadable. See these two discussions:

https://groups.google.com/d/msg/mojolicious/55jAXEbWzG0/kPY4BEjJBQAJ
https://groups.google.com/d/msg/mojolicious/6c-ZavT2KzQ/vTsJO2E6RQkJ

### Summary
If there were exceptions in source files that happen to have non-UTF-8 characters Mojo::Exception generated UTF-8 related warnings, even when these source files are completely legal perl.

### Motivation
The proposed commit makes no changes whatsoever to correctly UTF-8-encoded files, and does the simplest possible for non-UTF8 files - ignores them completely.

### References
https://groups.google.com/d/msg/mojolicious/55jAXEbWzG0/kPY4BEjJBQAJ
https://groups.google.com/d/msg/mojolicious/6c-ZavT2KzQ/vTsJO2E6RQkJ

### DIFF
--- a/lib/Mojo/Exception.pm
+++ b/lib/Mojo/Exception.pm
@@ -20,7 +20,15 @@ sub inspect {
   # Search for context in files
   for my $file (@files) {
     next unless -r $file->[0] && open my $handle, '<:utf8', $file->[0];
-    $self->_context($file->[1], [[<$handle>]]);
+    # If there are UTF-8 problems in the source file, don't store any context
+    my @lines = eval {
+      use warnings 'FATAL' => 'utf8';
+      <$handle>;
+    };
+    if ($@) {
+      next;
+    }
+    $self->_context($file->[1], [\@lines]);
     return $self;
   }
 


jberger:
In order for this to be considered, we'd first need to have a discussion of why/how to support non-ASCII/UTF-8 encodings. Templates are at the moment assumed to be UTF-8 (and they implicitly `use utf8` at the top of the template. The content that is read into the exception context really needs to by characters (not bytes) because this exception object might be serialized back into bytes later and double encoding should be avoided.

Anyway, all of that is moot because this cannot be considered because, by policy, features must come with documentation and tests, and the style of the code is not consistent with surrounding style.

pmorch:
## About the major point

> we first need to have a discussion of why/how to support non-ASCII/UTF-8 encodings

I realize that this is the most important of your objections. I'm going on a skiing vacation tomorrow (oh, yeah!) so I'll take that to the Google group when I'm back. I do understand that this is the major point.

I just want to be clear here that this is not about an issue with non-UTF-8 _templates_. My templates _are_ UTF-8. This is an issue where `Mojo::Exception` generates UTF-8 related-warnings when exceptions occur in _external modules_ whose source happens not to be valid UTF-8, even if such exceptions occur inside `eval`-s. (See test case at the bottom)

## About the other objections

I'd like to ask if we could side-step that more substantial objection for a moment, as I humbly have a few questions about other points:

> The content that is read into the exception context really needs to by characters (not bytes) because this exception object might be serialized back into bytes later and double encoding should be avoided.

How are any bytes introduced by this PR? Sources that are UTF-8 continue to be characters exactly as before, but sources where conversion fails because they contain illegal UTF-8 sequences are completely ignored, so still no non-character data can make it into the exception context. If all sources are UTF-8, _nothing at all_ changes; UTF-8 characters are maintained. Did I I miss something?

> the style of the code is not consistent with surrounding style

I've tried to follow [Mojolicious::Guides::Contributing - Contributing to Mojolicious](http://mojolicious.org/perldoc/Mojolicious/Guides/Contributing#CONTRIBUTING-CODE). I ran it through the `perltidy -pro=.perltidyrc lib/Mojo/Exception.pm` which added an empty line before the comment as the only diff to my patch. If that is objectionable I'll re-add that empty line. Are there other style-related problems did I miss? Is it that I have written a couple of short lines, where the surrounding code has longer lines? I'll happily change the style, but it will save us all time if I know what to fix.

> features must come with documentation and tests

About the documentation for ignoring non-UTF-8 sources: I'm happy to contribute documentation, but where? I've looked for documentation in three places: The guides, [`Mojo::Exception`](http://mojolicious.org/perldoc/Mojo/Exception) itself and in the wiki. I couldn't find any documentation for the requirement for sources to be  UTF-8 in any of the three sources, and so I didn't know where to write documentation for this either. Should I add a short mention to `Mojo::Exception` of the requirement of UTF-8-ness of sources and that all non-UTF-8 encoded sources are simply ignored? It doesn't look to me as if that would be consistent with the surrounding documentation and I'm affraid to be in violation of:

> The main focus of the included documentation should be on examples, no walls of text. (An example for every one or two sentences is a good rule of thumb)

I'll write tests, no problem. One of the tests; the one I was bitten by, had essentially this:

    #!/usr/bin/perl -w
    use strict;

    use Mojo::Template;
    use HTML::PrettyPrinter;

    my $mt = Mojo::Template->new;

    sub method {
      # Call some code in some other module whose source that happens to be
      # non-UTF-8. The eval could be here or elsewhere, e.g. in the non-UTF-8 module.
      eval {
        # HTML::PrettyPrinter dies on undef input
        HTML::PrettyPrinter->new()->format(undef);
      };
      return $@ ? 'error' : 'ok';
    }

    say $mt->render(<<'EOF', \&method);
    % my ($method) = @_;
    <div>
      <%= $method->() %>
    </div>
    EOF

PR #1057, opened 2017-02-16
Accept SameSite cookie attribute.

heikojansen:
Also allow it to be set for the session cookie.

https://tools.ietf.org/html/draft-ietf-httpbis-cookie-same-site-00

Only supported in some of the important browsers as of now (cf. http://caniuse.com/#feat=same-site-cookie-attribute). I believe it eventually will be supported by all, but I have no hard evidence on that.
If you consider it a premature addition maybe keep the pull request around in a branch?

### DIFF
--- a/lib/Mojo/Cookie/Response.pm
+++ b/lib/Mojo/Cookie/Response.pm
@@ -4,9 +4,9 @@ use Mojo::Base 'Mojo::Cookie';
 use Mojo::Date;
 use Mojo::Util qw(quote split_cookie_header);
 
-has [qw(domain expires httponly max_age origin path secure)];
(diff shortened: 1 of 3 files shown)


kraih:
Don't think we will support this before it hasn't been accepted as an RFC and is supported by more browsers.

kraih:
I'm closing this issue for now, since we've had too many bad experiences with supporting features based on drafts. But that does not mean it has been rejected, just that it's too early for a discussion.

heikojansen:
Apparently this is _still_ not part of RFC 6265 (or a successor) but according to http://caniuse.com/#feat=same-site-cookie-attribute all major browsers support this now.
Maybe it is time to reopen this pull request?

PR #1062, opened 2017-02-27
Added the config mechanism in the generate app command

tudorconstantin:
### Summary
Generate the config file and enable the config plugin with the `mojo generate app` command in a full blown Mojo app.

### Motivation
I think that the vast majority of web apps end up needing the config mechanism. So, instead of having the developers looking through the docs how to enable the plugin and use it, provide it to them automatically at create time.

### References
There are no open issues (or other pull requests for this).


Grinnz:
I don't think it's a good idea to set secrets by default (even ones obviously intended to be changed). That bypasses the effectiveness of the warning.

kraih:
Agreed, disabling the secret warning is a total deal breaker for me.

PR #1063, opened 2017-02-28
The config mechanism in the app generated with generate app command

tudorconstantin:
### Summary
Generate the config file and enable the config plugin with the `mojo generate app` command in the full blown Mojo app.

### Motivation

### References
This is my second PR for this small improvement,[ the first ](https://github.com/kraih/mojo/pull/1062) being rejected because the sample regarding how to use the config file disabled the warning for the default secrets.

### DIFF
--- a/lib/Mojolicious/Command/generate/app.pm
+++ b/lib/Mojolicious/Command/generate/app.pm
@@ -25,6 +25,9 @@ EOF
   my $app = class_to_path $class;
   $self->render_to_rel_file('appclass', "$name/lib/$app", $class);
 
+  # Config file
+  $self->render_to_rel_file('config', "$name/$name.conf");
+
   # Controller
   my $controller = "${class}::Controller::Example";
   my $path       = class_to_path $controller;
@@ -136,6 +139,12 @@ sub startup {
   # Documentation browser under "/perldoc"
   $self->plugin('PODRenderer');
 
+  # Config mechanism
+  $self->plugin('Config');
+
+  # $config holds the hashref returned by the config file
+  my $config = $self->app->config;
+
   # Router
   my $r = $self->routes;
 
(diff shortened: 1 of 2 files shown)


lib/Mojolicious/Command/generate/app.pm
```diff
@@ -202,3 +211,12 @@ and the layout "templates/layouts/default.html.ep",
 <%%= link_to 'here' => '/index.html' %> to move forward to a static page. To
 learn more, you can also browse through the documentation
 <%%= link_to 'here' => '/perldoc' %>.
+
+@@ config
+{
+  db => {
+      dsn => "...",
+      username => "...",
+      password => "...",
```
kraih: That's not how we configure database connections in the Mojolicious world. (https://github.com/kraih/mojo-pg/blob/master/examples/blog/blog.conf)

lib/Mojolicious/Command/generate/app.pm
```diff
@@ -202,3 +211,12 @@ and the layout "templates/layouts/default.html.ep",
...
+    },
+};
```
kraih: We also don't leave out the newline at the end.

lib/Mojolicious/Command/generate/app.pm
```diff
@@ -136,6 +139,12 @@ sub startup {
   # Documentation browser under "/perldoc"
   $self->plugin('PODRenderer');
 
+  # Config mechanism
+  $self->plugin('Config');
```
kraih: The comment should be more educational, like `# Load configuration from hash returned by "my_app.conf"`.

lib/Mojolicious/Command/generate/app.pm
```diff
@@ -136,6 +139,12 @@ sub startup {
...
+
+  # $config holds the hashref returned by the config file
+  my $config = $self->app->config;
```
kraih: This is not entirely correct.

kraih:
Afraid i have to reject this pull request again, since there's too many questionable points. I strongly suggest discussing the next patch on IRC first.

PR #1064, opened 2017-03-01
added the config mechanism in the app generated with generate app command

tudorconstantin:
### Summary
Generate the config file and enable the config plugin with the mojo generate app command in the full blown Mojo app.

### Motivation

### References
I hope I incorporated all the observations made to the previous PRs by sri and Grinnz


kraih:
Looks like you didn't actually try this one, because it doesn't work.

PR #1066, opened 2017-03-04
update document for load_class function

joez:
### Summary
Update document for Mojo::Loader::load_class function, because the return value is mismatch with the source code

### Motivation
I am reading the Mojo source code to learn such elegant framework, and find this document mismatch with code, would like to update it if my understanding is correct


kraih:
Thanks, applied.

joez:
it's my pleasure

PR #1067, opened 2017-03-05
The config mechanism in the app generated with generate app command

tudorconstantin:
### Summary
Generate the config file and enable the config plugin with the mojo generate app command in the full blown Mojo app.

### Motivation

### References
All the rejection reasons of the previous PRs, plus the discussions on IRC


kraih:
Thanks, applied.

tudorconstantin:
thank you for your patience and guidance :smile:

PR #1070, opened 2017-03-14
morbo learns --sleep to set the poll period of watch

leejo:
### Summary
Add `--sleep` to morbo that will be used by lib/Mojo/Server/Morbo.pm in the sleep call, will default to the previous value of 1 second.

### Motivation
If you have many watched files then morbo can sit on your CPU more than you would like, setting the sleep value to > 1 second eases this somewhat.

### References
N/A

### DIFF
--- a/lib/Mojo/Server/Morbo.pm
+++ b/lib/Mojo/Server/Morbo.pm
@@ -10,6 +10,7 @@ use POSIX 'WNOHANG';
 
 has daemon => sub { Mojo::Server::Daemon->new };
 has watch  => sub { [qw(lib templates)] };
+has sleep  => sub { 1 };
 
 sub modified_files {
   my $self = shift;
@@ -65,7 +66,7 @@ sub _manage {
   }
 
   $self->_spawn if !$self->{worker} && delete $self->{modified};
-  sleep 1;
+  sleep $self->sleep;
 }
 
 sub _spawn {
--- a/script/morbo
+++ b/script/morbo
@@ -9,12 +9,14 @@ getopt
   'l|listen=s' => \my @listen,
   'm|mode=s'   => \$ENV{MOJO_MODE},
   'v|verbose'  => \$ENV{MORBO_VERBOSE},
+  's|sleep=i'  => \my $sleep,
   'w|watch=s'  => \my @watch;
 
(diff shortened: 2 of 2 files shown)


kraih:
I'll have to give this a 👎 , since @leejo stated on IRC that this is not actually a problem for him and therefore entirely theoretic. To make an ugly implementation detail like this part of the public API there would have to a vert good reason.

jhthorsen:
I don't understand the use case. I hardly ever have the "CPU hog issue", but in some projects I just create a startup script with the appropriate "-w" switches to make it watch less files. The reasoning behind this is that I get confused if it takes too long for morbo to detect the changed files.

👎 from me as well.

leejo:
It's not a problem for me, but is for my colleagues, and FWIW setting the sleep period to 5s brings the cpu usage on my vm down from around 20% to less than 5%.

leejo:
I've rebased as per feedback from IRC and to remove the env variable to prevent exposing this part of the implementation.

leejo:
Bump! My colleagues say this has made a *significant* difference to their development environments. I've done some testing myself, changing sleep to 5s (through this change) cuts about 15% off my full test run time. It affects the host and guests like so on the proc table:

Current (sleep 1s):

**Guest**
```
%CPU %MEM    TIME+  COMMAND
  26  1.4   0:13.84 script/myapp
```

**Host**
```
PID    COMMAND      %CPU
10847  VBoxHeadless 56.2
```

Changing sleep to 5s:

**Guest**
```
%CPU %MEM    TIME+  COMMAND
  12  1.4   0:04.26 script/myapp
```

**Host** (not much difference here, but previously this would sit at the top of the running processes, now it only appears every 5s)
```
PID    COMMAND      %CPU
10847  VBoxHeadless 43.3
```

Repeating myself from IRC - we're not watching that many files (c. 600) and they're all in the application lib directory (`lib/`) and the mojo defaults (`myapp/templates`, `myapp/lib`). So we *could* glob some out but it wouldn't make much difference as we would then have to stop/start morbo in various cases.

kraih:
Thank you, i'm afraid this pull request as it stands did not pass the vote. While i do have a feeling that some version of this could pass, i don't know exactly what that would look like.

PR #1085, opened 2017-04-24
Allow undef text to indicate No children.

cheako:
### Summary
This is behaviour from CGI.pm
%= t meta => ( name => 'charset', content => 'UTF-8' ) => undef

Previously an end tag was generated.

### Motivation
To produce more accurately the HTML user intends.

### References
https://irclog.perlgeek.de/mojo/2017-04-24#i_14474630
https://pastebin.com/th12hkcS

### DIFF
--- a/lib/Mojo/DOM/HTML.pm
+++ b/lib/Mojo/DOM/HTML.pm
@@ -229,7 +229,7 @@ sub _render {
 
   # No children
   return $xml ? "$result />" : $EMPTY{$tag} ? "$result>" : "$result></$tag>"
-    unless $tree->[4];
+    if !$tree->[4] || $tree->[4]->[0] eq 'text' && !defined $tree->[4]->[1];
 
   # Children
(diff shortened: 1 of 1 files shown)


jhthorsen:
I don't get this. The meta tag is already supposed to have no end tag.

What are you trying to accomplish here?

Either way: without any tests and documentation, then this pull request won't get merged.

jberger:
Obviously I concur with @jhthorsen about needing tests and doc. Additionally I concur about not understanding the need. When I test I seem to get the expected result.

```bash
$ perl -Mojo -E 'a("/" => sub { $_->render(inline => q[<html><%= t meta => ( name => "charset", content => "UTF-8" ) %></html>]) })->start' get /
<html><meta content="UTF-8" name="charset"></html>
```

marcusramberg:
See no reason to leave this open after months without response from the original contributor.

PR #1086, opened 2017-04-24
Provide a more practical example regex

cheako:
### Summary
Currently user is restricted to letters e through t and that's a good example, but would almost never exist in the real world.  Thus provided is a regex that might be used to filter usernames.

### Motivation
I think it's less arbitrary and thus not confusing in an off the wall sort of way.

### References
No reply to my inquiry on IRC about this.
https://irclog.perlgeek.de/mojo/2017-04-23#i_14470938

### DIFF
--- a/lib/Mojolicious/Guides/Rendering.pod
+++ b/lib/Mojolicious/Guides/Rendering.pod
@@ -819,8 +819,10 @@ logic with methods like L<Mojolicious::Validator::Validation/"is_valid">.
     my $validation = $c->validation;
     return $c->render unless $validation->has_data;
 
-    # Validate parameters ("pass_again" depends on "pass")
-    $validation->required('user')->size(1, 20)->like(qr/^[e-t]+$/);
+    # Validate parameters ("user" size is within 1 - 20
+    # and content looks somewhat like a valid unix username)
+    $validation->required('user')->size(1, 20)
+      ->like(qr/^[a-z_][a-z0-9_-]*\$?$/);
     $validation->required('pass_again')->equal_to('pass')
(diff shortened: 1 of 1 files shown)


lib/Mojolicious/Guides/Rendering.pod
```diff
@@ -819,8 +819,8 @@ logic with methods like L<Mojolicious::Validator::Validation/"is_valid">.
     my $validation = $c->validation;
     return $c->render unless $validation->has_data;
 
-    # Validate parameters ("pass_again" depends on "pass")
-    $validation->required('user')->size(1, 20)->like(qr/^[e-t]+$/);
+    # Validate parameters ("user" size is withing 1 - 20 and content looks somewhat like a valid unix username)
```
jhthorsen: How is this related to the proposed change?

cheako: Changed the code and update the comment, silently ignoring that the comment was not altered after a copy/paste.

cheako: s/withing/within/

marcusramberg:
I don't like that this regex is longer and breaks the lines. If the problem is that e-t doesn't constitute a normal username restriction, I think we could change it to [a-z] or \w.

kraih:
Agree with @marcusramberg. https://github.com/kraih/mojo/commit/6b29ec87a5b6ec0730975b000fd341446b23c7c1

marcusramberg:
Closing this as it was obsoleted by sri 10 days ago.

PR #1087, opened 2017-04-24
Fix example meta charset to use name and content attribute names

cheako:
### Summary
The current example creates an attribute named charset when in HTML the attribute name should be content and charset belongs in another attribute named name.

### Motivation
To be correct and give proper expectation that there is no special handling of the meta tag in this instance.

### References


cheako:
Gasp, I thought to implement meta as a helper that supported the previous syntax only to find that meta might have an attribute named charset as seen in an example at w3schools.

Read:
https://www.w3schools.com/tags/tag_meta.asp

PR #1088, opened 2017-04-24
Add support for meta tag helper

cheako:
### Summary
Add handler for meta_tag that understands the intricacies of this HTML markup.

### Motivation
It's confusing to write ep using tag meta.

### References
Me being all confused by the tag examples, just didn't understand HTML.


cheako:
There are two small typos at
https://github.com/kraih/mojo/pull/1088/files#diff-ab8c5553c5f71a9d70038dac95d4d07dR75
A tab that snuck in at the tabstop after content-language and an extra 'a' after proposed.

cheako:
Hmm, double checking all of this.

PR #1089, opened 2017-04-24
Fixup handling of multiple radio buttons checked

cheako:
### Summary
Avoid setting checked attribute on multiple radio of the same name.

### Motivation
This is obviously incorrect handling on radio buttons.

### References
https://irclog.perlgeek.de/mojo/2017-04-24#i_14478333
https://irclog.perlgeek.de/mojo/2017-04-23#i_14473347

### DIFF
--- a/lib/Mojolicious/Plugin/TagHelpers.pm
+++ b/lib/Mojolicious/Plugin/TagHelpers.pm
@@ -80,7 +80,8 @@ sub _input {
     if ($type eq 'checkbox' || $type eq 'radio') {
       my $value = $attrs{value} // 'on';
       delete $attrs{checked};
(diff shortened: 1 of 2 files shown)


cheako:
Clearly this needs more discussion, I was under the mistaken impression that it was a simple fix.

Radiobuttons with one right and one wrong value, where the wrong value is last is a tricky beast to handle.  How do you know if the value is not used anywhere else?

marcusramberg:
@cheako I don't understand how this problem could occur in a normal form without manually hacking the query parameters. Can you explain how this is a real world problem?

marcusramberg:
Closing this due to lack of activity and failing tests.

PR #1090, opened 2017-04-25
Add support for meta tag helper

cheako:
### Summary
Add handler for meta_tag that understands the intricacies of this HTML markup.

### Motivation
It's confusing to write ep using tag meta.

### References
Me being all confused by the tag examples, just didn't understand HTML.


jhthorsen:
Even if the examples makes sense, I must say I think there are too many special cases to make this tag helper work properly.

marcusramberg:
:-1: from me as well for same reason.

marcusramberg:
Guess this also did not attract the required votes from mojo-core.

PR #1092, opened 2017-04-26
Test::Mojo# json_isa. Sometimes we need to check Perl type of the val…

akzhan:
…ue only.

### Summary

```perl
$t->json_isa("/projects", "ARRAY");
```

### Motivation
Sometimes we need to check Perl type of the JSON value only (need not to completely setup environment).

### References
None.

### DIFF
--- a/lib/Test/Mojo.pm
+++ b/lib/Test/Mojo.pm
@@ -169,6 +169,12 @@ sub json_is {
   return $self->_test('is_deeply', $self->tx->res->json($p), $data, $desc);
 }
 
+sub json_isa {
+  my ($self, $p, $reftype, $desc) = @_;
+  return $self->_test('isa_ok', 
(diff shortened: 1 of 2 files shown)


jberger:
I have a few comments, none of which are intended as an endorsement or rejection:

1) For consistency, if accepted, this would also need to be implemented for `json_message_isa`.
2) Are your arrays possibly empty? Is it not sufficient to check `->json_has('/projects/0')`?
3) When I get this involved in checking a data structure I typically bail out of the Test::Mojo system and start validating with something bigger like `is_deeply`, Test::Deep, or Test2::Bundle::Extended.

akzhan:
@jberger 

Firstly, thanks for mention about Test::Mojo::Role. It's just we need.

I suppose that we need no isa check in the core while it can be added as side role. Just because JSON and Perl types have no strict match.

So I close this Pull request.

PR #1093, opened 2017-05-11
better add_check documentation

2shortplanks:
### Summary
It's a minor improvement to the documentation

### Motivation
I spent too long scratching my head trying to work out that I should return false when validation succeeds rather than the other way around.  I had to read the source code, and that's never a good thing.

### References
Nothing comes to mind.

### DIFF
--- a/lib/Mojolicious/Validator.pm
+++ b/lib/Mojolicious/Validator.pm
@@ -133,11 +133,27 @@ implements the following new ones.
 
 Register a validation check.
 
-  $validator->add_check(foo => sub {
-    my ($validation, $name, $value, @args) = @_;
-    ...
-    return undef;
+  $validator->add_check( exclude_insensitve => sub {
+    my ($validation_obj, $name, $value, @args) = @_;
+      for (@args) {
+        # return false if validation passes
+        return undef if fc $_ eq fc $value;
+      }
+
+      # return true if validation fails
+      return 1;
   });
+
+The anonymous subroutine's return value indicates if the check fails (i.e.
+it should return true if the value fails to validate, or false if it is okay.)
+
+Once C<add_check> has been called, a new method will be usable on the
+validator class.
+
+  # any color name that isn't primary
+  $c->validator
+    ->required('color')
+    ->exclude_insensitve('red','green','blue');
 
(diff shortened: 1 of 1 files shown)


lib/Mojolicious/Validator.pm
```diff
@@ -133,12 +133,28 @@ implements the following new ones.
...
-    return undef;
+  $validator->add_check( exclude_insensitve => sub {
+    my ($validation_obj, $name, $value, @args) = @_;
+      for (@args) {
+        # return false if validation passes
+        return undef if fc $_ eq fc $value;
+      }
+
+      # return true if validation fails
+      return 1;
```
jhthorsen: The indentation is not right. There's some extra spaces on line 136 as well.

jhthorsen:
Maybe this can be fixed by adding a reference to https://metacpan.org/pod/distribution/Mojolicious/lib/Mojolicious/Guides/Rendering.pod#Adding-form-validation-checks instead?

lib/Mojolicious/Validator.pm
```diff
@@ -133,12 +133,28 @@ implements the following new ones.
...
+The anonymous subroutine's return value indicates if the check fails (i.e.
+it should return true if the value fails to validate, or false if it is okay.)
+
+Once C<add_check> has been called, a new method will be usable on the
+validator class.
+
+  # any color name that isn't primary
+  $c->validator
+    ->required('color')
+    ->exclude_insensitve('red','green','blue');
```
kraih: Shouldn't the docs of all `add_*` methods be somewhat consistent? This seems to go into a very different direction from the rest.

kraih:
Perhaps you should consider making changes to the real documentation instead. http://mojolicious.org/perldoc/Mojolicious/Guides/Rendering#Adding-form-validation-checks

2shortplanks:
I agree adding a link to the other documentation would be a good idea.

However, neither documentation actually specifies explicitly what arguments
the subroutine will receive.  We should improve that.

I'll work up a better documentation patch.

marcusramberg:
Excellent. Closing this and eagerly awaiting your new PR :+1:

PR #1096, opened 2017-05-18
Output formatters to allow list output of length 1

chy-causer:
### Summary
As discussed on IRC, this patch provides a mechanism to force a list in Mojolicious::Validator::Validation.pm output.

For me there are a few things that I'd welcome feedback:

 - The if ($output_formatter) seems a little weird here. Ideally I'd prefer there to be a default $output_formatter, but that proves tricky while maintaining complete backwards compatibility
https://github.com/kraih/mojo/compare/master...chy-causer:master#diff-ea8a39c17a3297cc4c5cc85000fae422R94
 - Not sure if ->required(with_gaps =>  'list')->is_valid is valid (if that makes sense.)
https://github.com/kraih/mojo/compare/master...chy-causer:master#diff-af7f52660125846348aa0b9dd5a72bf1R233

### Motivation

I want to have identically named input fields multiple times on a page and rely on the fact that the $validation->output will be an array, even if the number of non-blank elements given is one. If th

This patch provides the inverse functionality, where a scalar is returned even when more than one input is supplied. To me I'd expect this to be default behaviour, more in keeping with the normal usage of $c->param('foo') but I understand the reasoning for this and appreciate backwards compatibility is important so didn't change any existing tests.

### References
I started making sense around here: https://irclog.perlgeek.de/mojo/2017-04-28#i_14501813

### DIFF
--- a/lib/Mojolicious/Validator/Validation.pm
+++ b/lib/Mojolicious/Validator/Validation.pm
@@ -6,6 +6,13 @@ use Scalar::Util ();
 
 has [qw(csrf_token topic validator)];
 has [qw(input output)] => sub { {} };
+has output_formatters => sub {
+  {
+    first => sub { $_[0] },
+    last  => sub { $_[-1] },
+    list  => sub { [grep length, @_] },
+  };
+};
 
 sub AUTOLOAD {
   my $self = shift;
@@ -61,19 +68,37 @@ sub has_data { !!keys %{shift->input} }
 
 sub has_error { $_[1] ? exists $_[0]{error}{$_[1]} : !!keys %{$_[0]{error}} }
 
-sub is_valid { exists $_[0]->output->{$_[1] // $_[0]->topic} }
+sub is_valid {
+  my ($self, $topic) = ($_[0], $_[1] // $_[0]->topic);
+  return unless exists $self->output->{$topic};
+  my $val = $self->output->{$topic};
+  return ref $val ne 'ARRAY' || exists $val->[0];
+}
 
 sub optional {
   my ($self, $name, @filters) = @_;
 
   return $self->topic($name) unless defined(my $input = $self->input->{$name});
 
+  my ($output_formatters, $output_formatter) = $self->output_formatters;
+  if (my $output_format = (grep { $output_formatters->{$_} } @filters)[-1]) {
+    @filters = grep { not defined $output_formatters->{$_} } @filters;
+    $output_formatter = $output_formatters->{$output_format};
+  }
+
   my @input = ref $input eq 'ARRAY' ? @$input : ($input);
   for my $cb (map { $self->validator->filters->{$_} } @filters) {
     @input = map { $self->$cb($name, $_) } @input;
   }
-  $self->output->{$name} = ref $input eq 'ARRAY' ? \@input : $input[0]
-    if @input && !grep { !length } @input;
+
+  if ($output_formatter) {
+    my $output = $output_formatter->(@input);
+    $self->output->{$name} = $output if $output;
+  }
+  else {
+    $self->output->{$name} = ref $input eq 'ARRAY' ? \@input : $input[0]
+      if @input && !grep { !length } @input;
+  }
 
   return $self->topic($name);
 }
@@ -85,6 +110,7 @@ sub passed { [sort keys %{shift->output}] }
 sub required {
   my ($self, $name) = (shift, shift);
   return $self if $self->optional($name, @_)->is_valid;
+  delete $self->output->{$name};
   return $self->error($name => ['required']);
 }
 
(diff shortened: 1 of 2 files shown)


chy-causer:
The code submitted in the PR is when I prioritized the following conflicting criteria

* a way to scratch my own itch scratching first (obviously)
* full backwards compatibility
* code elegance.

I'll be the first to admit that the third criterion looks to be absent in the PR (although I did try, honest!) If you reorder this list (aka abandoning back-compat), then I believe some interesting things can happen.

Imagine the following. Assume each row in the table below is in isolation with the same boilerplate:

```perl
my $input = { single => '%', double => ['two', '2'], with_blanks => ['', 'not blank']};
my $val  = Mojolicious::Validator->new;
$val->input($input);
```

| Code  | Result |
| ------------- | ------------- |
| `$val->required('double')->like(qr/\d/)->is_valid`  | True  |
| `$val->required('double')->output->{double}`  | '2'  |
| `$val->required('double')->in('two')->is_valid`  | False  |
| `$val->required('double')->in('two')->{double}`  | undef  |
| `$val->list('double')->like(qr/\d/)->is_valid`  | False  |
| `$val->list('double')->like(qr/./)->is_valid`  | True  |
| `$val->list('double')->all_unique->is_valid`  | True  |
| `$val->list('double')->like(qr/./)->output->{double}`  | ['two', '2']  |
| `$val->list('nonexistent')->is_valid`  | True  |
| `$val->list('nonexistent')->output->{nonexistent}`  | []  |
| `$val->list('nonexistent')->has_number_elements(2)->is_valid`  | False  |
| `$val->list('nonexistent')->has_number_elements(2)->output->{nonexistent}`  | undef  |
| `$val->list('with_blanks')->output->{with_blanks}`  | ['not blank']  |
| `$val->required('single')->has_number_elements(1)->is_valid`  | True  |
| `$val->required('double')->has_number_elements(1)->is_valid`  | True  |
| `$val->list('with_blanks')->has_number_elements(1)->is_valid`  | True  |
| `$val->list('single')->output->{single}`  | ['%']  |

For me this would have the benefits of 

 * eliminating the need for defensive coding around $val->output, since the structure is predictable
 * allowing for more elaborate tests, such as the number of elements returned
 * (probably) giving a cleaner syntax than the current code. I personally find `$self->output->{$name} = ref $input eq 'ARRAY' ? \@input : $input[0]` a little jarring.

Of course the downside would be that this isn't fully backwards compatible. I would contend that a large number of sites would not notice the difference, but it certainly won't be 100%.

Given the modular nature of Mojolicious it is of course within my power to write my own module with this behaviour. I would unfortunately then lose out on all the TagHelpers goodness 😢

lib/Mojolicious/Validator/Validation.pm
```diff
@@ -61,19 +68,37 @@ sub has_data { !!keys %{shift->input} }
...
+  my $val = $self->output->{$topic};
+  return ref $val ne 'ARRAY' || exists $val->[0];
+}
 
 sub optional {
   my ($self, $name, @filters) = @_;
 
   return $self->topic($name) unless defined(my $input = $self->input->{$name});
 
+  my ($output_formatters, $output_formatter) = $self->output_formatters;
```
jhthorsen: This looks like a typo.

lib/Mojolicious/Validator/Validation.pm
```diff
@@ -61,19 +68,37 @@ sub has_data { !!keys %{shift->input} }
...
+
   my @input = ref $input eq 'ARRAY' ? @$input : ($input);
   for my $cb (map { $self->validator->filters->{$_} } @filters) {
     @input = map { $self->$cb($name, $_) } @input;
   }
-  $self->output->{$name} = ref $input eq 'ARRAY' ? \@input : $input[0]
-    if @input && !grep { !length } @input;
+
+  if ($output_formatter) {
+    my $output = $output_formatter->(@input);
```
jhthorsen: I think a reference should be passed to the output formatter, not a list.

lib/Mojolicious/Validator/Validation.pm
```diff
@@ -85,6 +110,7 @@ sub passed { [sort keys %{shift->output}] }
 sub required {
   my ($self, $name) = (shift, shift);
   return $self if $self->optional($name, @_)->is_valid;
+  delete $self->output->{$name};
```
jhthorsen: Why is this needed?

jberger:
I think I'm going to officially abstain on this vote. I think there should be a way to require/validate/select from multiple parameters, but as I'm not a user of the built-in validator I don't feel confident in voting on api design.

kraih:
I do think this feature should exist, but the implementation just isn't clean enough.

kraih:
I think we have consensus that we do want to see the feature added to Mojolicious, but that the implementation here is not right. So i'll close this pull request for now.

PR #1099, opened 2017-05-26
Add Minion version to Mojolicious::Command::version

bfaist:
### Summary
Add Minion version to Mojolicious::Command::version

### Motivation
It is helpful when using Minion with Mojolicious.  Otherwise it displays "n/a" like other optional modules.

### References
None.


jhthorsen:
👎 from me. I would rather have a `app minion version` command.

jberger:
I think I'm generally :-1: on this as well. If I was somehow convinced, I would think that all official spin-off projects would need to be included (ie, Mojo::Pg) or else some way to hook this page from external modules (which would probably end up being ridiculous).

PR #1100, opened 2017-06-02
Use MetaCPAN v1 API

aferreira:
### Summary
Currently "mojo version" silently fails to capture the latest Mojo version from CPAN, because

http://api.metacpan.org/v0/release/Mojolicious

says "MetaCPAN v0 API has been has been shut down! See the MetaCPAN v1 API should be used instead."

The solution is switch to http://fastapi.metacpan.org/v1/release/Mojolicious


kraih:
This patch appears to have lots of formatting errors.

kraih:
Thanks, fixed. https://github.com/kraih/mojo/commit/d35caf6121c742ad58c1b6ab39c94fa9fbe383da

PR #1107, opened 2017-07-19
Add location attributes; construct paths with Mojo::File

lindleyw:
### Summary
Replace literal pathnames with embedded `/` characters, use `Mojo::File` and its `child()` method to assemble component file paths. Along with the use of `decamelize()`, each call to `render_to_rel_file()` to create a file reflects how the framework will find that file during operation. Also: add a test for correct serving of the static file (which currently would fail if the application class is not under the `lib` directory).

### Motivation
to better explain components inter-relations and naming conventions, and to potentially use as a didactic resource in updated Guides

### DIFF
--- a/lib/Mojolicious/Command/generate/app.pm
+++ b/lib/Mojolicious/Command/generate/app.pm
@@ -2,7 +2,6 @@ package Mojolicious::Command::generate::app;
 use Mojo::Base 'Mojolicious::Command';
 
 use Mojo::Util qw(class_to_file class_to_path decamelize);
-use Mojo::Path;
 use Mojo::File;
 
 has description => 'Generate Mojolicious application directory structure';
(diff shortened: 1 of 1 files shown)


jberger:
IMO this adds a fair amount of complexity in order to allow generation of nonstandard application layouts. I understand that a power-user might want this, but the surely the opinionated case of doing the standard Perl distribution layout is an uncontroversial thing. Anyway, personally I don't see the need here (at least for a core module). As I said on IRC, a more powerful, configurable generator on CPAN would be a perfectly reasonable notion.

Also from a policy perspective, per http://mojolicious.org/perldoc/Mojolicious/Guides/Contributing#Rules

> Code has to be run through Perl::Tidy with the included .perltidyrc, and everything should look like it was written by a single person.

kraih:
Maybe also worth mentioning that the standard layout is referenced a lot in the documentation. http://mojolicious.org/perldoc/Mojolicious/Guides/Cookbook#Making-your-application-installable It would be pretty bad if newcomers got the impression that it was a good idea to change the defaults.

lindleyw:
Very good.  Withdrawn for further reflection.

PR #1108, opened 2017-07-20
slugify util function

Grinnz:
### Summary
A function to generate a URL-appropriate slug from an arbitrary string like a title.

### Motivation
This functionality is commonly used in web applications.

### References
Implementation and tests based on slugify from django utils: https://docs.djangoproject.com/en/1.11/ref/utils/#django.utils.text.slugify

### DIFF
--- a/lib/Mojo/Util.pm
+++ b/lib/Mojo/Util.pm
@@ -13,7 +13,8 @@ use List::Util 'min';
 use MIME::Base64 qw(decode_base64 encode_base64);
 use Pod::Usage 'pod2usage';
 use Symbol 'delete_package';
-use Time::HiRes ();
+use Time::HiRes        ();
+use Unicode::Normalize ();
 
 # Check for monotonic clock support
 use constant MONOTONIC =>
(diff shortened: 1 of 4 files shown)


lib/Mojo/Util.pm
```diff
@@ -248,6 +249,22 @@ sub secure_compare {
...
+  my ($value, $allow_unicode) = @_;
+  if ($allow_unicode) {
+
+    # force unicode semantics by upgrading string
+    utf8::upgrade($value = Unicode::Normalize::NFKC($value));
+    $value =~ s/[^\w\s-]+//g;
+  }
+  else {
+    $value = Unicode::Normalize::NFKD($value);
+    $value =~ s/[^a-zA-Z0-9_\p{PosixSpace}-]+//g;  # ascii word/space/dash chars
```
Grinnz: \p{PosixSpace} may be replaced with the equivalent ` \t\n\cK\f\r` but I'm not sure if that's clearer.

jberger:
This seems like it would be a very useful thing for users. I'm generally 👍

lib/Mojo/Util.pm
kraih: I don't mind `\p{PosixSpace}`.

lib/Mojo/Util.pm
```diff
@@ -709,6 +726,24 @@ Generate SHA1 checksum for bytes.
...
+replaced by a dash. By default, non-ASCII characters are normalized to ASCII
+word characters or removed, but if a true value is passed as the second
+parameter, all word characters will be allowed in the result according to
+unicode semantics.
+
+  # "this-is-my-resume"
+  slugify 'This is: my - résumé! ☃ ';
+
+  # "this-is-my-résumé"
+  slugify 'This is: my - résumé! ☃ ', 1;
```
kraih: I'll only give this a 👍 if we have an example that uses the text `Joel is a slug`. 😉

kraih:
👍 Joel is a slug.

PR #1112, opened 2017-07-23
:has selector for Mojo::DOM

pipcet:
### Summary
Add ".." and "..." selectors to Mojo/DOM/CSS.

This is more of an idea than a full PR at this point.

TL;DR:

```
mojo get http://www.tagesschau.de "a h4.headline .. a .dachzeile .. a"
```
harvests those A elements from tagesschau.de which contain h4 class=headline and class=dachzeile elements. The ".." backs up a level, like "cd .." does.

### Motivation
Some web sites fail to allow web harvesting because they mark embedded elements rather than containers: instead of

```
<p class="article">
<h4 class="headline">Blah</h4>
blah blah blah
```

they use something like

```
<div>
<h4 class="headline">Blah</h4>
blah blah blah
```

It would be nice to be able to harvest the containing <div> by writing
```
mojo get http://blah.com "h4.headline .. div"
```

So I hacked something up to allow that. It also allows "a ... b" to match any "b" ancestor of "a", but that's just an idea and a potential performance problem.

### Potential issues
- Drop compatibility with CSS.
- User-defined "CSS" selectors can take much longer to process.
- Not as powerful as XPath/XQuery

### DIFF
--- a/lib/Mojo/DOM/CSS.pm
+++ b/lib/Mojo/DOM/CSS.pm
@@ -19,7 +19,7 @@ my $ATTR_RE   = qr/
 
 sub matches {
   my $tree = shift->tree;
-  return $tree->[0] ne 'tag' ? undef : _match(_compile(shift), $tree, $tree);
+  return $tree->[0] ne 'tag' ? undef : _match(_compile(shift), $tree, undef);
 }
 
 sub select     { _select(0, shift->tree, _compile(@_)) }
@@ -28,9 +28,11 @@ sub select_one { _select(1, shift->tree, _compile(@_)) }
 sub _ancestor {
   my ($selectors, $current, $tree, $one, $pos) = @_;
 
+  return undef if defined($tree) && $current eq $tree;
   while ($current = $current->[3]) {
-    return undef if $current->[0] eq 'root' || $current eq $tree;
+    return undef if $current->[0] eq 'root';
     return 1 if _combinator($selectors, $current, $tree, $pos);
+    return undef if defined($tree) && $current eq $tree;
     last if $one;
   }
 
@@ -56,7 +58,7 @@ sub _combinator {
   # Selector
   return undef unless my $c = $selectors->[$pos];
   if (ref $c) {
-    return undef unless _selector($c, $current);
+    return undef unless _selector($c, $current, $tree);
     return 1 unless $c = $selectors->[++$pos];
   }
 
@@ -71,6 +73,22 @@ sub _combinator {
 
   # " " (ancestor)
   return _ancestor($selectors, $current, $tree, 0, ++$pos);
+}
+
+sub _relativize {
+  my $group = $_[0];
+
+  for my $selectors (@$group) {
+    shift @$selectors if ($selectors->[0] && !defined($selectors->[0]->[0]));
+    if ($selectors->[0] && ref $selectors->[0]) {
+      unshift @$selectors, ' ';
+    } elsif (defined($selectors->[0])) {
+    } else {
+      unshift @$selectors, ' ', [];
+    }
+  }
+
+  return $group;
 }
 
 sub _compile {
@@ -102,8 +120,9 @@ sub _compile {
     elsif ($css =~ /\G:([\w\-]+)(?:\(((?:\([^)]+\)|[^)])+)\))?/gcs) {
       my ($name, $args) = (lc $1, $2);
 
-      # ":matches" and ":not" (contains more selectors)
+      # ":matches", ":not", and ":has" (contains more selectors)
       $args = _compile($args) if $name eq 'matches' || $name eq 'not';
+      $args = _relativize(_compile($args)) if $name eq 'has';
 
       # ":nth-*" (with An+B notation)
       $args = _equation($args) if $name =~ /^nth-/;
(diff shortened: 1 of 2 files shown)


kraih:
I think what you want is the `:has` pseudo-class from the CSS4 selectors spec. https://drafts.csswg.org/selectors-4/#relational

kraih:
I know @jberger has been working towards supporting `:has` in his `:scope` support branch, but never got around to finishing it. https://github.com/kraih/mojo/compare/css_scope

pipcet:
Thanks for the link! You're absolutely right, `:has` would work as well.

Do you happen to know if anyone is working on XPath (or the full XQuery behemoth) support for Mojo?

Anyway, I've had a go at implementing just `:has`. It seems to work with a very simple test case, but I'll write some tests next... (The implementation's a hack, but I suspect only because I haven't fully understood CSS.pm)

kraih:
The biggest problem with `:has` is to get selectors starting with a combinator (`a:has(> img)`) to work.

kraih:
Guess it's worth mentioning that custom CSS selectors like `..` and `...` do not have a chance to get accepted. In case you were plannig on leaving them in.

pipcet:
My current approach is to match in the parent node of whichever node has the :has pseudo class; we construct a fake query of the form `xxx $args` for `a:has($args)`, then replace the "xxx" part by a special "equals" pattern that matches `a` precisely.

This fails to work for "a:has(b,c)"

pipcet:
Oh, no, ".." is quite unnecessary now. I'll take it out!

kraih:
I still believe that @jberger was right with his approach, and `dt:has(+ dt)` should be treated as `dt:has(:scope + dt)`.

pipcet:
If I understand `:scope` correctly (and that's precisely why I didn't do it that way), I agree. This is a stop-gap in the absence of @jberger's work.

kraih:
We don't care about the individual commits, just squash and force push.

pipcet:
I've pushed a version that passes existing tests, but fails to implement `:scope` siblings; that's this branch.

I've also made some additional changes to pass the two tests @jberger had disabled, i.e. implement scope siblings. That's at https://github.com/kraih/mojo/compare/master...pipcet:scope-siblings?expand=1 

(There's one unused function in this branch, `_scopize`).

Feel free to look at either or both! If you think it makes sense to prepare either version for inclusion, just let me know what needs changing.

jberger:
Since you've said that .. is unnecessary now, would you mind removing it so that the remaining bits are easier to review? It would really help

pipcet:
@jberger Hmm, I can't seem to find any traces of it in the latest versions. Am I missing something?

kraih:
The implementation looks much too complicated to me. Why is the whole `_relativize` stuff not just a single line in the beginning of the `_compile` function?
```perl
$css = ":scope $css" if $css =~ /^\s*[>+~]/;
```

kraih:
And if `_scopize` is dead code, why is it still there?

pipcet:
The ":scope $css" trick wouldn't deal with `:has(h1,h2,h3)`, which is the example in the CSS spec.

> And if `_scopize` is dead code, why is it still there?

Removed (that really belongs in the other version which separates `$tree` from `$scope` throughout CSS.pm, which makes `:scope` siblings and `:scope` within `:has` work).

> The implementation looks much too complicated to me

Thanks for having a look. I'll have another look to see whether I can simplify things, but I'd appreciate any advice or ideas.

kraih:
I see, the `:has(h1,h2,h3)` case does make things a whole lot more complicated. No wonder browsers don't support the feature yet.

pipcet:
Okay, I've made one minor simplification (make `_has` use `_select`, which it essentially duplicated), and added documentation. It's still a fairly complicated implementation, but I'm afraid the problem is complex.

Maybe we should close this and focus on implementing `:scope` and `:has` as completely as possible, as the other patch does? That treats `$scope` as an extra parameter for all functions that currently take a `$tree` argument, because they mean different things: `$tree` is the "highest" node we search, which must be the ancestor of `$scope` for `:scope` siblings to work.

kraih:
Implementing `:scope` and `:has` as completely as possible sounds good.

kraih:
Btw. If you're making `:scope` use a special op in the AST, why did you name it `equals` and not just `scope`?

kraih:
Also curious that the CSS4 selector spec doesn't mention `:has(> img, + dt)`, which is what makes `_relativize` so complicated.

pipcet:
Thanks for the comments!

> Btw. If you're making `:scope` use a special op in the AST, why did you name it `equals` and not just `scope`?

Sorry, fixed in what will become the new PR.

> Also curious that the CSS4 selector spec doesn't mention `:has(> img, + dt)`, which is what makes `_relativize` so complicated.

I'm not sure we could simplify `_relativize` very much if that weren't allowed; we'd just be able to drop the `for` loop, maybe.

As far as I can see, what we could do differently is to drop `_relativize`/`_scopize` and put that logic in `_compile`. Is that worth exploring?

kraih:
> As far as I can see, what we could do differently is to drop _relativize/_scopize and put that logic in _compile. Is that worth exploring?
Sure, as long as the code in `_compile` stays as clean as it is now.

PR #1113, opened 2017-07-25
Add 'special' attribute to Mojolicious::Static

### LINKED ISSUE #1094: MyApp in production serves a jQuery and other debug-only resources
```
$ mojo version
CORE
  Perl        (v5.18.2, darwin)    ## Mac OS X - 10.10.5 (14F2109)
  Mojolicious (7.30, Doughnut)
 - - 8< - -
```

### Steps to reproduce the behaviour
```
$ ./script/newly_generated_app get /mojo/jquery/jquery.js
$ MOJO_MODE=production hypnotoad  newly_generated_app/script/newly_generated_app
[Tue May 16 22:41:50 2017] [info] Listening at "http://*:8080"
Server available at http://127.0.0.1:8080
$  curl -q  http://localhost:8080/mojo/jquery/jquery.js  2>/dev/null |head -n1
/*! jQuery v3.1.1 | (c) jQuery Foundation | jquery.org/license */
```

### Expected behaviour
I would expect jQuery, and other `debug.html.ep`-only resources to 404 when the application is in production mode.

### Actual behaviour
You get jQuery 3.1.1:
`/*! jQuery v3.1.1 | (c) jQuery Foundation | jquery.org/license */`

Looking at the template one could imagine replacing the jQuery usage with plain-old-javascript as it's only a couple of hide/shows, but it might make more sense to arrange that content of that directory be served by a plugin/helper that's only loaded when the app is in development mode.

Thoughts?

lindleyw:
…to permit examining, overriding, or disabling any or all bundled files. May resolve #1094

### Summary
Enumerate bundled, static files so application can examine or modify their resolution at runtime. For example, a Mojolicious::Lite app could:

    delete app->static->special->{'favicon.ico'};

and then proceed to serve `favicon.ico` from a route.  

    app->static->special->{'mojo/jquery/jquery.js'} =
      app->home->child('resources', 'jquery.js');

would override the built-in jQuery for error pages with one in the app's `resources` directory.

Note: It might also be desirable to permit values which are Mojo::Asset objects, rather than filenames.

### References
[issue #1094 ](https://github.com/kraih/mojo/issues/1094)

### DIFF
--- a/lib/Mojolicious/Static.pm
+++ b/lib/Mojolicious/Static.pm
@@ -11,10 +11,19 @@ use Mojo::Util 'md5_sum';
 
 has classes => sub { ['main'] };
 has paths   => sub { [] };
+has special => sub {
 
-# Bundled files
-my $PUBLIC = Mojo::Home->new(Mojo::Home->new->mojo_lib_dir)
-  ->child('Mojolicious', 'resources', 'public');
+  # Bundled files are handled specially
+  my $public = Mojo::Home->new(Mojo::Home->new->mojo_lib_dir)
+    ->child('Mojolicious', 'resources', 'public');
+  return {
+    @{
+      $public->list_tree->map(
+        sub { join('/', @{$_->to_rel($public)}), $_->realpath; }
+      )
+    }
+  };
+};
 
 sub dispatch {
   my ($self, $c) = @_;
@@ -48,8 +57,9 @@ sub file {
   # Search DATA
   if (my $asset = $self->_get_data_file($rel)) { return $asset }
 
-  # Search bundled files
-  return $self->_get_file(path($PUBLIC, split('/', $rel))->to_string);
+  # Search special files
+  my $special = $self->special;
+  return exists $special->{$rel} ? $self->_get_file($special->{$rel}) : undef;
 }
 
 sub is_fresh {
@@ -158,6 +168,7 @@ Mojolicious::Static - Serve static files
   my $static = Mojolicious::Static->new;
   push @{$static->classes}, 'MyApp::Controller::Foo';
   push @{$static->paths}, '/home/sri/public';
+  delete $static->special->{'mojo/jquery/jquery.js'};
 
 =head1 DESCRIPTION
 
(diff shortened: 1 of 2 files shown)


kraih:
I like this solution. But i'd like for the attribute to have a more generic name, since anyone can add more "special" files. And of course the implementation needs to be cleaner, and tests/documentation added.

lib/Mojolicious/Static.pm
```diff
@@ -201,6 +213,23 @@ Directories to serve static files from, first one has the highest precedence.
...
+
+  delete app->static->special->{'favicon.ico'};
+
+and then proceed to serve favicon.ico from a route. Or:
+
+  app->static->special->{'mojo/jquery/jquery.js'} =
+    app->home->child('resources', 'jquery.js');
+
+would override the built-in jQuery for error pages with one in the app's
+C<resources> directory.
```
kraih: This is not the correct documentation format for attributes. Please replicate the style of the surrounding documentation.

lib/Mojolicious/Static.pm
```diff
@@ -48,8 +59,9 @@ sub file {
   # Search DATA
   if (my $asset = $self->_get_data_file($rel)) { return $asset }
 
-  # Search bundled files
-  return $self->_get_file(path($PUBLIC, split('/', $rel))->to_string);
+  # Search special files
+  return undef unless exists $self->special->{$rel};
+  return $self->_get_file($self->special->{$rel});
```
kraih: The attribute accessor is called twice.
```perl
my $special = $self->special;
return exists $special->{$rel} ? $self->_get_file($special->{$rel}) : undef;
```

lib/Mojolicious/Static.pm
```diff
@@ -11,10 +11,21 @@ use Mojo::Util 'md5_sum';
...
+  return {
+    @{
+      $public->list_tree->map(
+        sub {
+          $_->to_rel($public)->to_string, $_->realpath;
+        }
+      )
+    }
+  };
+};
```
kraih: I doubt the `$_->to_rel($public)->to_string` works on Windows.

kraih: Confirmed by AppVeyor test results.

PR #1115, opened 2017-07-28
implement :has,:scope CSSv4 selectors in Mojo::DOM::CSS

pipcet:
### Summary
Implement the `:has` and `:scope` selectors from the CSSv4 draft (https://drafts.csswg.org/selectors).

### Motivation
Allow harvesting HTML from sites based on a `:has` specification, to capture an ancestor of an identifiable element.

### References
 - #1112
 - @jberger's work at https://github.com/kraih/mojo/compare/css_scope has been included in this patch.

This patch could do with some clever simplification. Ideas and advice are very welcome. There are some tests, but there could be more, and I'd be particularly interested in tests that fail.

Major changes:
 - introduce `$scope` argument for most Mojo::DOM::CSS functions
 - make `$tree` argument mean "the highest node we'll look at" consistently, rather than "stop at this node, but don't stop if you've already passed it" as previously

I'm not at all married to the implementation, but I think the `:has` selector is something users would find useful.

### DIFF
--- a/lib/Mojo/DOM/CSS.pm
+++ b/lib/Mojo/DOM/CSS.pm
@@ -19,18 +19,29 @@ my $ATTR_RE   = qr/
 
 sub matches {
   my $tree = shift->tree;
-  return $tree->[0] ne 'tag' ? undef : _match(_compile(shift), $tree, $tree);
+  return $tree->[0] ne 'tag' ? undef : _match(_compile(shift), $tree, undef, $tree);
 }
 
-sub select     { _select(0, shift->tree, _compile(@_)) }
-sub select_one { _select(1, shift->tree, _compile(@_)) }
+sub _ptree {
+  my $tree = $_[0];
+
+  return $tree->[3] if ($tree->[0] eq 'tag');
+  return $tree;
+}
+
+sub select     { my $tree = shift->tree; _select(0, _ptree($tree), $tree, _scopize(_compile(@_))) }
+sub select_one { my $tree = shift->tree; _select(1, _ptree($tree), $tree, _scopize(_compile(@_))) }
 
 sub _ancestor {
-  my ($selectors, $current, $tree, $one, $pos) = @_;
+  my ($selectors, $current, $tree, $scope, $one, $pos) = @_;
 
+  if (defined($tree) && $current eq $tree || $current->[0] eq 'root') {
+    return undef;
+  }
   while ($current = $current->[3]) {
-    return undef if $current->[0] eq 'root' || $current eq $tree;
-    return 1 if _combinator($selectors, $current, $tree, $pos);
+    return 1 if _combinator($selectors, $current, $tree, $scope, $pos);
+    return undef if $current->[0] eq 'root';
+    return undef if defined($tree) && $current eq $tree;
     last if $one;
   }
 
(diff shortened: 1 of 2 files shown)


kraih:
I've played a bit with the code and found quite a few lines that don't break any tests when they are removed. The two functions `_scopize` and `_has_scope` really worry me, they seem very hackish and focused on only a few very specific selectors (not generic!), so selectors like `div:scope` just don't work at all.

kraih:
And then there's a fundamental design problem we need to discuss.
```
<foo>
  <foo>
    <bar></bar>
  </foo>
  <bar></bar>
</foo>
```

Because of a jQuery-ish built in scoping rule `$dom->at('foo')->find('foo > bar')` currently only matches the first `<bar>` tag, but not the second. While `$dom->at('foo')->find(':scope foo > bar')` would allow us to have distinct selectors for both. But we would have to make a small breaking change.

lib/Mojo/DOM/CSS.pm
```diff
@@ -149,32 +196,40 @@ sub _equation {
...
-  return !_match($args, $current, $current) if $class eq 'not';
+  return !_match($args, $current, $current, $scope) if $class eq 'not';
 
   # ":matches"
-  return !!_match($args, $current, $current) if $class eq 'matches';
+  return !!_match($args, $current, $current, $scope) if $class eq 'matches';
 
-  # ":empty"
-  return !grep { !_empty($_) } @$current[4 .. $#$current] if $class eq 'empty';
+  return undef if $current->[0] eq 'root';
```
kraih: This is one of the lines i removed without any tests breaking.

lib/Mojo/DOM/CSS.pm
```diff
@@ -149,32 +196,40 @@ sub _equation {
...
 
+  # ":checked"
+  return exists $current->[2]{checked} || exists $current->[2]{selected}
+    if $class eq 'checked';
+
+  # ":empty"
+  return !grep { !_empty($_) } @$current[4 .. $#$current] if $class eq 'empty';
+
+  # ":has"
+  return !!_has(1, $current, $args) if $class eq 'has';
```
kraih: Using a function here does not appear to serve a purpose.
```perl
return !!_select(1, _ptree($current), $current, $args) if $class eq 'has';
```

lib/Mojo/DOM/CSS.pm
```diff
@@ -51,26 +64,59 @@ sub _attr {
...
-  return _ancestor($selectors, $current, $tree, 0, ++$pos);
+  return _ancestor($selectors, $current, $tree, $scope, 0, ++$pos);
+}
+
+sub _has_scope {
+  my $selectors = $_[0];
+
+  for my $selector (@$selectors) {
+    next unless ref $selector && defined($selector->[0][0]) &&
+        $selector->[0][0] eq 'pc';
```
kraih: Removing `&& $selector->[0][0] eq 'pc'` does not make a difference here.

kraih:
Starting to get a feeling that the entire handling of scopes in this patch is wrong (and much too complicated) because of the implicit scoping rules applied by the `at` and `find` methods.

kraih:
I've thought a little bit about a more correct implementation, and [here's my attempt](https://github.com/kraih/mojo/compare/scope_and_has). Think it's only missing support for the [descendant combinator](https://drafts.csswg.org/selectors/#relative-selector) which turned out to be a little more complicated. But should be possible to add to `_compile` as well.

kraih:
I'll close this pull request, since the implementation is not correct.

pipcet:
Thanks for taking a look. I confess I haven't had time to investigate yet, but now there's a much more promising implementation I agree we should stop spending time on this one.

PR #1117, opened 2017-08-02
Fix ordering problem in unquote.

AndyA:
Hi there,

unquote needs to handle all backslash escaped characters before handling backslash itself otherwise \\\" is handled incorrectly (it returns " instead of \").

### DIFF
--- a/lib/Mojo/Util.pm
+++ b/lib/Mojo/Util.pm
@@ -309,8 +309,8 @@ sub unindent {
 sub unquote {
   my $str = shift;
   return $str unless $str =~ s/^"(.*)"$/$1/g;
-  $str =~ s/\\\\/\\/g;
   $str =~ s/\\"/"/g;
+  $str =~ s/\\\\/\\/g;
   return $str;
 }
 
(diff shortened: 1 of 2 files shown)


Grinnz:
This test is incorrect. The source code `'"\\\""'` creates the string `"\\""` which is not a correctly "quoted" string as it has an unescaped double-quote, so unquote cannot be expected to handle it correctly. The correctly escaped string `"\\\""` is correctly unquoted to `\"`.

    $ perl -E'say q{"\\\\\""}'
    "\\\""
    $ perl -MMojo::Util=unquote -E'say unquote(q{"\\\\\""})'
    \"

Grinnz:
If you have an example where a correctly quoted string is not correctly unquoted then it would be a good test to add.

AndyA:
Ah! Sorry for that. My mistake.

kraih:
So, is there actually a problem here or not?

AndyA:
No - there's no problem. Sorry for the noise.

PR #1120, opened 2017-08-09
Mojo::Base::with_roles patch

dotandimet:
### Summary
Add with_roles method to Mojo::Base, per the specs described here:
https://irclog.perlgeek.de/mojo/2017-08-08#i_14984146

### Motivation
Allow multiple developers to extend key Mojolicious classes (Test::Mojo, Mojo::UserAgent, etc)
using Role::Tiny (optional dependency)

### References
https://irclog.perlgeek.de/mojo/2017-08-08#i_14984146

### DIFF
--- a/lib/Mojo/Base.pm
+++ b/lib/Mojo/Base.pm
@@ -15,6 +15,10 @@ use IO::Handle ();
 my $NAME
   = eval { require Sub::Util; Sub::Util->can('set_subname') } || sub { $_[1] };
 
+# Role support requires Role::Tiny 2.000001+
+use constant ROLES =>
+  !!(eval { require Role::Tiny; Role::Tiny->VERSION('2.000001'); 1 });
+
 # Protect subclasses using AUTOLOAD
 sub DESTROY { }
 
@@ -61,6 +65,8 @@ sub attr {
   }
 }
 
+sub can_roles {ROLES}
+
 sub import {
   my $class = shift;
   return unless my $flag = shift;
@@ -99,6 +105,11 @@ sub tap {
   my ($self, $cb) = (shift, shift);
   $_->$cb(@_) for $self;
   return $self;
+}
+
+sub with_roles {
+  return Role::Tiny->create_class_with_roles(@_) if (ROLES);
+  Carp::croak "Role::Tiny 2.000001+ is required for with_roles method";
 }
 
 1;
@@ -210,6 +221,13 @@ executed at accessor read time if there's no set value, and gets passed the
 current instance of the object as first argument. Accessors can be chained, that
 means they return their invocant when they are called with an argument.
 
+=head2 can_roles
+
+  my $bool = Mojo::Base->can_roles();
+
+True if L<Role::Tiny> 2.000001+ is installed, indicates that roles are supported in L<Mojo::Base>
+derived classes.
+
 =head2 new
 
   my $object = SubClass->new;
@@ -237,6 +255,15 @@ spliced or tapped into) a chained set of object method calls.
 
   # Inject side effects into a method chain
   $object->foo('A')->tap(sub { say $_->foo })->foo('B');
+
+=head2 with_roles
+
+  my $new_class = Class->with_roles('Foo::Role1', 'Bar::Role2');
+  my $object = $new_class->new();
+
+Create and return a new class that extends the given class with the
+list of roles composed in order, using L<Role::Tiny>'s method
+C<create_class_with_roles>.
 
 =head1 SEE ALSO
 
--- a/lib/Mojolicious/Command/version.pm
+++ b/lib/Mojolicious/Command/version.pm
@@ -16,6 +16,7 @@ sub run {
     = Mojo::IOLoop::Client->can_socks ? $IO::Socket::Socks::VERSION : 'n/a';
   my $tls = Mojo::IOLoop::TLS->can_tls    ? $IO::Socket::SSL::VERSION  : 'n/a';
   my $nnr = Mojo::IOLoop::Client->can_nnr ? $Net::DNS::Native::VERSION : 'n/a';
+  my $roles = Mojo::Base->can_roles ? $Role::Tiny::VERSION : 'n/a';
 
   print <<EOF;
 CORE
@@ -27,6 +28,7 @@ OPTIONAL
   IO::Socket::Socks 0.64+ ($socks)
   IO::Socket::SSL 1.94+   ($tls)
   Net::DNS::Native 0.15+  ($nnr)
+  Role::Tiny 2.000001+    ($roles)
 
 EOF
 
--- a/t/mojo/roles.t
+++ b/t/mojo/roles.t
@@ -1,0 +1,59 @@
+use Mojo::Base -strict;
+
+use Test::More;
+
+BEGIN {
+  plan skip_all => 'Role::Tiny 2.000001+ required for this test!'
+    unless Mojo::Base->can_roles;
+}
+
+package Mojo::RoleTest::LOUD;
+use Role::Tiny;
+
+sub yell {'HEY!'}
+
+requires 'name';
+
+sub hello {
+  my ($self) = @_;
+  return $self->yell . ' ' . uc($self->name) . '!!!';
+}
+
+package Mojo::RoleTest::quiet;
+use Role::Tiny;
+
+requires 'name';
+
+sub whisper {
+  my ($self) = @_;
+  return 'psst, ' . lc($self->name);
+}
+
+package Mojo::RoleTest::Base;
+use Mojo::Base -base;
+
+has name => 'bob';
+
+sub hello {
+  my ($self) = shift;
+  return 'hello ' . $self->name;
+}
+
+package main;
+
+my $obj = Mojo::RoleTest::Base->new(name => 'Ted');
+is($obj->name,  'Ted',       'attr works');
+is($obj->hello, 'hello Ted', 'class method');
+
+my $obj2 = Mojo::RoleTest::Base->with_roles('Mojo::RoleTest::LOUD')->new;
+is($obj2->hello, 'HEY! BOB!!!', 'method from role overrides base method');
+is($obj2->yell,  'HEY!',        'new method from role');
+
+my $obj3 = Mojo::RoleTest::Base->with_roles('Mojo::RoleTest::quiet',
+  'Mojo::RoleTest::LOUD')->new(name => 'Joel');
+is($obj3->name,    'Joel',         'attr from base class');
+is($obj3->whisper, 'psst, joel',   'method from role1');
+is($obj3->hello,   'HEY! JOEL!!!', 'method override from role2');
+
+done_testing();
+


lib/Mojo/Base.pm
```diff
@@ -61,6 +61,10 @@ sub attr {
   }
 }
 
+sub can_roles {
+  return !!(eval 'require Role::Tiny' && $Role::Tiny::VERSION >= 2.000005);
```
jberger: This should be a constant and the `can_roles` method is just for exposure outside of the class. C.f. https://github.com/kraih/mojo/blob/be50b20ae6c80df78a2594fdd198c19364f649ad/lib/Mojo/IOLoop/Client.pm#L13-L15

lib/Mojo/Base.pm
```diff
@@ -101,6 +105,16 @@ sub tap {
   return $self;
 }
 
+sub with_roles {
+  if ($_[0]->can_roles()) {
```
jberger: Within the class, it should use the constant (see above), since this will be constant folded at compile time and not introduce overhead on each with_roles call

lib/Mojo/Base.pm
```diff
@@ -61,6 +61,10 @@ sub attr {
   }
 }
 
+sub can_roles {
+  return !!(eval 'require Role::Tiny' && $Role::Tiny::VERSION >= 2.000005);
+}
```
kraih: This is not consistent with our other optional dependencies.

lib/Mojo/Base.pm
```diff
@@ -101,6 +105,16 @@ sub tap {
   return $self;
 }
 
+sub with_roles {
+  if ($_[0]->can_roles()) {
+    Carp::croak "class and roles must be package names" if !!grep { ref $_ } @_;
```
kraih: Is there a good reason that argument validation is required here?

lib/Mojo/Base.pm
```diff
@@ -101,6 +105,16 @@ sub tap {
...
+    return Role::Tiny->create_class_with_roles(@_);
+  }
+  else {
+    Carp::croak "Role::Tiny 2.000005 is required for with_roles method";
```
kraih: Do we really require `2.000005`, or would a nicer looking `2.0` work too?

lib/Mojo/Base.pm
```diff
@@ -210,6 +224,13 @@ executed at accessor read time if there's no set value, and gets passed the
 current instance of the object as first argument. Accessors can be chained, that
 means they return their invocant when they are called with an argument.
 
+=head2 can_roles
+
+  my $roles_ok = Mojo::Base->can_roles()
+
+True if L<Role::Tiny> 2.000005+ is installed and roles are supported in Mojo::Base
```
kraih: `Mojo::Base` is not a link.

lib/Mojo/Base.pm
```diff
@@ -210,6 +224,13 @@ executed at accessor read time if there's no set value, and gets passed the
 current instance of the object as first argument. Accessors can be chained, that
 means they return their invocant when they are called with an argument.
 
+=head2 can_roles
+
+  my $roles_ok = Mojo::Base->can_roles()
```
kraih: We write `my $bool = ...` for methods that return booleans.

lib/Mojo/Base.pm
```diff
@@ -238,6 +259,15 @@ spliced or tapped into) a chained set of object method calls.
   # Inject side effects into a method chain
   $object->foo('A')->tap(sub { say $_->foo })->foo('B');
 
+=head2 with_roles
+
+  my $class = Some::BaseClass->with_roles(qw(MyApp::Role1 MyApp::Role2));
+  my $object = $class->new();
```
kraih: This is inconsistent with other method examples.

lib/Mojo/Base.pm
```diff
@@ -238,6 +259,15 @@ spliced or tapped into) a chained set of object method calls.
...
+
+Create a new class based on base, with the roles composed into it in order,
+using L<Role::Tiny>'s C<create_class_with_roles> method. Roles must be
+fully-qualified package names.
```
kraih: The description needs some work, `...a new class based on base...`?

lib/Mojolicious/Command/version.pm
```diff
@@ -27,6 +28,7 @@ OPTIONAL
   IO::Socket::Socks 0.64+ ($socks)
   IO::Socket::SSL 1.94+   ($tls)
   Net::DNS::Native 0.15+  ($nnr)
+  Role::Tiny 2.000005+ ($roles)
```
kraih: Indentation.

t/mojo/roles.t
```diff
@@ -0,0 +1,34 @@
+use Mojo::Base -strict;
+use Test::More;
+
+plan skip_all => "Role::Tiny required for roles" unless Mojo::Base->can_roles;
```
kraih: This is inconsistent with other tests for optional dependencies.

t/mojo/roles.t
```diff
@@ -0,0 +1,34 @@
+use Mojo::Base -strict;
```
kraih: Empty line missing.

t/mojo/roles.t
```diff
@@ -0,0 +1,34 @@
...
+use Test::More;
+
+plan skip_all => "Role::Tiny required for roles" unless Mojo::Base->can_roles;
+
+# First test derived from sri's comment here:
+# https://github.com/kraih/mojo/pull/1118#issuecomment-320930345
+# TODO: make this a subtest
```
kraih: We don't use these kinda comments.

t/mojo/roles.t
```diff
@@ -0,0 +1,34 @@
...
+
+package Test::Mojo::Role::Location;
+
+use Role::Tiny;
+use Test::More; # for is()
```
kraih: Used twice.

t/mojo/roles.t
```diff
@@ -0,0 +1,34 @@
...
+
+use Test::Mojo;
+
+plan tests => 3;
+
+my $t = Test::Mojo->with_roles('Test::Mojo::Role::Location')->new();
+$t->get_ok('http://mojolicio.us')
+  ->status_is(301)
+  ->location_is('http://mojolicious.org/')
+  ->or(sub { diag 'doh!' });
```
kraih: I don't think this is a good test case, keep it simple.

lib/Mojo/Base.pm
kraih: Also missing semicolon.

t/mojo/roles.t
```diff
@@ -0,0 +1,34 @@
...
+  return $t->success(is($t->tx->res->headers->location, $value, $desc));
+}
+
+1;
+
+package main;
+
+use Test::Mojo;
+
+plan tests => 3;
```
kraih: We don't use plans.

t/mojo/roles.t
```diff
@@ -0,0 +1,34 @@
...
+
+my $t = Test::Mojo->with_roles('Test::Mojo::Role::Location')->new();
+$t->get_ok('http://mojolicio.us')
```
kraih: Tests shouldn't require an internet connection.

lib/Mojo/Base.pm
Grinnz: I would recommend at least 2.000001 for an important fix to created class names. https://rt.cpan.org/Public/Bug/Display.html?id=103310

lib/Mojo/Base.pm
Grinnz: create_class_with_roles will error if anything other than a valid role name is passed, so probably not needed

Grinnz: It may be worthwhile checking that the invocant is not an object though

lib/Mojo/Base.pm
dotandimet: I'll change it to 2.000001+

lib/Mojo/Base.pm
dotandimet: I'll change it to only check the invocant.

kraih: That's not something we normally do.

dotandimet: I'll remove it and see how cryptic the error is in the tests...

dotandimet: OK, my error is nice, but if you try to invoke with_roles on an object, you get a terrifying exception, so I'm fine with that (something like "Can't locate Mojo/RoleTest/Base__WITH__Mojo/RoleTest/LOUD=HASH(0x17e56b0).pm in @INC", or whatever)

kraih:
Don't forget to squash your commits.

lib/Mojo/Base.pm
```diff
@@ -15,6 +15,10 @@ use IO::Handle ();
 my $NAME
   = eval { require Sub::Util; Sub::Util->can('set_subname') } || sub { $_[1] };
 
+# Role support requires Role::Tiny 2.000001+
+use constant ROLES =>
+  !!(eval 'require Role::Tiny' && $Role::Tiny::VERSION >= 2.000001);
```
aferreira: why not ` eval 'use Role::Tiny 2.000001 (); 1'; `
like at https://github.com/kraih/mojo/blob/master/lib/Mojo/IOLoop/TLS.pm#L11 ?

lib/Mojolicious/Command/version.pm
aferreira: 2.000001+

lib/Mojo/Base.pm
Grinnz: Actually the best method would be `eval { require Role::Tiny; Role::Tiny->VERSION('2.000001'); 1 }`

String eval is unnecessary for this, and the ->VERSION method (same as used by use) is the correct way to check a module's version.

t/mojo/roles.t
```diff
@@ -0,0 +1,59 @@
+use Mojo::Base -strict;
+
+BEGIN {
+  use Test::More;
+
+  plan skip_all => 'Role::Tiny 2.000001+ required for this test!'
+    unless Mojo::Base->can_roles;
+}
```
kraih: This is inconsistent with the other tests again...

lib/Mojo/Base.pm
kraih: That does look much better.

t/mojo/roles.t
```diff
@@ -0,0 +1,60 @@
+use Mojo::Base -strict;
+
+BEGIN {
+# bail in a BEGIN block, so we can use Role::Tiny in inline classes below
+  use Test::More;
```
kraih: Move the `use Test::More` outside the block and remove the comment.

lib/Mojolicious/Command/version.pm
```diff
@@ -27,6 +28,7 @@ OPTIONAL
   IO::Socket::Socks 0.64+ ($socks)
   IO::Socket::SSL 1.94+   ($tls)
   Net::DNS::Native 0.15+  ($nnr)
+  Role::Tiny 2.000001+ ($roles)
```
kraih: Indentation is still wrong.

PR #1125, opened 2017-08-27
Add regression test for controller loader bug

augensalat:
### Summary
`Mojo::Loader` fails loading controller classes under certain circumstances.

### Motivation
This is actually a bug report that provides a failing regression test. I have no idea how to fix without breaking at the other end.

### Setup
The long story:
1. A Mojolicious app (example code below) with two controller namespaces (`Foo` and `Foo::Bar`) where the seconds lies down in the hierarchy of the first.
2. `Foo` does not have a controller class, but only a template.
3. Start app and run queries in the following order
3.1 First request to `foo-bar#index` works
3.2 First request to `foo#index` works
3.3 Second request to `foo-bar#index` crashes
```Can't locate object method "new" via package "TestApp::Controller::Foo::Bar" (perhaps you forgot to load "TestApp::Controller::Foo::Bar"?) ...```

### Perception
This error was introduced with commit bfc5449 that should fix issue #944.
It is caused by to enthusiastic unloading of `Mojo::Util::_teardown` and `Symbol::delete_package` respectively.

### The code
App
```perl
package TestApp;
use Mojo::Base 'Mojolicious';

sub startup {
  my $r = shift->routes;
  $r->get('/')->to(controller => 'Foo', action => 'index');
  $r->get('/foo')->to(controller => 'Foo::Bar', action => 'index');
}

1;
```
Foo::Bar controller
```perl
package TestApp::Controller::Foo::Bar;

use Mojo::Base 'Mojolicious::Controller';

sub index { shift->render }

1;
```
Foo template
```perl
% my $c = shift;
Hello Mojo from the template <%= $c->url_for %>!
```
Foo-Bar template
```perl
: my $c = shift;
Hello Mojo from the other template <%= $c->url_for %>!
```

### DIFF
--- a/t/mojolicious/app-loader.t
+++ b/t/mojolicious/app-loader.t
@@ -1,0 +1,27 @@
+use Mojo::Base -strict;
+
+use Test::More;
+
+use FindBin;
+use lib "$FindBin::Bin/lib";
+
+use Mojolicious;
+use Test::Mojo;
+
+my $t = Test::Mojo->new('MojoliciousLoaderTest');
+
+
+# Application is already available
+$t->get_ok('/bar')->status_is(200)
+  ->header_is(Server => 'Mojolicious (Perl)')
+  ->content_like(qr/Hello Mojo from the other template \/bar!/);
+
+$t->get_ok('/foo')->status_is(200)
+  ->header_is(Server => 'Mojolicious (Perl)')
+  ->content_like(qr/Hello Mojo from the template \/foo!/);
+
+$t->get_ok('/bar')->status_is(200)
+  ->header_is(Server => 'Mojolicious (Perl)')
+  ->content_like(qr/Hello Mojo from the other template \/bar!/);
+
+done_testing();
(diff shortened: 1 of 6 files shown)


jhthorsen:
This is my suggestion:

```
diff --git a/lib/Mojo/Util.pm b/lib/Mojo/Util.pm
index b2ddcad1a..6bafca764 100644
--- a/lib/Mojo/Util.pm
+++ b/lib/Mojo/Util.pm
@@ -466,7 +466,18 @@ sub _teardown {
   # @ISA has to be cleared first because of circular references
   no strict 'refs';
   @{"${class}::ISA"} = ();
+  my ($ns, %keep) = (\%{"$class\::"});
+
+  use strict 'refs';
+  for my $k (keys %$ns) {
+    next unless my $stash = *{$ns->{$k}}{HASH};
+    my $sv = B::svref_2object($stash);
+    $keep{$k} = $stash if UNIVERSAL::isa($sv, 'B::HV') and $sv->NAME;
+  }
+
   delete_package $class;
+  no strict 'refs';
+  *{"$class\::$_"} = $keep{$_} for keys %keep;
 }
```

It's a bit ugly, but all the tests pass.

PR #1126, opened 2017-08-29
Pass arguments for filters

KES777:
### Summary
Sometimes we are required to check incoming IDs not only for format `->like( qr/\d+/ )`, but also for existence in DB.
Having this patch we can provide model name for filter.

### Motivation
These changes allow to pass arguments for filters

### References
no reference

**UPD**
I did not link [blog post](http://blogs.perl.org/users/eugen_konkov/2017/08/how-to-pass-arguments-for-mojolicious-filter.html) because description did not ask about that:

>### References
>LIST RELEVANT ISSUES, PULL REQUESTS AND IRC/MAILING-LIST DISCUSSIONS HERE

Thank you @simbabque

### DIFF
--- a/lib/Mojolicious/Validator/Validation.pm
+++ b/lib/Mojolicious/Validator/Validation.pm
@@ -70,9 +70,10 @@ sub optional {
 
   my @input = ref $input eq 'ARRAY' ? @$input : ($input);
   my $filters_cb =  $self->validator->filters;
-  for my $cb ( @filters) {
-    $cb =  $filters_cb->{$cb};
-    @input = map { $self->$cb($name, $_) } @input;
+  for my $cb (@filters) {
+    ($cb, my @args) =  ref $cb ? @$cb : $cb;
+    $cb = $filters_cb->{$cb};
+    @input = map { $self->$cb($name, $_, @args) } @input;
   }
   $self->output->{$name} = ref $input eq 'ARRAY' ? \@input : $input[0]
     if @input && !grep { !length } @input;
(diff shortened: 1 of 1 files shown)


kraih:
No tests, no documentation, no explanation, i'm afraid this pull request has to be rejected right away. http://mojolicious.org/perldoc/Mojolicious/Guides/Contributing#CONTRIBUTING-CODE

simbabque:
To their defense, they did write a [blog post](http://blogs.perl.org/users/eugen_konkov/2017/08/how-to-pass-arguments-for-mojolicious-filter.html) about this. It's not documentation, but it's a start.

kraih:
@simbabque The description of this pull request specifically states `References: no reference`.

KES777:
by previous experience you always afraid =(
Here is just idea how to improve your excellent framework.
If you like this idea any of core developers may make the candy: more test, more documentation, more explanation.

Thank you for Mojolicious ;-)

kraih:
@KES777 If you really believe this can some day reach our requirements then i can also leave it open as a `work in progress` indefinitely.

kraih:
@KES777 You will have to change the title to something descriptive however, since "Small improvement" is completely useless.

marcusramberg:
I think it would be better if @KES777 reopens this as a complete pull request with tests and documentation if he wants to pursue this feature further.

KES777:
At least you may cherry pick this commit:https://github.com/kraih/mojo/pull/1126/commits/c4d30a47c4eb2085baac72bef37237708e1a81be which removes useless method calls in loop

PR #1128, opened 2017-09-03
Improve Code Contrast in PODRenderer and Project Site

merrilymeredith:
### Summary
Some of the colors in prettified code sections don't meet a minimum suggested contrast of 4.5:1 for accessibility, this PR bumps the colors in the light theme -- used in PODRenderer and the project website but not debug pages -- to that minimum.  A second commit removes stripes.png, which is 1px white/8% gray vertical stripes that hamper readability and can cause artifacts on some displays, for a solid average of the two.  The dark theme on the debug page doesn't have this issue.

### Motivation
Accessibility is important for new and current users and we should take complaints about it very seriously.  Not everyone has good sight, or will stay young, or can correct their vision, or afford the nicest equipment.  Comments like "con crit please" and "wtf did you do to your DE" are unwelcoming, inconsiderate, etc.

### References
http://code-of-conduct.mojolicious.org/
https://irclog.perlgeek.de/mojo/2017-09-02#i_15107586
https://irclog.perlgeek.de/mojo/2017-09-03#i_15107708

### Rendered Changes
Firefox 52 ESR, Windows 10, 96 dpi:
contrast-example


kraih:
A theme with more contrast might make sense, but i don't like the removal of the background. 👎

marcusramberg:
I agree with kraih here, support the contrast part, not the removal of the background. 🌮

jberger:
I don't have any opinion. Per our usual convention, this PR hasn't reached consensus and has since "aged out". I think we are in agreement that style is up for discussion so continue on IRC or forks and please open a new PR for the next iteration. Thanks!

kraih:
We now have a theme with a little more contrast.

PR #1129, opened 2017-09-06
fix SSL_ca_file reference to undef

mudler:
### Summary
since IO::Socket::SSL >= 1.965 `\undef` means no CA, that is specificed as `''` in older versions

### Motivation
Otherwise we observe errors like:
> Mojo::Reactor::Poll: I/O watcher failed: SSL_ca_file SCALAR(0x2490188) does not exist at /usr/lib/perl5/vendor_perl/5.18.2/IO/Socket/SSL.pm line 1616.

### References
https://progress.opensuse.org/issues/23934#change-62720

### DIFF
--- a/lib/Mojo/IOLoop/TLS.pm
+++ b/lib/Mojo/IOLoop/TLS.pm
@@ -9,6 +9,9 @@ use Scalar::Util 'weaken';
 use constant TLS => $ENV{MOJO_NO_TLS}
   ? 0
   : eval 'use IO::Socket::SSL 1.94 (); 1';
+use constant DEFAULT => eval { IO::Socket::SSL->VERSION('1.965') }
+  ? ''
+  : \undef;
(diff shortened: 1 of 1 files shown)


lib/Mojo/IOLoop/TLS.pm
```diff
@@ -9,6 +9,9 @@ use Scalar::Util 'weaken';
 use constant TLS => $ENV{MOJO_NO_TLS}
   ? 0
   : eval 'use IO::Socket::SSL 1.94 (); 1';
+use constant DEFAULT => eval { IO::Socket::SSL->VERSION("1.965") }
```
kraih: We use single quotes whenever possible.

mudler: right, i actually do that too, don't know why i used double quotes there

PR #1132, opened 2017-09-14
Permit Mojo::Util::tablify to work on non-rectangular arrays.

lindleyw:
### Summary
Should a non-rectangular array (one in which not all rows have equal numbers of columns) be passed to Mojo::Util::tablify, the resulting columns or entirely empty rows will be skipped rather than producing errors from missing sprintf() arguments.

### Motivation
Users who write simple classes based on Mojolicious::Command should be able to produce arrays for table output, perhaps from a variety of functions or a JSON structure, without having to ensure that each row has an equal number of columns.

### References
https://irclog.perlgeek.de/mojo/2017-09-14#i_15164351

### DIFF
--- a/lib/Mojo/Util.pm
+++ b/lib/Mojo/Util.pm
@@ -282,8 +282,8 @@ sub tablify {
     }
   }
 
-  my @formats = (map({"\%-${_}s"} @spec[0 .. $#spec - 1]),'%s');
-  return join '', map { sprintf join(' ', @formats[0..scalar @$_ - 1]) . "\n", @$_ } @$rows;
+  my @formats = map({"\%-${_}s"} @spec[0 .. $#spec - 1]), '%s';
+  return join '', map { sprintf join('  ', @formats[0 .. scalar @$_ - 1]) . "\n", @$_ } @$rows;
 }
 
 sub term_escape {
--- a/lib/Mojo/Util.pm
+++ b/lib/Mojo/Util.pm
@@ -284,7 +284,7 @@ sub tablify {
 
   my @formats = (map({"\%-${_}s"} @spec[0 .. $#spec - 1]), '%s');
   return join '',
-    map { sprintf join('  ', @formats[0 .. scalar @$_ - 1]) . "\n", @$_ }
+    map { sprintf join('  ', @formats[0 .. $#$_]) . "\n", @$_ }
     @$rows;
 }
 
--- a/lib/Mojo/Util.pm
+++ b/lib/Mojo/Util.pm
@@ -282,7 +282,7 @@ sub tablify {
     }
   }
 
-  my @formats = map({"\%-${_}s"} @spec[0 .. $#spec - 1]), '%s';
+  my @formats = (map({"\%-${_}s"} @spec[0 .. $#spec - 1]), '%s');
   return join '', map { sprintf join('  ', @formats[0 .. scalar @$_ - 1]) . "\n", @$_ } @$rows;
 }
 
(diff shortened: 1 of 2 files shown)


jberger:
Well clearly you need to make sure it passes existing tests. Once you've done that can you please add a test and verify the formatting; that second line looks longer than the character limit.

lib/Mojo/Util.pm
```diff
@@ -282,8 +282,8 @@ sub tablify {
     }
   }
 
-  my $format = join '  ', map({"\%-${_}s"} @spec[0 .. $#spec - 1]), '%s';
-  return join '', map { sprintf "$format\n", @$_ } @$rows;
+  my @formats = (map({"\%-${_}s"} @spec[0 .. $#spec - 1]),'%s');
```
Grinnz: Parentheses around the map are unnecessary. This should be run through perltidy.

lib/Mojo/Util.pm
```diff
@@ -282,8 +282,8 @@ sub tablify {
...
+  return join '', map { sprintf join(' ', @formats[0..scalar @$_ - 1]) . "\n", @$_ } @$rows;
```
Grinnz: Columns should be joined on 2 spaces, this appears to be the cause of test failures. `@formats[0..scalar @$_ - 1]` can be shortened to `@formats[0..$#$_]`

lindleyw:
I am still unsure why: `$ TEST_UNIX=1 prove t/mojo` never reports these failures on my system locally.

jhthorsen:
I'm neutral to the change, though I can't come up with a case myself where I would give `tablify()` "non-rectangular arrays".

jberger:
I lean slightly with @jhthorsen, especially since the data could be rectangularlized fairly easily by the user (push to unfilled rows with `''`). This is compounded by because this is positional (rather than say named) the empty columns must come at the end (empty columns in the middle will have to be rectangularized manually in order to make sense).

That said, if it can be made more flexible and DWIM without any cost, I don't see any reason to oppose (once the PR works and tests are added).

jhthorsen:
@lindleyw: Could you please provide a real life example for when you don't have rectangular data?

latk:
@lindleyw Running `prove t/mojo` will use the *installed* version of Mojo. Use the `-l`flag to add the `lib` directory to the test script's `@INC`. In general, `prove` is invoked as `prove -lr t`. See the [prove docs](https://metacpan.org/pod/prove) for more details.

Since this is a MakeMaker based dist, you can also run the tests as `perl Makefile.PL && make test`.

kraih:
No need to add extra lines.
```diff
diff --git a/lib/Mojo/Util.pm b/lib/Mojo/Util.pm
index c6c2e3491..5438d1621 100644
--- a/lib/Mojo/Util.pm
+++ b/lib/Mojo/Util.pm
@@ -282,8 +282,8 @@ sub tablify {
     }
   }

-  my $format = join '  ', map({"\%-${_}s"} @spec[0 .. $#spec - 1]), '%s';
-  return join '', map { sprintf "$format\n", @$_ } @$rows;
+  my @fs = (map({"\%-${_}s"} @spec[0 .. $#spec - 1]), '%s');
+  return join '', map { sprintf join('  ', @fs[0 .. $#$_]) . "\n", @$_ } @$rows;
 }

 sub term_escape {
```

kraih:
Has this pull request been abandoned?

lindleyw:
I was away for awhile. Back now; will proceed.

kraih:
I kinda wanted to merge this for the last release, but the pull request has still not been fixed. ☹️

kraih:
Resolved by @1150.

Issue #1134, opened 2017-10-02
Warnings when encoding to MIME-Header


Tekki:
* Mojolicious version: 7.46
* Perl version: 5.26.0 (Perlbrew)
* Operating system: Debian Jessie

### Steps to reproduce the behavior

Encode a name to MIME-Header:
```
use strict;
use warnings;
use utf8;
use feature 'say';

use Mojo::Util 'encode';

my $name = 'Rolf Stöckli';
say encode 'MIME-Header', $name;
```

### Expected behavior

Should print

`=?UTF-8?B?Um9sZiBTdMO2Y2tsaQ==?=`

### Actual behavior

Prints a lot of warnings
```
Use of uninitialized value in subroutine entry at /home/tekki/perl5/perlbrew/perls/perl-5.26.0/lib/5.26.0/x86_64-linux/Encode/MIME/Header.pm line 243.
...
Use of uninitialized value in subroutine entry at /home/tekki/perl5/perlbrew/perls/perl-5.26.0/lib/5.26.0/x86_64-linux/Encode/MIME/Header.pm line 243.
=?UTF-8?B?Um9sZiBTdMO2Y2tsaQ==?=
```

Grinnz:
Mojo::Util::encode is just a wrapper for Encode. The following script has the same behavior.

```
use strict;
use warnings;
use utf8;
use feature 'say';

use Encode 'find_encoding';

my $name = 'Rolf Stöckli';
say find_encoding('MIME-Header')->encode($name);
```

dotandimet:
Yep, looks like Perl 5.26 is the source of the warnings.

Tekki:
If you `use Encode 'encode';` there are no warnings.
I'm not sure why, maybe because of `$check ||= 0;` at https://github.com/dankogai/p5-encode/blob/master/Encode.pm#L177.

Tekki:
@Grinnz The following script gives no warnings on 5.26, and works fine on 5.14 and 5.20:
```
use strict;
use warnings;
use utf8;
use feature 'say';

use Encode 'find_encoding';

my $name = 'Rolf Stöckli';
say find_encoding('MIME-Header')->encode($name, 0);
```
Why not call it this way?

kraih:
I guess we might just as well pass along the default if some encoding modules require it. https://github.com/kraih/mojo/commit/29f56b09a1eb5735ecb99a3aa5f7b2854cfa3e86

PR #1135, opened 2017-10-03
Fix warnings with Encode >= 2.87

augensalat:
### Summary

With `Encode` 2.87 or higher `Mojo::Util::encode('MIME-Header', $string)`
issues a warning for every character in `$string`.

### Motivation

Keeping my errorlog sparse. ;-)

### References
.

### DIFF
--- a/lib/Mojo/Util.pm
+++ b/lib/Mojo/Util.pm
@@ -127,7 +127,7 @@ sub dumper {
   Data::Dumper->new([@_])->Indent(1)->Sortkeys(1)->Terse(1)->Useqq(1)->Dump;
 }
 
-sub encode { _encoding($_[0])->encode("$_[1]") }
+sub encode { _encoding($_[0])->encode("$_[1]", 0) }
 
 sub extract_usage {
   my $file = @_ ? "$_[0]" : (caller)[1];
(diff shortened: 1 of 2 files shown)


jberger:
Though it isn't mentioned in the PR, it appears that this is meant to address #1135 . I have just a couple questions. Has this second option ($check) been available for a long time and suddenly it emits warnings (ie, does this support sufficiently old versions of Encode to unconditionally include $check = 0). Second, as this appears to be an optional value to encode, has an issue been opened on that project (Encode itself) to document the regression and if so can it be linked here?

Tekki:
(... meant to address #1134)
Even in 2.00, $check in Encode::encode has a default value of 0. It's Encode::MIME::Header that at some place started to emit these warnings. There doesn't seem to be an open issue for this problem.

augensalat:
I have to admit that I did not check the open issues in advance. Just stumbled about the problem today and fixed it.

Grinnz:
I am not sure silencing warnings for all encode operations just because of this bug in Encode::MIME::Header is a good idea.

jberger:
But if we don't allow passing `$check` through, and the default is `0` then isn't that effectively the same as passing a `0` in unconditionally?

jberger:
I believe this is closed by https://github.com/kraih/mojo/commit/29f56b09a1eb5735ecb99a3aa5f7b2854cfa3e86

PR #1136, opened 2017-10-09
add link to Perl Tutorial on the Perl Maven site

szabgab:
If you think my tutorial is worth recommending to the Mojolicious newbies.


jberger:
Thank you for thinking of us. After discussion, if we were to change the recommended perl tutorials we feel that we would owe it to our users to assess the state of the field and currently no one has the time to do so. Our current recommendations have served our readers well and therefore we have decided to keep them as is for the time being.

PR #1149, opened 2017-11-03
ojo_signatures.t: Suppress proxies

cpansprout:
Without this, I get a test failure because it’s trying to ask a proxy
for localhost, which of course comes back with the wrong content.


kraih:
Thanks, applied.

PR #1151, opened 2017-11-07
Export "with" at "use Mojo::Base -base"

aferreira:
### Summary
Now that Mojo::Base has support for roles with
```
use Mojo::Base -role;
```
it would be nice to have "with" exported to the caller package – alongside "has" – when doing `use Mojo::Base -base` . For example:
```
package Foo;
use Mojo::Base -role;
after foo => sub { say "Bye." }

package FooFoo;
use Mojo::Base -base;
with 'Foo';         # comes from Mojo::Base just like "has"
sub foo { say "Foo" }

package main;
FooFoo->new->foo;
# Foo
# Bye.
```

### Motivation
In the same spirit that `use Mojo::Base -role` allows to create roles without mentioning `Role::Tiny`, this change allows to consume roles (statically) without mentioning `Role::Tiny`.

So instead of
```
package Foo;
use Mojo::Base -base;
use Role::Tiny::With;
with 'SomeRole';
```
one can write
```
package Foo;
use Mojo::Base -base;
with 'SomeRole';
```

### References
None

### DIFF
--- a/lib/Mojo/Base.pm
+++ b/lib/Mojo/Base.pm
@@ -81,11 +81,14 @@ sub import {
     require "$file.pm";
   }
 
-  # "has" and possibly ISA
+  # Exports and possibly ISA
   if ($flags[0]) {
     no strict 'refs';
     push @{"${caller}::ISA"}, $flags[0] unless $flags[0] eq '-role';
-    Mojo::Util::monkey_patch($caller, 'has', sub { attr($caller, @_) });
+    my %subs = (has => sub { attr($caller, @_) });
+    $subs{with} = sub { Role::Tiny->apply_roles_to_package($caller, @_) }
+      unless !ROLES || $flags[0] eq '-role';
+    Mojo::Util::monkey_patch($caller, %subs);
   }
 
   # Mojo modules are strict!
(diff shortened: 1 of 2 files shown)


Grinnz:
It would be simpler to export Role::Tiny::With to the caller. However either would require Role::Tiny to become a hard dependency of Mojolicious. So instead I would just suggest users to use Role::Tiny::With for this functionality.

jberger:
I don't love the notion of exporting more keywords for existing flags. That said here is a possible alternate proposal, adding `-with => 'My::Cool::Role'` to the import http://paste.debian.net/994498/ . Overall I don't see the need for "with" as acutely as "-role" simply for the fact that the "has" behavior was harder to get otherwise, while as @Grinnz mentions, `Role::Tiny::With` exists for this purpose.

jberger:
Then again, as @Grinnz mentioned on IRC, because composition would happen at compile-time in an import statement that wouldn't allow for run-time satisfying of required methods, say via "has". So my alternative is out.

kraih:
👎 This doesn't add all that much value and the code is not very pretty.

kraih:
To be more specific, i can think of two good reasons for adding stuff to `Mojo::Base`. **a)** A feature that makes it easier to extend Mojolicious and we can show in the docs, like [role support](http://mojolicious.org/perldoc/Mojolicious/Guides/Testing#Extending-Test::Mojo) we have now, or **b)** Something we actually use in the core distribution. Neither is the case here.

kraih:
I'm afraid it looks like this pull request did not pass the vote. Thank you for your contribution.

PR #1153, opened 2017-11-07
use class_to_path to load modules in Mojo::Base

Grinnz:
### Summary
With circular require issues solved, class_to_path can be reused in Mojo::Base for loading modules.

### Motivation
Code reuse

### DIFF
--- a/lib/Mojo/Base.pm
+++ b/lib/Mojo/Base.pm
@@ -76,9 +76,8 @@ sub import {
   }
 
   # Module
-  elsif ((my $file = $flags[0]) && !$flags[0]->can('new')) {
-    $file =~ s!::|'!/!g;
-    require "$file.pm";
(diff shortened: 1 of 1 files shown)


lib/Mojo/Base.pm
```diff
@@ -76,9 +76,8 @@ sub import {
   }
 
   # Module
-  elsif ((my $file = $flags[0]) && !$flags[0]->can('new')) {
-    $file =~ s!::|'!/!g;
-    require "$file.pm";
+  elsif ($flags[0] && !$flags[0]->can('new')) {
+    require(Mojo::Util::class_to_path($flags[0]));
```
Grinnz: Parentheses are needed for Mojo::Util::class_to_path to avoid circular require issues. Parentheses are needed for require to avoid weird syntax errors possibly related to indirect object syntax.

PR #1154, opened 2017-11-13
n_map - eval on fixed size parcels of colllections

kiwiroy:
### Summary
Evaluate callback for, or call a method on, each `$n` elements in the collection and create a new collection from the results. The most basic use case would be tabling a flat collection.

### Motivation
This might be a useful addition to `Mojo::Collection`.

### References
none

### DIFF
--- a/lib/Mojo/Collection.pm
+++ b/lib/Mojo/Collection.pm
@@ -52,6 +52,14 @@ sub map {
   return $self->new(map { $_->$cb(@_) } @$self);
 }
 
+sub n_map {
+  my ($self, $n, $cb) = (shift, shift, shift);
+  return
+    $self->new(map  { $_->$cb(@_) }
+               map  { $self->new(@$self[$_..List::Util::min($_+$n-1, $#$self)]) }
+               grep { 0 == $_ % $n } 0 .. $#$self);
+}
+
 sub new {
   my $class = shift;
   return bless [@_], ref $class || $class;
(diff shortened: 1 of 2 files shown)


kraih:
Are there any common use cases?

Grinnz:
Another name for a method like this could be 'bundle', similar to https://metacpan.org/pod/List::UtilsBy#bundle_by.

jberger:
As to use cases, I could probably use it vs pairmap here: https://github.com/jberger/Mojo-Chrome/blob/8d0daab7972d6666e3a5534125815e1a30edaae6/lib/Mojo/Chrome.pm#L129 but I'm not sure that I actually would nor that that is enough.

IMHO I think this and other List::Utils and UtilsBy could be encouraged to start as Roles and if any prove extra popular/handy we can consider adding them to the core collection at that point.

Grinnz:
I've put together a role to add most of the functions from List::UtilsBy. https://github.com/Grinnz/Mojo-Collection-Role-UtilsBy

kraih:
Without good use cases i have to vote 👎. And i like the `UtilsBy` role.

kiwiroy:
This started some awesome conversation and code from @Grinnz. I'm closing this as roles are the real solution here. Thanks.

Grinnz:
The mentioned role has been released to CPAN. https://metacpan.org/pod/Mojo::Collection::Role::UtilsBy

PR #1156, opened 2017-11-16
small doc improvements

tim-2:
### Summary
- add "num" to default registered validation checks in doc
- correct link in Mojo::IOLoop doc


kraih:
Thanks, i was about to apply your pull request, but then you added a second, questionable change that will have to be discussed on IRC first.

PR #1157, opened 2017-11-19
Initialize $r to avoid "Use of uninitialized value" error

rouzier:
### Summary
When running the get command the errors appear
```
Use of uninitialized value $r in vec at /usr/share/perl5/vendor_perl/Mojolicious/Command/get.pm line 19.
Use of uninitialized value $r in scalar assignment at /usr/share/perl5/vendor_perl/Mojolicious/Command/get.pm line 19.
```
### Motivation
To avoid noise when testing mojo from the cli.

### DIFF
--- a/lib/Mojolicious/Command/get.pm
+++ b/lib/Mojolicious/Command/get.pm
@@ -16,7 +16,7 @@ sub run {
   my ($self, @args) = @_;
 
   # Data from STDIN
-  vec(my $r, fileno(STDIN), 1) = 1;
+  vec(my $r = '', fileno(STDIN), 1) = 1;
   my $in = !-t STDIN && select($r, undef, undef, 0) ? join '', <STDIN> : undef;
 
   my $ua = Mojo::UserAgent->new(ioloop => Mojo::IOLoop->singleton);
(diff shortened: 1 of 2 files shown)


kraih:
Can't confirm this problem.

rouzier:
This happens on perl 5.16.3 on centos

rouzier:
I can provide a test for this.

lib/Mojolicious/Command/get.pm
```diff
@@ -16,7 +16,8 @@ sub run {
   my ($self, @args) = @_;
 
   # Data from STDIN
-  vec(my $r, fileno(STDIN), 1) = 1;
+  my $r = '';
+  vec($r, fileno(STDIN), 1) = 1;
```
kraih: Why not `vec(my $r = '', fileno(STDIN), 1) = 1;`?

rouzier: That would work also

rouzier: done

kraih:
Thanks. We don't normally care about warnings on [legacy versions of Perl](http://mojolicious.org/perldoc/Mojolicious/Guides/FAQ#Which-versions-of-Perl-are-supported-by-Mojolicious), but the fix here was very minimal so i went ahead anyway. I couldn't merge the test however. https://github.com/kraih/mojo/commit/35fc9d6e7be52695a29fce523c9e523d158b6529

PR #1161, opened 2017-12-06
Only spawn Net::DNS::Native threads if Mojo::IOLoop::Client is used

Grinnz:
### Summary
Currently, anything loading Mojo::IOLoop will automatically spawn a Net::DNS::Native thread pool (if it's installed, and $ENV{MOJO_NO_NNR} isn't set). With this change, the thread pool is only spawned once something calls ->connect on a Mojo::IOLoop::Client object, where Net::DNS::Native is used.

Example: `perl -MMojo::IOLoop -E'sleep 5' & sleep 2 && pgrep -aw perl`

### Motivation
It seems unnecessary to spawn thread pools for processes that will never need them, like many Mojolicious applications that only serve content.


jberger:
Are there any fork-safety or alternatively fork-optimization (early loading) considerations here? If not then this seems like an easy win.

kraih:
I think there might be potential for fork/thread bugs to become harder to debug, since threads would be started much later in the lifecycle of mojo processes. But i'm not aware of a concrete example, and this pull request passed the vote, so i'll merge it for now.

PR #1163, opened 2017-12-13
Don't set content-length for websocket calls

forki:
### Summary
Don't set content-length for websocket calls

### Motivation
ASP.NET Core is rejecting mojo websocket calls right now.

### References

https://github.com/kraih/mojo/issues/828

### DIFF
--- a/lib/Mojo/Message.pm
+++ b/lib/Mojo/Message.pm
@@ -106,8 +106,12 @@ sub fix_headers {
   }
   else {
     my $size = $self->body_size
-    if (!$is_websocket || $size > 0) { $headers->content_length($size) }
-    else { $headers->remove('Content-Length') }
+    if (!$is_websocket || $size > 0) { 
+      $headers->content_length($size) 
+    }
+    else { 
+      $headers->remove('Content-Length')
+    }
   }
   return $self;  
 }
--- a/lib/Mojo/Message.pm
+++ b/lib/Mojo/Message.pm
@@ -104,7 +104,7 @@ sub fix_headers {
   }
   else {
     my $size = $self->body_size;
-    my $is_websocket = defined $headers->sec_websocket_version; 
+    my $is_websocket = defined $headers->sec_websocket_version || defined $headers->sec_websocket_accept; 
     if (!$is_websocket || $size > 0) {
       $headers->content_length($size) 
     }
--- a/lib/Mojo/Message.pm
+++ b/lib/Mojo/Message.pm
@@ -101,7 +101,9 @@ sub fix_headers {
   my $is_websocket = defined $websocket;
   if ($content->is_multipart) { $headers->remove('Content-Length') }
   elsif ($content->is_chunked || $headers->content_length) { return $self }
-  if   ($content->is_dynamic) { $headers->connection('close') }
+  if ($content->is_dynamic) { 
+    $headers->connection('close') 
+  }
   else {
     my $size = $self->body_size
     if (!$is_websocket || $size > 0) { $headers->content_length($size) }
--- a/lib/Mojo/Message.pm
+++ b/lib/Mojo/Message.pm
@@ -104,8 +104,7 @@ sub fix_headers {
   }
   else {
     my $size = $self->body_size;
-    my $websocket = $headers->sec_websocket_accept;
-    my $is_websocket = defined $websocket;
+    my $is_websocket = $headers->sec_websocket_version;
     if (!$is_websocket || $size > 0) { 
       $headers->content_length($size) 
     }
--- a/lib/Mojo/Message.pm
+++ b/lib/Mojo/Message.pm
@@ -103,7 +103,7 @@ sub fix_headers {
   if ($content->is_multipart) { $headers->remove('Content-Length') }
   elsif ($content->is_chunked || $headers->content_length) { return $self }
   if   ($content->is_dynamic) { $headers->connection('close') }
-  elsif (!$is_websocket || $size > 0) { $headers->content_length($self->body_size) }
+  elsif (!$is_websocket || $size > 0) { $headers->content_length($size) }
   else { $headers->remove('Content-Length') }
 
   return $self;
--- a/lib/Mojo/Message.pm
+++ b/lib/Mojo/Message.pm
@@ -104,7 +104,7 @@ sub fix_headers {
   }
   else {
     my $size = $self->body_size;
-    my $is_websocket = $headers->sec_websocket_version;
+    my $is_websocket = $headers->sec_websocket_version > 0;
     if (!$is_websocket || $size > 0) { 
       $headers->content_length($size) 
     }
--- a/lib/Mojo/Message.pm
+++ b/lib/Mojo/Message.pm
@@ -104,12 +104,11 @@ sub fix_headers {
   }
   else {
     my $size = $self->body_size;
-    my $websocket_version = $headers->sec_websocket_version;
-    my $is_websocket = false;
-    if (definded $websocket_version) {
-      $is_websocket = $websocket_version > 0;
+    my $websocket_version = 0;
+    if (definded $headers->sec_websocket_version) {
+      $websocket_version = $headers->sec_websocket_version;
     }
-    if (!$is_websocket || $size > 0) { 
+    if ($websocket_version > 0 || $size > 0) { 
       $headers->content_length($size) 
     }
     else { 
--- a/lib/Mojo/Message.pm
+++ b/lib/Mojo/Message.pm
@@ -99,14 +99,16 @@ sub fix_headers {
   my $headers = $content->headers;
   my $websocket = $headers->sec_websocket_accept;
   my $is_websocket = defined $websocket;
-  
   if ($content->is_multipart) { $headers->remove('Content-Length') }
   elsif ($content->is_chunked || $headers->content_length) { return $self }
   if   ($content->is_dynamic) { $headers->connection('close') }
-  elsif (!$is_websocket || (my $size = $self->body_size)) { $headers->content_length($size) }
-  else { $headers->remove('Content-Length') }
+  else {
+    my $size = $self->body_size
+    if (!$is_websocket || $size > 0) { $headers->content_length($size) }
+    else { $headers->remove('Content-Length') }
 
-  return $self;
+    return $self;
+  }
 }
 
 sub get_body_chunk {
--- a/lib/Mojo/Message.pm
+++ b/lib/Mojo/Message.pm
@@ -104,7 +104,8 @@ sub fix_headers {
   }
   else {
     my $size = $self->body_size;
-    if (not defined $headers->sec_websocket_version || $size > 0) {
+    my $is_websocket = defined $headers->sec_websocket_version; 
+    if (!$is_websocket || $size > 0) {
       $headers->content_length($size) 
     }
     else { 
--- a/lib/Mojo/Message.pm
+++ b/lib/Mojo/Message.pm
@@ -104,7 +104,11 @@ sub fix_headers {
   }
   else {
     my $size = $self->body_size;
-    my $is_websocket = $headers->sec_websocket_version > 0;
+    my $websocket_version = $headers->sec_websocket_version;
+    my $is_websocket = false;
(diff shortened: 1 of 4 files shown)


forki:
I also created https://github.com/aspnet/KestrelHttpServer/pull/2217 for server-side.

lib/Mojo/Message.pm
```diff
@@ -97,10 +97,14 @@ sub fix_headers {
...
   my $headers = $content->headers;
+  my $websocket = $headers->sec_websocket_accept;
+  my $isWebsocket = defined $websocket;
+  my $size = $self->body_size;  
   if ($content->is_multipart) { $headers->remove('Content-Length') }
   elsif ($content->is_chunked || $headers->content_length) { return $self }
   if   ($content->is_dynamic) { $headers->connection('close') }
-  else                        { $headers->content_length($self->body_size) }
+  elsif (!$isWebsocket || $size > 0) { $headers->content_length($self->body_size) }
+  else { $headers->remove('Content-Length') }
```
jhthorsen: Pretty sure this code won't be merged, because of the naming convention and lack of tidying. Maybe something like this would be accepted instead?

```
  if ($content->is_dynamic) {
    $headers->connection('close');
  }
  elsif (!defined $headers->sec_websocket_accept || $self->body_size > 0) {
    $headers->content_length($self->body_size);
  }
```

forki: looks pretty much like the code above!? I tried to keep it as similar as possible

jhthorsen:
I think this PR makes sense based on https://tools.ietf.org/html/rfc7230#section-3.3.2

> A server MUST NOT send a Content-Length header field in any response with a status code of 1xx (Informational) or 204 (No Content).

forki:
@jhthorsen this part of the spec talks about "a server" - we are trying to patch websocket clients. (is that the right place?)

lib/Mojo/Message.pm
```diff
@@ -97,10 +97,14 @@ sub fix_headers {
   # Content-Length or Connection (unless chunked transfer encoding is used)
   my $content = $self->content;
   my $headers = $content->headers;
+  my $websocket = $headers->sec_websocket_accept;
+  my $isWebsocket = defined $websocket;
```
olegwtf: perl style of variables naming is `$is_websocket`, not `$isWebsocket`

forki: done. thx

forki:
according to @part-timeDev we hit that code part but tcpdump reveals that still Content-Length is sent

lib/Mojo/Message.pm
```diff
@@ -97,10 +97,14 @@ sub fix_headers {
   # Content-Length or Connection (unless chunked transfer encoding is used)
   my $content = $self->content;
   my $headers = $content->headers;
+  my $websocket = $headers->sec_websocket_accept;
+  my $is_websocket = defined $websocket;
```
kraih: This line does not appear to serve a real purpose.

lib/Mojo/Message.pm
```diff
@@ -97,10 +97,14 @@ sub fix_headers {
...
+  my $size = $self->body_size;
```
kraih: You're calling `body_size` twice now.

lib/Mojo/Message.pm
forki: no. it was for clearer intent. feel free to suggest a diffent way

lib/Mojo/Message.pm
forki: fixed

kraih: And calling it here does have a performance cost for some cases.

forki: Does 62138b8 work? Sorry I never did any perl. This is just trying to help a vendor to connect to our backend

forki:
@kraih is there another place that sets the content-length? According to @part-timeDev we hit that code, but tcpdump still shows that naugthy header

lib/Mojo/Message.pm
forki: ok fixed that

forki:
Ok. I'm done. It now works for us. Would appreciate merge and release. Thanks everyone

lib/Mojo/Message.pm
```diff
@@ -99,10 +99,20 @@ sub fix_headers {
...
+    $headers->connection('close') 
+  }
+  else {
+    my $size = $self->body_size;
+    my $is_websocket = defined $headers->sec_websocket_version || defined $headers->sec_websocket_accept; 
+    if (!$is_websocket || $size > 0) {
+      $headers->content_length($size) 
+    }
+    else { 
+      $headers->remove('Content-Length')
```
jhthorsen: Do you need to remove the Content-Length header?

forki: that's the whole point here. to get rid of it

jhthorsen: But is the header there at this point? I thought `content_length()` was not called in this case...

forki:
Look above in same method. There is yet another version of remove. I copied
that.

Am 13.12.2017 17:07 schrieb "Jan Henning Thorsen" <notifications@github.com
>:

*@jhthorsen* commented on this pull request.
------------------------------

In lib/Mojo/Message.pm
<https://github.com/kraih/mojo/pull/1163#discussion_r156703268>:

> @@ -99,10 +99,20 @@ sub fix_headers {
   my $headers = $content->headers;
   if ($content->is_multipart) { $headers->remove('Content-Length') }
   elsif ($content->is_chunked || $headers->content_length) { return $self }
-  if   ($content->is_dynamic) { $headers->connection('close') }
-  else                        { $headers->content_length($self->body_size) }
-
-  return $self;
+  if ($content->is_dynamic) {
+    $headers->connection('close')
+  }
+  else {
+    my $size = $self->body_size;
+    my $is_websocket = defined $headers->sec_websocket_version ||
defined $headers->sec_websocket_accept;
+    if (!$is_websocket || $size > 0) {
+      $headers->content_length($size)
+    }
+    else {
+      $headers->remove('Content-Length')

But is the header there at this point? I thought content_length() was not
called in this case...

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/kraih/mojo/pull/1163#discussion_r156703268>, or mute
the thread

forki:
Look above in same method. There is yet another version of remove. I copied
that.

Am 13.12.2017 17:07 schrieb "Jan Henning Thorsen" <notifications@github.com

kraih:
@jhthorsen I think that is an actual bug in Mojolicious. Unrelated to this pull request though.

forki:
@kraih do I need to do anything else?

kraih:
@jhthorsen Oh, actually i missed recent changes to this pull request that added WebSocket handshake response handling. So now i guess it is related and might block this pull request, since response handling is done incorrectly.

kraih:
@forki Yes, you need to correctly handle handshake responses now. Which i assume will require small changes to `Mojo::Message::Reponse`.

kraih:
@forki I'd also like to see an RFC link explaining the situation with WebSocket handshake requests and `Content-Length: 0`.

forki:
> Yes, you need to correctly handle handshake responses now

what do you mean exactly?

> see an RFC link explaining the situation with WebSocket handshake requests

O_o what link and where?

forki:
something like https://tools.ietf.org/html/rfc7230#page-30

       A server MUST NOT send a Content-Length header field in any response
   with a status code of 1xx (Informational) or 204 (No Content).  A
   server MUST NOT send a Content-Length header field in any 2xx
   (Successful) response to a CONNECT request (Section 4.3.6 of
   [RFC7231]).

kraih:
@forki **a)** The link @jhthorsen posted earlier shows that 1xx responses (and 204) MUST NOT have a `Content-Length` header. So that's a bug in the HTTP stack that if fixed takes care of WebSocket handshake responses too. **b)** That is my question for you, when we make changes to the HTTP/WebSocket stack the primary focus is spec compliance, so we generally require a quote from a related spec that shows the mistake a pull request is meant to fix.

forki:
@kraih IMHO the spec is unclear about that. that's what I tried to explain to you guys and the asp.net server guys like @davidfowl - so I sent 2 PRs. One PR to the server (see aspnet/KestrelHttpServer#2217) to understand mojo and one PR to mojo to be compatible with existing asp.net core servers.  

@davidfowl said he'd like to know how other servers behave. I assume you run websockets against couple of different servers? node? IIS? apache?

forki:
scratch that. @part-timeDev found it in https://tools.ietf.org/html/rfc7230#page-30

    A user agent SHOULD send a Content-Length in a request message when
    no Transfer-Encoding is sent and the request method defines a meaning
    for an enclosed payload body.  For example, a Content-Length header
    field is normally sent in a POST request even when the value is 0
    (indicating an empty payload body).  A user agent SHOULD NOT send a
    Content-Length header field when the request message does not contain
    a payload body and the method semantics do not anticipate such a
    body.

So mojo is not spec compliant

kraih:
@forki Well, SHOULD is merely a recommendation, not a requirement. Handling every request method separately (especially with extensions like WebDav) seems rather tedious. So i'm actually not sure if it's one worth following. Now i'm also curious what other HTTP clients default to and if they handle request methods differently.

forki:
@kraih yes I reworded my comment. I read it now as mojo should not send it, but if it does the server needs to accept it anyway.

> Now i'm also curious what other HTTP clients default to 

tbh mojo is the first client that I have heard of that sends content-length for GET requests. That just makes no sense.

kraih:
I'm open to changing the default to not sending `Content-Length` headers if there's more advantages than disadvantages. But i think we can conclude that this should not be a WebSocket specific change, as intended by this pull request.

forki:
@kraih no it shoud not be websocket specific. But I want to point you to https://github.com/kraih/mojo/issues/828#issuecomment-130419425 - so basically you already tried that change and reverted it. So what I did was just to limit it to websockets where it actually showed issues in the real world

kraih:
@forki Well, that's a bummer, no idea what to do then. But making it WebSocket specific would feel very wrong.

khellang:
It would be interesting to know _why_ a985ac7cfe2488320431cf696cd2020bc3c9df90 "caused problems with Chrome"... It's been over 2 years. That's quite a while in browser-land :smile:

forki:
also "caused problems with Chrome" clearly talks about server side. we need to patch the client side of mojo.

forki:
@kraih I changed the PR to remove the header for GET requests

kraih:
@khellang Afraid i don't remember specifics, but it might actually have been Chrome that got it right. In the IRC logs from the time i did find references to the RFC that look suspicious. https://tools.ietf.org/html/rfc7230#section-3.3.3
>    7.  Otherwise, this is a response message without a declared message
>       body length, so the message body length is determined by the
 >      number of octets received prior to the server closing the
 >     connection.

lib/Mojo/Message/Request.pm
```diff
@@ -77,6 +77,12 @@ sub fix_headers {
   return $self unless my $info = $proxy->userinfo;
   $headers->proxy_authorization('Basic ' . b64_encode($info, ''))
     unless $headers->proxy_authorization;
+
+  # RFC 7230 3.3.2. suggests not to send Content-Length
+  if (self->method == 'GET') {
```
kraih: This is not Perl.

kraih: In fact, this is completely wrong in every way.

kraih:
Afraid i don't see this pull request going anywhere, we really need to discuss what the correct behaviour would be first before worrying about implementation. Please open an issue for that or join the IRC channel so we can have that discussion there.

forki:
Can you please reopen this? I really appreciate your input and I'm trying to fix it. But without CI it's hard

kraih:
Just run the tests locally.
```
$ git clone git@github.com:kraih/mojo.git
$ cd mojo
$ perl Makefile.PL; make test; make clean
```

forki:
I don't even have a perl, CI worked fine until you closed it. I just want to help you guys.

PR #1166, opened 2017-12-14
Be more compliant with RFC7280 regarding Content-Length

jberger:
Addresses #1163 by removing the `Content-Length` from the request when there is no body. There are still some concerns with this fix. The `body_size` method can be expensive. Also when considering the line of the RFC this addresses:

> A user agent SHOULD NOT send a Content-Length header field when the request message does not contain a payload body and the method semantics do not anticipate such a body.

This PR only addresses the first part and does not consider the portion that states "and the method semantics do not anticipate such a body".

### DIFF
--- a/lib/Mojo/Message/Request.pm
+++ b/lib/Mojo/Message/Request.pm
@@ -72,6 +72,9 @@ sub fix_headers {
     $headers->authorization('Basic ' . b64_encode($info, ''));
   }
 
+  # RFC 7230 3.3.2
+  $headers->remove('Content-Length') if ($headers->content_length // '') eq '0';
+
   # Basic proxy authentication
   return $self unless (my $proxy = $self->proxy) && $self->via_proxy;
   return $self unless my $info = $proxy->userinfo;
--- a/t/mojo/request.t
+++ b/t/mojo/request.t
@@ -49,6 +49,7 @@ is $req->version,           '1.1', 'right version';
 is $req->url,               '/', 'right URL';
 is $req->cookie('a'), undef, 'no value';
 is $req->body, '', 'no content';
+ok !defined $req->headers->content_length, '"Content-Length" does not exist';
 
 # Parse HTTP 1.1 message with huge "Cookie" header exceeding line limit
 # (alternative)
@@ -64,6 +65,7 @@ is $req->version,           '1.1', 'right version';
 is $req->url,               '/', 'right URL';
 is $req->cookie('a'), undef, 'no value';
 is $req->body, '', 'no content';
+ok !defined $req->headers->content_length, '"Content-Length" does not exist';
 
 # Parse HTTP 1.1 message with content exceeding line limit
 $req = Mojo::Message::Request->new;
@@ -98,6 +100,7 @@ is $req->version,           '1.1', 'right version';
 is $req->url,               '/', 'right URL';
 is $req->headers->header('Foo'), undef, 'no "Foo" value';
 is $req->body, '', 'no content';
+ok !defined $req->headers->content_length, '"Content-Length" does not exist';
 
 # Parse broken HTTP 1.1 message with start-line exceeding line limit
 $req = Mojo::Message::Request->new;
@@ -111,6 +114,7 @@ is $req->version, '1.1', 'right version';
 is $req->url,     '',    'no URL';
 is $req->cookie('a'), undef, 'no value';
 is $req->body, '', 'no content';
+ok !defined $req->headers->content_length, '"Content-Length" does not exist';
 
 # Parse broken HTTP 1.1 message with start-line exceeding line limit
 # (alternative)
@@ -124,6 +128,7 @@ is $req->version, '1.1', 'right version';
 is $req->url,     '',    'no URL';
 is $req->cookie('a'), undef, 'no value';
 is $req->body, '', 'no content';
+ok !defined $req->headers->content_length, '"Content-Length" does not exist';
 
 # Parse pipelined HTTP 1.1 messages exceeding leftover limit
 $req = Mojo::Message::Request->new;
@@ -265,6 +270,7 @@ is $req->headers->upgrade, 'websocket', 'right "Upgrade" value';
 is $req->headers->sec_websocket_key, 'abcdef=',
   'right "Sec-WebSocket-Key" value';
 is $req->body, '', 'no content';
+ok !defined $req->headers->content_length, '"Content-Length" does not exist';
 
 # Parse HTTP 1.0 start-line and headers, no body
 $req = Mojo::Message::Request->new;
@@ -1248,10 +1254,10 @@ is $req->method,      'GET', 'right method';
 is $req->version,     '1.1', 'right version';
 is $req->url,         '/demo', 'right URL';
 is $req->url->to_abs, 'http://example.com/demo', 'right absolute URL';
-is $req->headers->connection,     'Upgrade',     'right "Connection" value';
-is $req->headers->upgrade,        'websocket',   'right "Upgrade" value';
-is $req->headers->host,           'example.com', 'right "Host" value';
-is $req->headers->content_length, 0,             'right "Content-Length" value';
+is $req->headers->connection, 'Upgrade',     'right "Connection" value';
+is $req->headers->upgrade,    'websocket',   'right "Upgrade" value';
+is $req->headers->host,       'example.com', 'right "Host" value';
+ok !defined $req->headers->content_length, '"Content-Length" does not exist';
 is $req->headers->sec_websocket_accept, 'abcdef=',
   'right "Sec-WebSocket-Key" value';
 is $req->headers->sec_websocket_protocol, 'sample',
@@ -1276,10 +1282,10 @@ is $req->method,      'GET', 'right method';
 is $req->version,     '1.1', 'right version';
 is $req->url,         '/demo', 'right URL';
 is $req->url->to_abs, 'http://example.com/demo', 'right absolute URL';
-is $req->headers->connection,     'Upgrade',     'right "Connection" value';
-is $req->headers->upgrade,        'websocket',   'right "Upgrade" value';
-is $req->headers->host,           'example.com', 'right "Host" value';
-is $req->headers->content_length, 0,             'right "Content-Length" value';
+is $req->headers->connection, 'Upgrade',     'right "Connection" value';
+is $req->headers->upgrade,    'websocket',   'right "Upgrade" value';
+is $req->headers->host,       'example.com', 'right "Host" value';
+ok !defined $req->headers->content_length, '"Content-Length" does not exist';
 is $req->headers->sec_websocket_accept, 'abcdef=',
   'right "Sec-WebSocket-Key" value';
 is $req->headers->sec_websocket_protocol, 'sample',
@@ -1292,10 +1298,10 @@ is $clone->method,      'GET', 'right method';
 is $clone->version,     '1.1', 'right version';
 is $clone->url,         '/demo', 'right URL';
 is $clone->url->to_abs, 'http://example.com/demo', 'right absolute URL';
-is $clone->headers->connection,   'Upgrade',     'right "Connection" value';
-is $clone->headers->upgrade,      'websocket',   'right "Upgrade" value';
-is $clone->headers->host,         'example.com', 'right "Host" value';
-is $req->headers->content_length, 0,             'right "Content-Length" value';
+is $clone->headers->connection, 'Upgrade',     'right "Connection" value';
+is $clone->headers->upgrade,    'websocket',   'right "Upgrade" value';
+is $clone->headers->host,       'example.com', 'right "Host" value';
+ok !defined $clone->headers->content_length, '"Content-Length" does not exist';
 is $clone->headers->sec_websocket_accept, 'abcdef=',
   'right "Sec-WebSocket-Key" value';
 is $clone->headers->sec_websocket_protocol, 'sample',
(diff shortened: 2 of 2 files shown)


forki:
> and the method semantics do not anticipate such a body.

I still think you should sent the header in POST requests. That's why I checked for method eq GET in my PR.

t/mojo/request.t
```diff
@@ -49,6 +49,7 @@ is $req->version,           '1.1', 'right version';
 is $req->url,               '/', 'right URL';
 is $req->cookie('a'), undef, 'no value';
 is $req->body, '', 'no content';
+ok !defined $req->headers->content_length, '"Content-Length" does not exist';
```
jhthorsen: Would it be more robust to do the following..?
```
is int(@{$req->headers->every_header('Content-Length')}), 0, '"Content-Length" does not exist';
```

lib/Mojo/Message/Request.pm
```diff
@@ -72,6 +72,9 @@ sub fix_headers {
     $headers->authorization('Basic ' . b64_encode($info, ''));
   }
 
+  # RFC 7230 3.3.2
+  $headers->remove('Content-Length') if ($headers->content_length // '') eq '0';
```
jhthorsen: I agree with @forki that you should also check if it's a GET request. And why do you do `// ''` instead of `// 0`?

kraih:
I think what's good enough for curl is good enough for us, `curl -v -X POST http://mojolicious.org` does not send a `Content-Length: 0` header.

forki:
@kraih directly above the cited text we can read:

    A user agent SHOULD send a Content-Length in a request message when
    no Transfer-Encoding is sent and the request method defines a meaning
    for an enclosed payload body. For example, a Content-Length header
    field is normally sent in a POST request even when the value is 0
    (indicating an empty payload body). 

I think this makes it very clear that they propose to send it.

kraih:
@forki And that's merely a SHOULD, that other very popular HTTP implementations don't follow.

forki:
I know. but the other one is also a SHOULD NOT. This is a weird discussion tbh. The RFC proposes how a client should behave. Of course you can do whatever you want. It's your project. But these suggestions actually make sense IMHO. In the GET case you can't have a body, so sending a Content-Length makes no sense. In POST, the method semantics usually use the body, so sending the Content-Length is useful. Even sending Content-Length 0, because then server may or may not shortcut things because it does not need to look at the body stream.

I mean our problem is solved in both cases. If this PR gets through then we will be very happy.

jhthorsen:
@kraih: What about...

```
$ curl -v --data-raw "" http://localhost:3000
...
> Content-Length: 0
```

kraih:
It doesn't look like we will be able to reach consensus on this, perhaps we should leave everything the way it was for now. At least until a strong enough argument for one of the proposed changes emerges.

jhthorsen:
I'm pro getting rid of `Content-Length: 0` for GET.

forki:
there are 2 strong arguments: 
1) spec is very clear on what it favors
2) it increases compat with a server technology

kraih:
@jhthorsen And about...
```
$ curl -v -X GET --data-raw "" http://mojolicious.org
...
> Content-Length: 0
```

khellang:
Wouldn't removing `Content-Length` for `GET` requests only make this change _less_ intrusive in addition to following the spec recommendations more closely? I don't see why this is even a discussion 😄

forki:
@kraih 

    $ curl -v -X GET --data-raw "" http://mojolicious.org

curl is actually smart here and sends a POST in that case. So it's consistent with what we discussed.

kraih:
@forki Not with curl 7.54.0 or 7.57.0.

kraih:
@khellang Then we also have to ask the question what the real goal here is. Why would we pick a single HTTP method and add a workaround specifically for it, instead of compiling a list of all affected HTTP methods to get actual spec compliance?

forki:
@kraih yes wrong code in clipboard. I confused myself and removed the comment.

@kraih and I think I now leave the discussion. I tried to help with making your lib support yet another server and keep it minimal invasive. You closed WIP PR over syntax issues. Fine. It's your project.
This was my first excursion into perl ecosystem. Can't say it was a nice trip. I'm really disappointed right now.

Well, on the other hand kestrel is clearly violating spec and I hope they have more interest in getting things fixed. So bye.

khellang:
> Why would we pick a single HTTP method and add a workaround specifically for it, instead of compiling a list of all affected HTTP methods to get actual spec compliance?

Depends how pragmatic you want to be. Either way, you'd be better off than the status quo :smile:

Checking for `GET` would probably get you 90% there. As RFC 7231 states:

> GET is the primary mechanism of information retrieval and the focus of almost all performance optimizations.

If you want to compile a list of every single HTTP method where "the method semantics do not anticipate [...] a body" that would be even better, but will probably require more work for little value.

kraih:
@forki Spec compliance is hard, that has nothing to do with the Perl ecosystem. I closed your WIP PR because you were just throwing stuff at the wall to see what sticks, instead of participating in a discussion first to figure out how to solve the problem correctly. It just wasn't very productive, and neither are your snide remarks about those trying hard here to find a solution.

t/mojo/request.t
kraih: Why?

jhthorsen: I guess "Why?" is the same as "No" 😄

kraih:
@khellang You really don't think it would be very odd to generate different requests for GET and HEAD by default?

khellang:
> You really don't think it would be very odd to generate different requests for GET and HEAD by default?

Yes, I think it would be weird. But I think it's equally weird to send `Content-Length: 0` for `GET`, `HEAD` and `OPTIONS` requests, or omit it for `POST`, `PUT` and `PATCH`.

kraih:
I think we can conclude that the proposed solution here is definitely incorrect and not what we want to do.

kraih:
But, we also do not want to treat GET and HEAD requests differently. So we have no alternative to move ahead with either.

jberger:
I kept being troubled by the [quote](https://tools.ietf.org/html/rfc7230#section-3.3.2)

> and the method semantics do not anticipate such a body

I wanted to understand what it meant and see if I could find any such definitions. After much reading of specs, I'm going to lay out some findings.

The closest thing I can find to support for removing the Content-Length header comes from [RFC 7230 3.3](https://tools.ietf.org/html/rfc7230#section-3.3)

>    The presence of a message body in a request is signaled by a
>    Content-Length or Transfer-Encoding header field.  Request message
>    framing is independent of method semantics, even if the method does
>    not define any use for a message body.

However this would therefore also include any method that does not include a body, including say POST which this thread has already shown people prefer to have. Note that this section replaced the obsoleted [RFC 2616 4.3](https://tools.ietf.org/html/rfc2616#section-4.3) states that

>    A server SHOULD
>    read and forward a message-body on any request; if the request method
>    does not include defined semantics for an entity-body, then the
>    message-body SHOULD be ignored when handling the request.

which would have given stronger credence to removing the Content-Length for any such request, however as it was removed in the new version, I think that is an indication that any method that may have a body is allowed to.

Later when considering how new methods might be added in the future [RFC 7231 8.1.2](https://tools.ietf.org/html/rfc7231#section-8.1.2) states (referring back to the previous section)

>    Since message parsing (Section 3.3 of [RFC7230]) needs to be
>    independent of method semantics (aside from responses to HEAD),
>    definitions of new methods cannot change the parsing algorithm or
>    prohibit the presence of a message body on either the request or the
>    response message.  Definitions of new methods can specify that only a
>    zero-length message body is allowed by requiring a Content-Length
>    header field with a value of "0".

This states outright that method parsing ought to be independent of the method semantics. It further states that new methods that require empty bodies must require the Content-Length header and that the value be 0. This indicates for me that this behavior of setting such a header is encouraged.

I still wanted to understand what "do not anticipate such a body" meant though so I continued looking. The methods themselves are defined in [RFC 7231 4.3](https://tools.ietf.org/html/rfc7231#section-4.3). The definitions of GET and HEAD include the statement

>    A payload within a GET request message has no defined semantics;
>    sending a payload body on a GET request might cause some existing
>    implementations to reject the request.

This is a far cry from saying that the body is not anticipated, just that it has no meaning. Indeed servers like Elasticsearch [allow](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-body.html) such requests. When thinking about it, this usage makes sense. If you'd like to submit a safe request (as defined by [RFC 7231 4.2.1](https://tools.ietf.org/html/rfc7231#section-4.2.1)), as one wants when simply searching, GET is the clear choice. The fact that any such body has no defined meaning does not imply that it is without use, clearly it is for being safe, and additionally it is not required that the server ignore it. No this cannot be the meaning of "not anticipated".

So the question is then, what is? There is only one request method with prohibited from having a body, TRACE. As seen in [RFC 7231 4.3.8](https://tools.ietf.org/html/rfc7231#section-4.3.8).

> A client MUST NOT send a message body in a TRACE request.

This is the kind of language I would expect to see when stating that a body is not anticipated. I can see why a useragent SHOULD NOT send a Content-Length in such an event, because it certainly would not be anticipated. The contrast of that definition with the definition of GET is clear.

With all that stated I can say that there are two remaining implications for Mojo. 

First, if we feel it is worthwhile, Mojolicious should remove Content-Length from TRACE requests and forbid them to have message bodies.

Second, we may consider removing Content-Length headers for messages that do not have bodies, however, I see no reason to treat GET and HEAD (and OPTIONS incidentally) and differently than POST, PUT, and DELETE in that regard. After all, as quoted from [RFC 7230 3.3](https://tools.ietf.org/html/rfc7230#section-3.3) before

>    Request message
>    framing is independent of method semantics, even if the method does
>    not define any use for a message body.

---

In regards to interpersonal dynamics at play so far in this PR and the previous one, let me say, @forki while we appreciate your bringing this to our attention and we do hope that your problem is resolved, whether here or on the server side, Pull Requests are not intended as workspaces for running code. If you must work that way, feel free to enable Travis on your clone. There are many people that subscribe to our repository and giving them all notifications so that you may experiment is not fair to them.

Additionally, it is not encouraging for us who work on Mojolicious to hear that you will not give us the courtesy of installing the interpreter and cloning the repo to test (we have worked hard on making the dependencies effectively nil) while you expect us to give you our time and effort to consider your changes. Certainly you can see an asymmetry there?

Now I have put in hours of reading RFCs and writing code to evaluate this problem and attempt a sane mitigation. I see that [Kestral](https://github.com/aspnet/KestrelHttpServer/pull/2217) is on the matter too. I hope that any future correspondence from your shows an equal amount of effort on your part or at least appreciation for those of us who have. Specifically any future references to what the RFC "clearly state" should come with clear quotations that refute my above. 

This response is not intended in any way as a personal attack. Indeed quite the opposite. This is stating that this situation is taken very seriously and is being dealt with as professionally as possible. I would hope you would treat us as such too.

I'm going to leave this PR open for a little while to encourage discussion of my above points, however, without clear RFC evidence or very strong indication that the "living standard" of HTTP is markedly and decidedly different, and widely implemented as such, I don't see this PR being merged.

khellang:
Just Googling "content-length zero" makes my head spin 😖 

There's a myriad of issues, StackOverflow posts and forum threads for browsers, client libraries and servers.
Some servers expect `Content-Length: 0`, some reject it. Some clients omit it (for some or all HTTP methods), some include it.
There's tons of references to different RFCs, both old and new. Some use strong language, some are very vague.

I don't think we'll be able to reach consensus here. Let's hope MS will merge the fix on the server-side 🤞

forki:
sorry @jberger that's just not true. I cloned the repo and tried to run it locally. It did not work for me, but you are right I could probably just spent more time to get it work eventually. But I still tried to find a way to participate and was actively working on a solution to the problem. Regarding assymetry: I spent a lot of time to look through the code base to pick up your style. I was learning. /shrug
I don't even use your project myself.

I actually sent WIP PPs to both sides. Server and Client in order to bring you guys together and find a way to resolve the incompat.

This is the first project that I'm trying to work on that did not appreciate WIP PRs. If you look an my github profile then you will see that I'm working on a couple of projects. I guess that policy is OK but it's new to me.

jberger:
@forki I'm sorry to hear that it didn't work for you. If you were having trouble, you could have asked us, we want to make things easy.

As to the "not appreciate WIP PRs": actually, yes our project is more stringent than most, a PR is usually assumed to be in complete state unless explicitly said so. That being said, we closed the PR on the 17th commit, and one that would (a) not have compiled and (b) would have broken GET requests with bodies; we didn't know how to take such a request and so we closed it.

Your reply to the effect of asking us to reopen the PR so you could continue to use the CI was our first indication that you weren't running the tests locally (or at least the first time I realized it). I have similarly been involved in open source for a long time and have many project involvements and it was the first time I had seen that.

Look I'm not trying to be unreasonable, we would have been happy to help get your perl and mojo installed. We still would though now that I've done this research I don't see it helping your chances of getting a fix included. Code is no longer the issue, RFC compliance is.

---

@khellang yeah, I've been doing those searches too. Amazing isn't it? The problem here of course is that if we change something, soon the next person will have some other, likely incompatible edge case. Say a server that requires Content-Length on all posts. We would have to justify this change for why we broken their application and without RFCs I couldn't do so. Seeing as Kestral appears to be making moves to correct the (actually) clear non-compliance on their end, I think everyone can walk away satisfied.

forki:
Since I'm trying to find a positive outcome here. Just a suggestion: maybe
add a contribution guideline. I actually looked for one but didn't find one
immediately.
So instead of just complaining about an issue I tried to get my hands dirty
and investigate myself. And as you said: code was never the issue.

Am 15.12.2017 17:56 schrieb "Joel Berger" <notifications@github.com>:

jberger:
Like perhaps in the [Contributing Guide](http://mojolicious.org/perldoc/Mojolicious/Guides/Contributing#CONTRIBUTING-CODE) which is linked to by Github in a yellow box when opening a pull request? I'm sorry if you didn't see it, I'm not sure how much clearer it could be. If this conversation doesn't come back to technical merits quickly I think it might be about time to wrap it up.

forki:
Yes. Did not see it. Looked into readme. But thanks.

Am 15.12.2017 18:41 schrieb "Joel Berger" <notifications@github.com>:

jberger:
With no technical rebuttal forthcoming, I'm going to close this at least for now

forki:
Just to let you know: my PR to Kestrel was merged so things are going to be resolved Server-side. Sorry for all the trouble

jberger:
It was a really interesting thing to see how hard it can be to interpret the RFCs on seemingly such a simple issue. I'm glad to hear that your root problem will be solved in the end. Cheers and good luck!

PR #1168, opened 2017-12-16
Update Util.pm

lucas1:
### Summary
Added conditional Operator with function defined

### Motivation
Use of uninitialized value $_[1] in string at /usr/local/share/perl5/Mojo/Util.pm line 134.

### DIFF
--- a/lib/Mojo/Util.pm
+++ b/lib/Mojo/Util.pm
@@ -131,7 +131,7 @@ sub dumper {
   Data::Dumper->new([@_])->Indent(1)->Sortkeys(1)->Terse(1)->Useqq(1)->Dump;
 }
 
-sub encode { _encoding($_[0])->encode("$_[1]", 0) }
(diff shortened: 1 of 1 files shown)


jhthorsen:
I’m not sure if I see the point of adding this check. I see what it does, it’s  just that I don’t see it as encode()s responsibility.

Also, there’s no tests and no symmetry with other functions.

CandyAngel:
This silences any indication that undef may have been unintentionally passed in. As the output seems to be the same before/after the patch (empty string), I'd prefer to see the warning than not, if I'm not doing my own checks.

kraih:
If that's the intention then my vote is 👎.

lucas1:
I found error, I'm using module Mojo::Redis2, when value is undef return this error:
my $token = $self->req->headers->authorization
$self->redis->get($token);

kraih:
I interpret that as confirmation that the error message was useful and silencing it would actually be counter productive.

PR #1171, opened 2017-12-19
With "use Mojo::Base -role", "has" is a non-method

aferreira:
Proposed solution to https://github.com/kraih/mojo/issues/1170

This pull request delays promoting the caller package into a role to after exporting `has`. That makes `has` to be marked as a non-method and then treated just like `with`, `requires`, `after`, `before` and `around`.

### DIFF
--- a/lib/Mojo/Base.pm
+++ b/lib/Mojo/Base.pm
@@ -72,7 +72,6 @@ sub import {
   # Role
   elsif ($flags[0] eq '-role') {
     Carp::croak 'Role::Tiny 2.000001+ is required for roles' unless ROLES;
-    eval "package $caller; use Role::Tiny; 1" or die $@;
   }
 
   # Module
@@ -82,9 +81,16 @@ sub import {
 
   # "has" and possibly ISA
   if ($flags[0]) {
-    no strict 'refs';
-    push @{"${caller}::ISA"}, $flags[0] unless $flags[0] eq '-role';
     Mojo::Util::monkey_patch($caller, 'has', sub { attr($caller, @_) });
+    if ($flags[0] eq '-role') {
+
+      # import of Role::Tiny must come after injecting has
+      eval "package $caller; use Role::Tiny; 1" or die $@;
+    }
+    else {
+      no strict 'refs';
+      push @{"${caller}::ISA"}, $flags[0] unless $flags[0] eq '-role';
+    }
   }
 
   # Mojo modules are strict!
--- a/t/mojo/roles.t
+++ b/t/mojo/roles.t
@@ -97,6 +97,13 @@ is $obj7->name,    'Joel',         'base attribute';
 is $obj7->whisper, 'psst, joel',   'method from first role';
 is $obj7->hello,   'HEY! JOEL!!!', 'method from second role';
 
+# Multiple Mojo::Base roles
+my $obj8 = Mojo::RoleTest->with_roles('+quiet', 'Mojo::RoleTest::Hello')
+  ->new(name => 'Joel');
+is $obj8->name,    'Joel',        'base attribute';
+is $obj8->whisper, 'psst, joel',  'method from first role';
+is $obj8->hello,   'hello mojo!', 'method from second role';
+
 # Classes that are not subclasses of Mojo::Base
 my $stream = Mojo::ByteStream->with_roles('Mojo::RoleTest::Hello')->new;
 is $stream->hello, 'hello mojo!', 'right result';


jberger:
Thanks for pointing out the flaw, the solution, and some tests. I'm going to take a stab at making the change less intrusive to the code.

jberger:
I think something like

```diff
diff --git a/lib/Mojo/Base.pm b/lib/Mojo/Base.pm
index a2b113a00..b13d2cb2e 100644
--- a/lib/Mojo/Base.pm
+++ b/lib/Mojo/Base.pm
@@ -72,7 +72,6 @@ sub import {
   # Role
   elsif ($flags[0] eq '-role') {
     Carp::croak 'Role::Tiny 2.000001+ is required for roles' unless ROLES;
-    eval "package $caller; use Role::Tiny; 1" or die $@;
   }
 
   # Module
@@ -82,9 +81,16 @@ sub import {
 
   # "has" and possibly ISA
   if ($flags[0]) {
-    no strict 'refs';
-    push @{"${caller}::ISA"}, $flags[0] unless $flags[0] eq '-role';
     Mojo::Util::monkey_patch($caller, 'has', sub { attr($caller, @_) });
+    if ($flags[0] eq '-role') {
+
+      # import of Role::Tiny must come after injecting has
+      eval "package $caller; use Role::Tiny; 1" or die $@;
+    }
+    else {
+      no strict 'refs';
+      push @{"${caller}::ISA"}, $flags[0] unless $flags[0] eq '-role';
+    }
   }
 
   # Mojo modules are strict!
diff --git a/t/mojo/roles.t b/t/mojo/roles.t
index ca13dabb1..00255ad2d 100644
--- a/t/mojo/roles.t
+++ b/t/mojo/roles.t
@@ -97,6 +97,13 @@ is $obj7->name,    'Joel',         'base attribute';
 is $obj7->whisper, 'psst, joel',   'method from first role';
 is $obj7->hello,   'HEY! JOEL!!!', 'method from second role';
 
+# Multiple Mojo::Base roles
+my $obj8 = Mojo::RoleTest->with_roles('+quiet', 'Mojo::RoleTest::Hello')
+  ->new(name => 'Joel');
+is $obj8->name,    'Joel',        'base attribute';
+is $obj8->whisper, 'psst, joel',  'method from first role';
+is $obj8->hello,   'hello mojo!', 'method from second role';
+
 # Classes that are not subclasses of Mojo::Base
 my $stream = Mojo::ByteStream->with_roles('Mojo::RoleTest::Hello')->new;
 is $stream->hello, 'hello mojo!', 'right result';
```

Would be more in keeping with what already exists. Assuming other agree, could you either squash that into your PR or change your PR to submit just the failing tests and I'll append the fix?

kraih:
I don't think either solutions are ideal. There has to be a cleaner way to fix this.

lib/Mojo/Base.pm
```diff
@@ -72,7 +72,6 @@ sub import {
   # Role
   elsif ($flags[0] eq '-role') {
     Carp::croak 'Role::Tiny 2.000001+ is required for roles' unless ROLES;
-    eval "package $caller; use Role::Tiny; 1" or die $@;
   }
```
marcusramberg: Think this block should go away and the croak should be moved down into the new if -roles check below.

lib/Mojo/Base.pm
```diff
@@ -82,9 +81,16 @@ sub import {
...
     Mojo::Util::monkey_patch($caller, 'has', sub { attr($caller, @_) });
+    if ($flags[0] eq '-role') {
+
+      # import of Role::Tiny must come after injecting has
+      eval "package $caller; use Role::Tiny; 1" or die $@;
+    }
+    else {
+      no strict 'refs';
+      push @{"${caller}::ISA"}, $flags[0] unless $flags[0] eq '-role';
+    }
```
marcusramberg: The unless here seems superflous since the unless already indicates the lack of '-role'?

jhthorsen:
What about this?

```perl
diff --git a/lib/Mojo/Base.pm b/lib/Mojo/Base.pm
index a2b113a00..717df706f 100644
--- a/lib/Mojo/Base.pm
+++ b/lib/Mojo/Base.pm
@@ -66,24 +66,18 @@ sub import {
   # Base
   if ($flags[0] eq '-base') { $flags[0] = $class }

-  # Strict
-  elsif ($flags[0] eq '-strict') { $flags[0] = undef }
-
   # Role
-  elsif ($flags[0] eq '-role') {
+  if ($flags[0] eq '-role') {
+    Mojo::Util::monkey_patch($caller, 'has', sub { attr($caller, @_) });
     Carp::croak 'Role::Tiny 2.000001+ is required for roles' unless ROLES;
     eval "package $caller; use Role::Tiny; 1" or die $@;
   }

-  # Module
-  elsif ($flags[0] && !$flags[0]->can('new')) {
-    require(Mojo::Util::class_to_path($flags[0]));
-  }
-
-  # "has" and possibly ISA
-  if ($flags[0]) {
+  # Module and not -strict
+  elsif ($flags[0] !~ /^-/) {
     no strict 'refs';
-    push @{"${caller}::ISA"}, $flags[0] unless $flags[0] eq '-role';
+    require(Mojo::Util::class_to_path($flags[0])) unless $flags[0]->can('new');
+    push @{"${caller}::ISA"}, $flags[0];
     Mojo::Util::monkey_patch($caller, 'has', sub { attr($caller, @_) });
   }
```

jhthorsen:
Closed by a9d214d

aferreira:
Thanks @jhthorsen Your solution is a lot more elegant.

PR #1173, opened 2017-12-21
JSON encode utf8 flagged scalars as strings

haarg:
Values that start as numbers won't get the utf8 flag enabled implicitly
through other operations, so it is a good signal that we should encode
as a string.  Additionally, this avoids planned future perl warnings
about performing bitwise operations on strings containing wide
characters.

Discussion regarding deprecation warning:
https://www.nntp.perl.org/group/perl.perl5.porters/2017/12/msg248248.html

### DIFF
--- a/lib/Mojo/JSON.pm
+++ b/lib/Mojo/JSON.pm
@@ -251,7 +251,8 @@ sub _encode_value {
   # Number
   no warnings 'numeric';
   return $value
-    if length((my $dummy = '') & $value)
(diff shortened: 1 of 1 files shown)


Grinnz:
Note that this is actually one of the correct usages of is_utf8: to inspect the internal state of the string.

Grinnz:
this has also been applied to JSON::PP: https://github.com/makamaka/JSON-PP/pull/35

PR #1176, opened 2017-12-28
Morbo fixups

mohawk2:
### Summary
Current `t/mojo/morbo.t` has a bug: it does not test the local code, but the installed code instead.

### Motivation
Tests should test the correct thing. Also I have added a couple of doc snippets.

### References
I found this while having a look at making Morbo event-driven, as discussed on IRC.

### DIFF
--- a/lib/Mojo/Server/Morbo.pm
+++ b/lib/Mojo/Server/Morbo.pm
@@ -127,6 +127,10 @@ L<Mojo::Server::Morbo> implements the following attributes.
   $morbo      = $morbo->backend(Mojo::Server::Morbo::Backend::Poll->new);
 
 Backend, usually a L<Mojo::Server::Morbo::Backend::Poll> object.
+Control this by setting the environment variable
+C<MOJO_MORBO_BACKEND>. Its value will be appended to
+C<Mojo::Server::Morbo::Backend::>, and the class of that name needs to
+implement the L<Mojo::Server::Morbo::Backend> interface.
 
 =head2 daemon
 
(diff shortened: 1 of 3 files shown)


kraih:
Afraid this pull request has to be rejected because of the unexplained documentation changes. Making an environment variable part of the public API is a pretty big deal. I really don't like it when new features that require discussion are hidden in completely unrelated pull requests.

kraih:
The proposed fix also appears to be wrong, everything works as expected for me already.

mohawk2:
You're right about the module-finding thing. Proof: I switched to system Perl, which does not have Mojolicious installed. Works fine:

```shell
make && TEST_MORBO=1 MORBO_VERBOSE=1 prove -b t/mojo/morbo.t
```

Does not work, because `blib` only affects current process, instead of environment variable so not child processes:

```shell
make && TEST_MORBO=1 MORBO_VERBOSE=1 perl -Mblib t/mojo/morbo.t
```

Does work by setting relevant environment variables, including preserving value of `PERL5LIB` for `perlbrew`:

```shell
make && TEST_MORBO=1 MORBO_VERBOSE=1 PERL5LIB=blib/lib:blib/arch${PERL5LIB:+:$PERL5LIB} perl t/mojo/morbo.t
```

I would note that adding `-Mblib` to the command line in `morbo.t` *guarantees* it will use the `blib` version, which seems like a worthwhile thing to do. What do you think?

**Note**, running the above shows that the `MORBO_VERBOSE` output is swallowed, and I argue that this means the switching in `morbo.t` from `-|` to `|-` is the right thing to do. What do you think?

kraih:
Afraid i do not follow your reasoning at all, everybody just uses `TEST_MORBO=1 prove -l t/mojo/morbo.t` and that works perfectly fine.

mohawk2:
As discussed on IRC just now, it certainly will work. However, it swallows any stdout, such as that produced when `MORBO_VERBOSE` is true.

PR #1177, opened 2017-12-31
Mojo::Promise->all to take non-Promise objects

mohawk2:
### Summary
Currently `Mojo::Promise->all` cannot take non-Promise objects as an argument. This is surprising for those who've used JavaScript Promises, whose `all` method can do so.

### Motivation
This current behaviour does not match the expectations of those who've used JS Promises. There, the expectation is for accepting non-Promises, and for interoperation.

### References
As discussed on IRC: https://irclog.perlgeek.de/mojo/2017-12-30#i_15633913

### DIFF
--- a/lib/Mojo/Promise.pm
+++ b/lib/Mojo/Promise.pm
@@ -20,14 +20,20 @@ sub all {
   my $results   = [];
   my $remaining = scalar @promises;
   for my $i (0 .. $#promises) {
-    $promises[$i]->then(
-      sub {
-        $results->[$i] = [@_];
-        $all->resolve(@$results) if --$remaining <= 0;
-      },
-      sub { $all->reject(@_) }
-    );
+    if ($promises[$i] && blessed $promises[$i] && $promises[$i]->can('then')) {
+      $promises[$i]->then(
+        sub {
+          $results->[$i] = [@_];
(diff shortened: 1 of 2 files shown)


mohawk2:
Updated and force-pushed.

lib/Mojo/Promise.pm
```diff
@@ -20,14 +20,20 @@ sub all {
...
+          $all->resolve(@$results) if --$remaining <= 0;
+        },
+        sub { $all->reject(@_) }
+      );
+    } else {
+      $results->[$i] = [$promises[$i]];
+      --$remaining;
+    }
   }
+  $all->resolve(@$results) if $remaining <= 0; # all non-promises
```
kraih: This would only work with the global `Mojo::IOLoop` singleton, since there is no chance to change the value of the `ioloop` attribute before a `next_tick` is initiated.

kraih: And that means i believe for this pull request to be accepted we would have to deprecate the `ioloop` attribute first.

kraih:
This has been discussed at length on IRC and we could not reach a clear decision on how to move forward. The feature is not required by Promises/A+, its usefulness questionable and the correct implementation very hard. I think it might be best to reject the feature for now.

PR #1179, opened 2018-01-03
Promise->all,race reprecated as allows non-global ioloop

mohawk2:
### Summary
`all` and `race` need to be instance methods since blocking in `Mojo::UserAgent` requires `ioloop` attribute to be controllable.

### Motivation
Keep M::UA workable. Also allow non-then-ables as `all`/`race`-able entities, since the `->resolve` when all are a non-then-able takes place before any `ioloop` override is possible.

### References
Discussion today on IRC channel.

### DIFF
--- a/Changes
+++ b/Changes
@@ -1,5 +1,7 @@
 
 7.61  2018-01-03
+  - Un-deprecated use of Mojo::Promise::all and Mojo::Promise::race as instance
+    methods.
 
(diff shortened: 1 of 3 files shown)


kraih:
I don't think we actually want to do this, `$promise->all($promise2, $promise3)` just looks wrong.

Issue #1185, opened 2018-01-30
Mojo::Cookie::Response does not set the domain parameter according to RFC 6265


expbbc:
* Mojolicious version: 7.61
* Perl version: 5.24
* Operating system: Debian Stretch / Ubuntu Xenial

### Steps to reproduce the behavior
1. Use Mojo::UserAgent to retrieve a cookie from a site that does not set the `domain` parameter.
2. Test the domain parameter of this cookie in the UA's cookie jar.

### Expected behavior
The cookie should have a host_only flag set, and have its domain set as specified in RFC 6265 Section 5.3 bullet point 6: https://tools.ietf.org/html/rfc6265#section-5.3

>    6.   If the domain-attribute is non-empty: … Otherwise … Set the cookie's domain to the canonicalized request-host

### Actual behavior
An extra 'origin' attribute was introduced to attempt to fix this problem which holds the original source of the cookie: https://github.com/kraih/mojo/commit/8e1895893360fcde403df19e51035f0cdb72477f

This introduces an additional bug however, in that two cookies with the same name from the same domain are considered distinct due to the unset 'domain' parameter: https://github.com/kraih/mojo/blob/master/t/mojo/cookiejar.t#L289 — RFC 6265 Section 5.3 bullet point 11 indicates these conflict and the old one should be removed from the store.

> Let old-cookie be the existing cookie with the same name, domain, and path as the newly created cookie … Remove the old-cookie from the cookie store

The result of this is that cookies can be submitted to sites regardless of them having a blank domain set due to the non-standard origin attribute. Without knowledge of this attribute cookies cannot be serialized or copied correctly.

This tiny mojo lite app exhibits the behaviour, the cookie stored in the jar should include the domain name of its source.

```perl
use Mojolicious::Lite;
use Mojo::UserAgent;
use Data::Dump 'pp';

get '/set' => sub {
	my $c = shift;
	$c->cookie(test => 'ing');
	return $c->rendered(200);
};

get '/say' => sub {
	my $c = shift;
	return $c->render(
		text => "Say handler got cookies: " . join ', ', map { pp($_) } $c->req->cookies->@*
	);
};

my $ua   = Mojo::UserAgent->new;
# This request sets a testing cookie without a domain parameter
# according to RFC 6265 Section 5.3 bullet 6 this should be set to the request-host
$ua->get("/set");

# Printing the cookie's parameters show no domain. Serializing or copying this cookie without
[...]

PR #1186, opened 2018-01-31
Improve RFC6265 compliance by replacing 'origin' attribute with 'host_only'

expbbc:
### Summary
This PR improves RFC6265 compliance by correctly setting the `domain` parameter of cookies when they are stored. Currently the domain is left empty and another attribute `origin` is promoted to domain when the cookie is used in a request. This results in incorrectly permitting multiple identically named cookies in addition to not divulging the domain of the cookie once stored.

Specifically, Section 5.3 bullet points 4, 6 & 11 apply: https://tools.ietf.org/html/rfc6265#section-5.3

* 4: If the cookie-attribute-list contains an attribute with an attribute-name of "Domain": … Otherwise: Let the domain-attribute be the empty string.
* 6: If the domain-attribute is non-empty: … Otherwise: Set the cookie's host-only-flag to true. Set the cookie's domain to the canonicalized request-host.
* 11: If the cookie store contains a cookie with the same name, domain, and path as the newly created cookie: … Remove the old-cookie from the cookie store

### Motivation
These changes rectify these two divergences from RFC6265 and increase compliance with servers returning cookies without domain parameters, specifically, the domain parameter is now set when the cookie is stored in the UserAgent's CookieJar which allows these cookies to be inspected and copied elsewhere without having to have knowledge of the 'origin' parameter.

### Deprecation
Other than internal uses which are addressed by this PR, only [one external use](https://metacpan.org/source/EJUNGLE/Mojo-UserAgent-CookieJar-Role-Persistent-0.002/lib/Mojo/UserAgent/CookieJar/Role/Persistent.pm#L43) of the origin attribute was located on CPAN. This use already tests for the `domain` parameter ahead of `origin` and so should be unaffected by this PR.

`Mojo::Cookie::Response::origin` is moved from a 'has' accessor to an explicit method which calls `attr` and prints a deprecation notice in order to inform users.

No other impact is expected, cookie collection will still set the `origin` parameter as the ultimate fallback, but will additionally always set `domain` and `domain` is now the tested parameter when filtering cookies for a particular domain.

### References
Fixes issue #1185


kraih:
Is this pull request going anywhere?

kraih:
I'm closing this pull request for now, please open a new one once you are ready. Normally a work-in-progress pull request is kept open so it can be discussed while refinements are being made. But this pull request has not reached the stage where it can be discussed yet.

PR #1188, opened 2018-02-01
add cURL-style -u for username/password entry to the get command

jberger:
It can be hard to specify modern passwords generated by password managers into the url format given the restrictions on characters. It is much easier to do so only concerning bash escaping rather than worrying about url escaping too. To this end, this PR implements a -u flag (much like) what cURL provides.

### DIFF
--- a/lib/Mojolicious/Command/get.pm
+++ b/lib/Mojolicious/Command/get.pm
@@ -3,8 +3,9 @@ use Mojo::Base 'Mojolicious::Command';
 
 use Mojo::DOM;
 use Mojo::IOLoop;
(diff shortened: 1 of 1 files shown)


kraih:
> Summary
> DESCRIBE THE BIG PICTURE OF YOUR CHANGES HERE
>
> Motivation
> EXPLAIN WHY YOU BELIEVE THESE CHANGES ARE NECESSARY HERE
>
> References
> LIST RELEVANT ISSUES, PULL REQUESTS AND IRC/MAILING-LIST DISCUSSIONS HERE

You convinced me! 😄

PR #1187, opened 2018-02-01
Add section on Reverse Proxy in Cookbook/CONCEPTS

polettix:
### Summary

Add a section about Reverse Proxies, to explain what they are, why you should care about them and how to solve the problem they introduce.

### Motivation

- Information about reverse proxies and how to set them up is a bit scattered around and not always immediately evident in some deployment use cases.
- Looking for `mojolicious reverse proxy` in Google returns solutions that are unnecessary/overkill in the general case (which only involves some configuration).

### References

- [Start of discussion in IRC](https://irclog.perlgeek.de/mojo/2018-01-30#i_15758733)
- [Suggestion to start a PR](https://irclog.perlgeek.de/mojo/2018-02-01#i_15766494)

PR #1189, opened 2018-02-01
Minor wording changes to improve readability

bimmerlabs:


kraih:
I know this is meant to recommend changes to #1187, but i don't think using pull requests like this would work. Best i can do is reference the original pull request in this comment and hope the original author sees it.

PR #1192, opened 2018-02-13
Only one identical cookie should be accepted per domain

expbbc:
### Summary
When a cookie is received by a UA, it should check if there are any existing stored cookies with an identical:
* Domain
* Path
* Name

If any cookies with identical values are found, they should be replaced with the new cookie:
```
11.  If the cookie store contains a cookie with the same name,
        domain, and path as the newly created cookie:
...
        4.  Remove the old-cookie from the cookie store.
```

Mojolicious does *not* set the cookie domain if it is unset, but instead stores an additional `origin` parameter to store the original request-host. RFC 6265 also requires that cookies without a valid domain set are assigned the domain of their request-host:
```
   6.   If the domain-attribute is non-empty:
...
        Otherwise:

           Set the cookie's host-only-flag to true.

           Set the cookie's domain to the canonicalized request-host.
```

 As a result, the original cookie is never removed from the store, and is in fact submitted to the host:

```perl
use Mojolicious::Lite;
use Mojo::UserAgent;
use Data::Dump 'pp';

get '/with' => sub {
	my $c = shift;
	$c->cookie(test => 'WITH DOMAIN', {domain => '127.0.0.1'});
	$c->rendered(200);
};
get '/without' => sub {
	my $c = shift;
	$c->cookie(test => 'WITHOUT DOMAIN');
	$c->rendered(200);
};

get '/say' => sub {
	my $c = shift;
	return $c->render(
		text => "Say handler got cookies: " . join ', ', map { pp($_) } $c->req->cookies->@*
	);
};

my $ua   = Mojo::UserAgent->new;
# This request sets a testing cookie without a domain parameter
# according to RFC 6265 Section 5.3 bullet 6 this should be set to the request-host
$ua->get("/without");
# This request now sets a cookie with a domain parameter. This should override the existing cookie
# and end up the only one in the jar according to RFC 6265 Section 5.3 bullet 11
$ua->get("/with");

# This request now passes BOTH cookies despite them having identical paths and names
say $ua->get("/say")->res->body;
```
```
Say handler got cookies:
    bless({ name => "test", value => "WITHOUT DOMAIN" }, "Mojo::Cookie::Request"),
    bless({ name => "test", value => "WITH DOMAIN" }, "Mojo::Cookie::Request")
```

### Motivation
This change encompasses only the corrections to tests in order to make them representative of RFC 6265. The behaviour of domain-less cookie handling also needs fixing in order to comply with the RFC.

Additionally, it may be considered a security risk that two cookies with identical parameters can be submitted to the same site.

### References
Prior issue: https://github.com/kraih/mojo/issues/1185
PR to fix behaviour in Cookie/UA: https://github.com/kraih/mojo/pull/1186
RFC Section: https://tools.ietf.org/html/rfc6265#section-5.3

sri's request from IRC:
```
<@sri> exp-innit: so you could make a failing test case for cookiejar.t?
<@sri> if an incorrect cookie is sent that would be a serious bug
...
<@sri> no, i want a mojo/cookiejar.t test case that fails but shouldn't
<@sri> the most minimal it can be
```


kraih:
Calling for a vote on this: @jberger, @jhthorsen, @marcusramberg.

kraih:
If these findings can be confirmed to be consistent with Browser behaviour i think this is a (weak) security issue and should be fixed as quickly as possible.

kraih:
I've confirmed that browsers do in fact only send one cookie, so we are leaking old cookies.

PR #1195, opened 2018-02-20
Implement SameSite attribute for sessions cookie.

elcamlost:
### Summary
There is a draft for RFC 6525 
It introduce new cookie attribute - SameSite. It helps to fight against csrf attack.
From the draft
>4.1.2.7.  The SameSite Attribute

>The "SameSite" attribute limits the scope of the cookie such that it will only be attached to requests if those requests are same-site, as defined by the algorithm in Section 5.2.  For example, requests for "https://example.com/sekrit-image" will attach same-site cookies if and only if initiated from a context whose "site for cookies" is "example.com".
>
>If the "SameSite" attribute's value is "Strict", the cookie will only be sent along with "same-site" requests.  If the value is "Lax", the
 cookie will be sent with same-site requests, and with "cross-site" top-level navigations, as described in Section 5.3.7.1.  If the "SameSite" attribute's value is neither of these, the cookie will be ignored.

Modern versions of Chrome and Firefox supports it. 

Maybe Mojolicious can support it too?

### Motivation
It's very simple thing and it adds a bit of security to mojolicious. 

### References
https://www.ietf.org/id/draft-ietf-httpbis-rfc6265bis-02.txt RFC draft
https://caniuse.com/#search=samesite

### DIFF
--- a/lib/Mojo/Cookie/Response.pm
+++ b/lib/Mojo/Cookie/Response.pm
@@ -4,9 +4,9 @@ use Mojo::Base 'Mojo::Cookie';
 use Mojo::Date;
 use Mojo::Util qw(quote split_cookie_header);
 
-has [qw(domain expires host_only httponly max_age path secure)];
+has [qw(domain expires host_only httponly max_age path secure samesite)];
 
-my %ATTRS = map { $_ => 1 } qw(domain expires httponly max-age path secure);
+my %ATTRS = map { $_ => 1 } qw(domain expires httponly max-age path secure samesite);
 
 sub parse {
   my ($self, $str) = @_;
@@ -31,7 +31,6 @@ sub parse {
 
 sub to_string {
   my $self = shift;
-
   # Name and value
   return '' unless length(my $name = $self->name // '');
   my $value = $self->value // '';
@@ -55,6 +54,9 @@ sub to_string {
 
   # "Max-Age"
   if (defined(my $max = $self->max_age)) { $cookie .= "; Max-Age=$max" }
+
+  # "SameSite"
+  if (defined(my $samesite = $self->samesite)) { $cookie .= '; SameSite'.ucfirst($samesite) }
 
   return $cookie;
 }
@@ -115,6 +117,17 @@ the cookie's L</"domain">.
 
 HttpOnly flag, which can prevent client-side scripts from accessing this
 cookie.
+
+=head2 samesite
+
+  my $same_site_policy = $cookie->samesite;
+  $cookie  = $cookie->samesite('strict');
+
+SameSite attribute, currently supported by firefox and chrome.
+If its value is 'strict', browsers send cookie only  to same-site requests.
+If its value is 'lax', browsers send them with cross-site requests too.
+All other values are ignored.
+
 
 =head2 max_age
 
--- a/lib/Mojolicious/Sessions.pm
+++ b/lib/Mojolicious/Sessions.pm
@@ -6,6 +6,7 @@ use Mojo::Util qw(b64_decode b64_encode);
 
 has [qw(cookie_domain secure)];
 has cookie_name        => 'mojolicious';
+has same_site          => 'lax';
 has cookie_path        => '/';
 has default_expiration => 3600;
 has deserialize        => sub { \&Mojo::JSON::j };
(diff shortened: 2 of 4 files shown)


kraih:
The patch does not appear to do anything i'm afraid.

kraih:
You might also want to take a look at the documentation, we have certain expectations for pull requests. http://mojolicious.org/perldoc/Mojolicious/Guides/Contributing#CONTRIBUTING-CODE

elcamlost:
> The patch does not appear to do anything i'm afraid.
Patch
1. [adds same_site attr to Mojolicious::Sessions object, set it to default 'lax' value](https://github.com/kraih/mojo/pull/1195/files#diff-36abbba45830478a3109f6c00b11e0c2R9)
2. [passes value of this attribute to cookie options](https://github.com/kraih/mojo/pull/1195/files#diff-36abbba45830478a3109f6c00b11e0c2R59).

And, the most important, we can change same_site attr value during application startup to 'strict' and browsers (firefox and chrome) will stop send session cookie in cross-site requests.

>You might also want to take a look at the documentation, we have certain expectations for pull requests. http://mojolicious.org/perldoc/Mojolicious/Guides/Contributing#CONTRIBUTING-CODE
I read it, thanks.

jberger:
I think what @kraih is saying is that though your code adds a field to the constructor hash for the cookie, the code that uses that hash to write the cookie doesn't handle it and therefore the net result is no change to the cookie. Further, we have a rule when considering code changes that they must come with documentation (which yours has, thanks) and tests (which yours does not). As a courtesy, we can leave this open for a little while for those to be added. If after a few days these problems are not corrected, as a matter of policy, this PR will be closed.

Please feel free to hop on irc at #mojo on irc.perl.org if you'd like to chat about things.

elcamlost:
Thanks, I finally understood, what Sebastian meant saying "The patch does not appear to do anything i'm afraid.".

I added attribute support to Mojo::Cookie::Response, so value from $options hashref (inside Mojolicious::Sessions) correctly handled by Mojo::Cookie::Response->to_string.

Also I added tests.

kraih:
Is this a work in progress pull request? It does not pass its own tests currently.

kraih:
I'll add a work in progress label for now, that means this pull request is currently not eligible for a vote.

lib/Mojo/Cookie/Response.pm
```diff
@@ -56,6 +55,9 @@ sub to_string {
   # "Max-Age"
   if (defined(my $max = $self->max_age)) { $cookie .= "; Max-Age=$max" }
 
+  # "SameSite"
+  if (defined(my $samesite = $self->samesite)) { $cookie .= '; SameSite'.ucfirst($samesite) }
```
oalders: I believe this should be:

`if (defined(my $samesite = $self->samesite)) { $cookie .= '; SameSite='.ucfirst($samesite) }`

oalders: Although, if you're fixing up the casing, you may want `ucfirst(lc($samesite))`.

kraih:
Thank you, this feature does seem very reasonable. Since there has been no activity on this patch in a few days, i took the liberty of making a new one, so we can test this out in the wild as soon as possible.

kraih:
Turns out this wasn't such a good idea after all, since the feature has already been superseded by a new feature called cookie name prefixes. https://wpdev.uservoice.com/forums/257854-microsoft-edge-developer/suggestions/17140412-support-samesite-cookie-option

kraih:
I'm not sure anymore about the spec situation of `SameSite`, maybe it has been superseded by a new spec. It appears there might be too much conflicting information out there to include the feature for now. It's a security feature after all, things should be clearer.

elcamlost:
http://httpwg.org/http-extensions/rfc6265bis.html
Latest draft includes both features - SameSite attribute and cookie prefixes.

Also see this email from Mike West https://mailarchive.ietf.org/arch/msg/httpbisa/HQBpLTBHNnERWFdhGcfZDujc9sI Message states that both features are considered to be included in RFC and prefixes do not superseed samesite.

>Since there has been no activity on this patch in a few days
Actually I planned to finish it today )) But it doesn't matter. I'm happy it implemented. 
Thanks to you and i'm sorry that my patch was a bit dirty. I've learned my lesson and will be more accurate in the future.

PR #1203, opened 2018-03-18
Match optimization

KES777:
### Summary
Because `{ captures }` are localized they are will be unchanged
when `->_match` returns. So we are not required to restore them

### Motivation
Because [`$captures` is a hash reference to `$self->{ captures }`](https://github.com/kraih/mojo/blob/master/lib/Mojolicious/Routes/Match.pm#L49)

[The code](https://github.com/kraih/mojo/blob/master/lib/Mojolicious/Routes/Match.pm#L90) effectively do nothing, eg. `$hashref =  $hashref`.  
Maybe you mean copy by value `$hashref = { %$hashref }` as this is done for `$self->stack`:

     $self->stack([@{$snapshot[0]}])->{captures} = { %{ $snapshot[1] }};

But `{captures}` are localized so this copy by value is not required

### References
related to this [commit](https://github.com/kraih/mojo/commit/175cc94bb7)

### DIFF
--- a/lib/Mojolicious/Routes/Match.pm
+++ b/lib/Mojolicious/Routes/Match.pm
@@ -84,10 +84,10 @@ sub _match {
   }
 
   # Match children
-  my @snapshot = $r->parent ? ([@{$self->stack}], $captures) : ([], {});
+  my $snapshot = $r->parent ? [@{$self->stack}] : [];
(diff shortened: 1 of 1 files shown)


kraih:
Thanks, the commit message was sadly full of typos, so i've applied the change manually.

PR #1204, opened 2018-03-26
make max optional in num validation check

tim-2:
### Summary
make max value optional in num validation check

### Motivation
`num(1)` looks better than `num(1, 999)` or `num(1, 'inf')`

### DIFF
--- a/lib/Mojolicious/Validator.pm
+++ b/lib/Mojolicious/Validator.pm
@@ -38,7 +38,7 @@ sub _in {
 sub _num {
   my ($validation, $name, $value, $min, $max) = @_;
   return 1 if $value !~ /^[0-9]+$/;
-  return defined $min && $max ? $min > $value || $max < $value : undef;
+  return defined $min && $min > $value || defined $max && $max < $value;
 }
 
 sub _size {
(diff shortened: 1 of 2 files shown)


kraih:
So, if `->num(2, undef)` becomes a feature, shouldn't `->num(undef, 2)` also be one?

marcusramberg:
I'm positive to this feature but would like to see @kraih 's suggestion implemented for consistency.

tim-2:
kraih: you're right. The number check could be done with `Scalar::Util::looks_like_number` to support negative numbers.

Grinnz:
That is not at all what he suggested. And looks_like_number should never be used for validation.

kraih:
@marcusramberg Technically, my suggestion seems to already be implemented, and just needs to be documented and tested.

marcusramberg:
@tim-2 are you planning on updating this PR with tests and documentation for optional min value?

marcusramberg:
I'm 👍 on accepting this PR now that the missing tests are added. Seems both cases are already documented.

kraih:
Thanks, applied. In the future please squash your commits.

PR #1214, opened 2018-04-25
Prohibit path traversal for static files with embedded directory separators 

tomk3003:
### Summary
GET requests with embedded url_escaped backslashes can be used to access local files outside of webroot 

### Motivation
I think Mojo::Path is correct in splitting the URL only on slashes. The problem only occurs when the filesystem is accessed. I don't know whether this is the only location where this happens, but the reported bug can be fixed by checking the path parts for directory separators of the current OS.

### References
 https://irclog.perlgeek.de/mojo/2018-04-23#i_16083074 and following.

### DIFF
--- a/lib/Mojolicious/Static.pm
+++ b/lib/Mojolicious/Static.pm
@@ -9,6 +9,8 @@ use Mojo::Home;
 use Mojo::Loader qw(data_section file_is_binary);
 use Mojo::Util qw(encode md5_sum);
 
+use constant IS_WINDOWS => ($^O eq 'MSWin32' or $^O eq 'cygwin');
+
 # Bundled files
 my $PUBLIC = Mojo::Home->new(Mojo::Home->new->mojo_lib_dir)
   ->child('Mojolicious', 'resources', 'public');
@@ -32,6 +34,9 @@ sub dispatch {
   my $path  = $req->url->path;
   $path = $stash->{path} ? $path->new($stash->{path}) : $path->clone;
   return undef unless my @parts = @{$path->canonicalize->parts};
+
+  # Don't serve when a part has embedded path separators on windows
+  if (IS_WINDOWS) { /\\/ && return undef for @parts }
 
   # Serve static file and prevent path traversal
   return undef if $parts[0] eq '..' || !$self->serve($c, join('/', @parts));
--- a/t/mojolicious/app.t
+++ b/t/mojolicious/app.t
@@ -419,6 +419,11 @@ $t->get_ok('/../../mojolicious/secret.txt')->status_is(404)
 $t->get_ok('/.../mojolicious/secret.txt')->status_is(404)
   ->header_is(Server => 'Mojolicious (Perl)')->content_like(qr/Page not found/);
 
+# Try to access a file which is not under the web root via path
+# traversal (escaped backslashes)
+$t->get_ok('/..%5C..%5Cmojolicious%5Csecret.txt')->status_is(404)
+  ->header_is(Server => 'Mojolicious (Perl)')->content_like(qr/Page not found/);
+
 # Check If-Modified-Since
 $t->get_ok('/hello.txt' => {'If-Modified-Since' => $mtime})->status_is(304)
   ->header_is(Server => 'Mojolicious (Perl)')->content_is('');
--- a/t/mojolicious/production_app.t
+++ b/t/mojolicious/production_app.t
@@ -107,6 +107,11 @@ $t->get_ok('/../../mojolicious/secret.txt')->status_is(404)
 $t->get_ok('/.../mojolicious/secret.txt')->status_is(404)
   ->header_is(Server => 'Mojolicious (Perl)')->content_like(qr/Page not found/);
 
+# Try to access a file which is not under the web root via path
+# traversal in production mode (escaped backslashes)
+$t->get_ok('/..%5C..%5Cmojolicious%5Csecret.txt')->status_is(404)
+  ->header_is(Server => 'Mojolicious (Perl)')->content_like(qr/Page not found/);
+
 # Embedded production static file
 $t->get_ok('/some/static/file.txt')->status_is(200)
   ->header_is(Server => 'Mojolicious (Perl)')
--- a/t/mojolicious/testing_app.t
+++ b/t/mojolicious/testing_app.t
@@ -51,4 +51,10 @@ $t->get_ok('/.../mojolicious/secret.txt')->status_is(404)
   ->header_is(Server => 'Mojolicious (Perl)')
   ->content_like(qr/Testing not found/);
 
+# Try to access a file which is not under the web root via path
+# traversal in testing mode (escaped backslashes)
+$t->get_ok('/..%5C..%5Cmojolicious%5Csecret.txt')->status_is(404)
+  ->header_is(Server => 'Mojolicious (Perl)')
+  ->content_like(qr/Testing not found/);
+
 done_testing();


lib/Mojolicious/Static.pm
```diff
@@ -33,6 +34,9 @@ sub dispatch {
   $path = $stash->{path} ? $path->new($stash->{path}) : $path->clone;
   return undef unless my @parts = @{$path->canonicalize->parts};
 
+  # Don't serve when a part has embedded path separators
+  splitdir(canonpath($_)) > 1 && return undef for @parts;
```
Grinnz: Might be better to do @{[splitdir canonpath $_]} since it's not documented that splitdir will return the number of components in scalar context.

t/mojolicious/app.t
```diff
@@ -419,6 +419,11 @@ $t->get_ok('/../../mojolicious/secret.txt')->status_is(404)
 $t->get_ok('/.../mojolicious/secret.txt')->status_is(404)
   ->header_is(Server => 'Mojolicious (Perl)')->content_like(qr/Page not found/);
 
+# Try to access a file which is not under the web root via path
+# traversal (url_ecaped backslashes)
```
kraih: Typo.

kraih:
Is there a significant performance cost?

tomk3003:
I think you are not interested in the performance on windows?
I tested on my ubuntu notebook.  The performance hit is depending on the URLs path depth.
With a depth of 10, old and new produce about 460 req/s. There is no significant impact.
With a depth of 25 the performance drops to about 450 req/s.
Is that acceptable?

kraih:
But your changes don't just affect Windows, so performance impact does matter quite a bit.

dmanto:
Just to comment that these two mojo tests:
```
use Mojolicious::Lite;
use Test::More;
use Test::Mojo;

get '/' => { data => 'Hello World' };

my $t = Test::Mojo->new;

$t->get_ok('/..\..\..\..\..\..\..\..\..\..\..\windows\win.ini')
  ->content_unlike(qr/extensions/);
$t->get_ok(
    '/..%5C..%5C..%5C..%5C..%5C..%5C..%5C..%5C..%5C..%5C..%5Cwindows%5Cwin.ini')
  ->content_unlike(qr/extensions/);

done_testing();
```
passes ok after aplying the PR

The 'extensions' word happens to be part of the leaked file (c:\windows\win.ini). When tested on strawberry perl 5.26.1 over Win 10 using the master branch it fails, and you can see in the report messages the contents of the leaked file.

So it seems to correct the issue.

kraih:
There is definitely a measurable performance loss.
```perl
perl -Ilib -Mojo -E 'a("/a/b/c/d/e/f/g/h/i/j/k/l/m/n/o/p/q" => {data => "works"})->start' daemon -m production -l http://*:8080
```
`2004.76` Requests/sec before, and `1881.92` Requests/sec after.

tomk3003:
@kraih: exactly, that is the reason I benchmarked on Linux(Ubuntu). I can only benchmark the windows performance hit tomorrow when I will have access to windows again. But since Windows needs the fix, a performance hit would be more acceptable there.

tomk3003:
What do you use to measure the performance against the server?

dmanto:
setting env var MOJO_LOG_LEVEL=debug will show requests/sec.

i.e.:
```
MOJO_LOG_LEVEL=debug perl -Ilib -Mojo -E 'a("/a/b/c/d/e/f/g/h/i/j/k/l/m/n/o/p/q" => {data => "works"})->start' daemon -m production -l http://*:8080
```

kraih:
```
wrk -c 100 -d 10s http://127.0.0.1:8080/a/b/c/d/e/f/g/h/i/j/k/l/m/n/o/p/q
```

t/mojolicious/app.t
```diff
@@ -419,6 +419,11 @@ $t->get_ok('/../../mojolicious/secret.txt')->status_is(404)
 $t->get_ok('/.../mojolicious/secret.txt')->status_is(404)
   ->header_is(Server => 'Mojolicious (Perl)')->content_like(qr/Page not found/);
 
+# Try to access a file which is not under the web root via path
+# traversal (url_escaped backslashes)
```
kraih: `url_escaped` looks weird, it's neither a real word not a function.

lib/Mojolicious/Static.pm
```diff
@@ -34,8 +33,8 @@ sub dispatch {
   $path = $stash->{path} ? $path->new($stash->{path}) : $path->clone;
   return undef unless my @parts = @{$path->canonicalize->parts};
 
-  # Don't serve when a part has embedded path separators
-  @{[splitdir canonpath $_]} > 1 && return undef for @parts;
+  # Don't serve when a part has embedded path separators on windows
+  if ( $^O eq 'MSWin32' or $^O eq 'cygwin' ) { m!\\! && return for @parts };
```
haarg: Should probably use a constant for this.

tomk3003:
Yes, the first commit contained only the changes to the tests and all 3 failed on the appveyor build
https://ci.appveyor.com/project/tomk3003/mojo/build/1 if you want to check. The second fail was because cygwin was not using File::Spec::Win32

Should I use a constant as @haarg suggested. If yes then how? 'my $is_win_os =...' ?

I'll squash the comments when there are no more change requests.

t/mojolicious/app.t
tomk3003: what's better `url escaped` or just `escaped`?

lib/Mojolicious/Static.pm
```diff
@@ -33,6 +33,9 @@ sub dispatch {
   $path = $stash->{path} ? $path->new($stash->{path}) : $path->clone;
   return undef unless my @parts = @{$path->canonicalize->parts};
 
+  # Don't serve when a part has embedded path separators on windows
+  if ( $^O eq 'MSWin32' or $^O eq 'cygwin' ) { m!\\! && return for @parts };
+
```
chy-causer: Just a suggestion: Could you `use constant IS_WINDOWS => ($^O eq 'MSWin32' or $^O eq 'cygwin');` at the top of the file, and then use that 
```
if (IS_WINDOWS) { .... }
```
?
The if-statement should be removed at compile time so you'd get no speed degradation on Linux et al. I haven't tried this mind you.

tomk3003: will do

kraih:
Since @tomk3003 has said this pull request is ready i'm calling for a formal vote (@jberger, @jhthorsen, @marcusramberg). My vote will be neutral, since i very much hate Windows special cases (`$^O eq 'MSWin32'`), but recognise the seriousness of the security issue it is meant to fix.

lib/Mojolicious/Static.pm
```diff
@@ -33,6 +35,9 @@ sub dispatch {
   $path = $stash->{path} ? $path->new($stash->{path}) : $path->clone;
   return undef unless my @parts = @{$path->canonicalize->parts};
 
+  # Don't serve when a part has embedded path separators on windows
+  if (IS_WINDOWS) { m!\\! && return for @parts };
```
kraih: The bare return is wrong, since it's context sensitive. We consider that a security issue.

kraih: Also, why not `/\\/`?

kraih:
Calling off the vote for now since a new security issue has been introduced in the meantime.

jhthorsen:
I don't mind `IS_WINDOWS`, since this is a windows bug.

👍 from me.

Grinnz:
It's not a windows bug but a bug on any system that uses directory separators different from URLs. However since Windows is the only system this fix is meant to address this seems fine to me.

tomk3003:
I just applied @kraih 's suggested regex change I missed over the security issue

yuki-kimoto:
I implement a patch to fix this problem.

https://github.com/yuki-kimoto/mojo/commit/bcfa57676598fdbdeec0478e483aaf7f0a9bbfdd

This patch only fobidden backslash of path when statci file is served.
File path expression must be URL path expression.
But in windows, backslash is allowed in file path expression.
I think this is problem.


>there's a performance cost everyone has to pay for a windows problem
Portability or performance. This is oposite thing.
If one thing is more important, another is limitted.
I want to use plubic method. but for Windows, I want not to slow down Mojolicious.


> alternative implementations?
If backslash is cheked in Mojo::Static::dispatch, all of @parts must be searched. This damages performance much.
I also think using $^O is not good.
In first implementation, before @parts is build, check if path contains backslash.

tomk3003:
I don't like the platform specific workaround either. But it is fast for non Windows platforms. 
The patch suggested by @yuki-kimoto will most likely solve the issue but with a yet to measure performance impact and the additional restriction that backslash will be dissallowed on all OSes (which feels a bit unclean, but will not cause any real problems, I guess). 
If there was a portable way to get the list of allowed filesystem path separators this would be easy. But alas it isn't.

yuki-kimoto:
@tomk3003 @kraih 

>simply disallowing backslash for static file paths on all platforms does seem like a good idea.

>the additional restriction that backslash will be dissallowed on all OSes (which feels a bit unclean, but will not cause any real problems, I guess)

In public directory(This has only static files), developer don't use file names which contains backslash.

In template, backslash use is all of ok in my  implementation proposal.

yuki-kimoto:
> If there was a portable way to get the list of allowed filesystem path separators this would be easy. But alas it isn't.

This seems to be impossible. File::Spec and Config don't provide information of allowed filesystem path separators.

yuki-kimoto:
@kraih

>And i do wonder if it can be optimized to a point where it is not noticeable on benchmarks anymore.

I don't yet do benchmark. Is my implementation bad performance?

dmanto:
> While @yuki-kimoto does not make a particularly strong argument, simply disallowing backslash for static file paths on all platforms does seem like a good idea. And i do wonder if it can be optimized to a point where it is not noticeable on benchmarks anymore.

we could join the @parts array right before comparisons, and then check everything with just one regex, like
```
 ...

  # Serve static file and prevent path traversal
  my $canon_path = join '/', @parts;
  return undef if $canon_path =~ /^..\/|.*\\/ || !$self->serve($c, $canon_path);

...

```

This is probably a bit slower that just the `$parts[0] eq '..', but shouldn't be that much.
 
I forked tomk3003, checkout lib/Mojolicious/Static.pm to original and made these changes, check against cygwin and pure win10 and all tests pass. Branch is 
["static_traverse_check"](https://github.com/dmanto/mojo/commit/62cf3984f1cfa80a51531c095a09efeb64ff1db5)

On an ubuntu linode server (tests also pass) I run master kraijh/mojo version on a daemon on port 8081, and this branch on port 8080, with similar results:

```
➜  mojo git:(28c156b) wrk -c 100 -d 1m http://127.0.0.1:8081/a/b/c/d/e/f/g/h/i/j/k/l/m/n/o/p/q
Running 1m test @ http://127.0.0.1:8081/a/b/c/d/e/f/g/h/i/j/k/l/m/n/o/p/q
  2 threads and 100 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency   106.25ms   12.37ms 196.10ms   84.51%
    Req/Sec   475.98     95.93     0.86k    86.48%
  56498 requests in 1.00m, 7.93MB read
Requests/sec:    940.84
Transfer/sec:    135.22KB
➜  mojo git:(28c156b) wrk -c 100 -d 1m http://127.0.0.1:8080/a/b/c/d/e/f/g/h/i/j/k/l/m/n/o/p/q
Running 1m test @ http://127.0.0.1:8080/a/b/c/d/e/f/g/h/i/j/k/l/m/n/o/p/q
  2 threads and 100 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency   106.68ms   14.90ms 230.27ms   85.99%
    Req/Sec   476.03     95.53   848.00     85.63%
  56200 requests in 1.00m, 7.89MB read
Requests/sec:    935.90
Transfer/sec:    134.51KB
➜  mojo git:(28c156b)

```

After running several times, values varies between 900 and 1100 reqs/sec, and I don't see big differences between both versions

yuki-kimoto:
@dmanto

This seems like to be better than my implementation because this implementation don't use private attribute and merge checking logic into one regular exppresion.

Benchmark result is in 5/1000.

kraih:
Thank you. I've merged #1217 since it does not require Windows to be a special case.

kraih:
I have no idea if that's the right decision, unfortunately we do not have enough reviewers with Windows expertise. That also means Windows will stay officially unsupported.

yuki-kimoto:
> we do not have enough reviewers with Windows expertise.

Is is mean Windows expertise  need Mojolicious core team?

PR #1215, opened 2018-04-26
Patch 1

dmanto:
### Summary
GET requests with backslashes (or url_escaped backslashes) can be used to access files outside of webroot on Windows hosts. 

### Motivation
There is already a proposed solution (PR #1214), but some performance issues were found.
Proposal approach is to consider any path starting with '..' as an intent of path traversal. Thats simple to check because current code is checking for exact match already.

### References
PR #1214, https://irclog.perlgeek.de/mojo/2018-04-23#i_16083074 and following.

### DIFF
--- a/lib/Mojolicious/Static.pm
+++ b/lib/Mojolicious/Static.pm
@@ -34,7 +34,7 @@ sub dispatch {
   return undef unless my @parts = @{$path->canonicalize->parts};
 
   # Serve static file and prevent path traversal
-  return undef if $parts[0] eq '..' || !$self->serve($c, join('/', @parts));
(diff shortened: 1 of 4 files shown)


Grinnz:
What about paths that don't start with `..`?

Grinnz:
There is also the case where a part does not start with `..` but still ends up traversing for example: `images\..\..\..\..\..\..\windows\win.ini`

tomk3003:
Yes, this is the reason why you have to check all elements of `@parts` not only the first one. Even the last one could go up to the root if it contains `'..%FC..%FC..%FC..%FC..'`

kraih:
Thanks for your contribution. We are closing this pull request since it doesn't work.

PR #1216, opened 2018-05-02
Remove ambiguity in use of shift

Grinnz:
### Summary
This causes warnings on 5.10.

### Motivation
Clean up warnings

### References
https://irclog.perlgeek.de/mojo/2018-05-02#i_16119092


Grinnz:
Changed to just reorder the syntax to remove the ambiguity, so that stringification still occurs when trim() does not modify its argument as mentioned by @haarg.

yusrideb:
#1216 after testing, causes warnings on 5.10 has clean up.

PR #1217, opened 2018-05-08
added static files traversal tests and squashed

dmanto:
### Summary
GET requests with embedded backslashes can be used to access local files on Windows hosts

### Motivation
Path separator used in Windows systems is the backslash.

Using backslashes in the path part of a URL can be used on those systems to bypass Mojolicious protection against directory traversal.

To avoid that, this PR blocks any url with backlashes inside the path part of it. 

### References
https://irclog.perlgeek.de/mojo/2018-04-23#i_16083074 and following.


kraih:
Thank you. It is unfortunate that none of our Windows users have stepped up to review this pull request. But it appears to have the most complete tests and least performance impact, therefore i'm merging it.

Issue #1223, opened 2018-05-16
Mojolicious does not check SSL certificates by default


jes:
* Mojolicious version: 7.79
* Perl version: 5.22.1
* Operating system: Ubuntu Xenial

### Steps to reproduce the behavior
mojofetch:

```
#!/usr/bin/perl

use strict;
use warnings;

use Mojo::UserAgent;

my $url = shift or die "usage: mojofetch URL\n";

my $ua = Mojo::UserAgent->new();

$ua->get($url, sub {
    my ($ua, $tx) = @_;
    print $tx->res->body, "\n";
    exit 0;
});

Mojo::IOLoop->start;
```

### Expected behavior

`./mojofetch https://expired.badssl.com/` should not return the content of the page and should trigger some sort of error, because the certificate is expired. Similar applies for self-signed.badssl.com, wrong.host.badssl.com etc., see the list on https://badssl.com/ for more test cases.

### Actual behavior

`./mojofetch https://expired.badssl.com/` does return the content of the page, and doesn't print any errors or warnings.

### More

I looked in `Mojo::IOLoop::TLS` and it appears as though the `_expand` function is responsible for this problem. In the event that something inside Mojolicious doesn't explicitly set `tls_*` options, `_expand` defaults them all to off and no verification. This even overrides `IO::Socket::SSL`'s defaults. Even if you set defaults using `IO::Socket::SSL->set_defaults`, they are overridden to the no-verification behaviour that Mojo defaults to.

This is really serious, but I don't know enough about Mojo to know the most effective way to fix it without breaking other features. Hopefully somebody else can do so.

kraih:
This is something we'd consider if there was a well thought out proposal. The topic keeps coming up every now and then, but nobody ever follows up, so i wouldn't get my hopes up.

jes:
In the absence of any actual fix, I would think a good first step would be a big scary warning in all the relevant documentation.

I for one had *no idea* that Mojolicious treats certificate verification as optional.

jes:
I strongly urge you to reconsider.

This insecure-by-default policy is causing real people to write real software that is vulnerable to MITM attacks, while they have no idea about it because it's not documented, just to make built-in HTTPS testing easier?

I also contend that where the defaults are so obviously broken, a change in the defaults is required.

[...]

PR #1226, opened 2018-05-19
Enable SSL verification by default

jes:
### Summary
Enable SSL verification by default by not setting SSL_verify_mode to 0x00. Add an option to Mojo::UserAgent ("verify") to allow people to explicitly choose non-verifying behaviour. Make Test::Mojo automatically turn off verification so SSL testing still works as before.

I don't expect this is enough to merge straight away, but I think it's a good starting point.

### Motivation
Prevent people from inadvertently writing insecure software.

### References
Issue #1223 "Mojolicious does not check SSL certificates by default".


lib/Mojo/UserAgent.pm
```diff
@@ -16,6 +16,7 @@ use constant DEBUG => $ENV{MOJO_USERAGENT_DEBUG} || 0;
 
 has ca              => sub { $ENV{MOJO_CA_FILE} };
 has cert            => sub { $ENV{MOJO_CERT_FILE} };
+has verify          => sub { $ENV{MOJO_VERIFY} };
```
kraih: I think the option should simply be called `insecure`, like in curl.

lib/Mojo/UserAgent.pm
```diff
@@ -1025,6 +1025,13 @@ L<Mojo::Promise> object instead of accepting a callback.
     warn "Connection error: $err";
   })->wait;
 
+=head2 verify
+
+  my $verify_mode = $ua->verify;
+  $ua->verify(0); # disable TLS certificate verification
+
+Set the TLS verify_mode. See SSL_verify_mode in IO::Socket::SSL.
```
kraih: The example and description are inconsistent with other attribute descriptions.

lib/Test/Mojo.pm
```diff
@@ -17,7 +17,7 @@ use Mojo::Util qw(decode encode);
 use Test::More ();
 
 has [qw(message success tx)];
-has ua => sub { Mojo::UserAgent->new->ioloop(Mojo::IOLoop->singleton) };
+has ua => sub { Mojo::UserAgent->new(verify => 0)->ioloop(Mojo::IOLoop->singleton) };
```
kraih: Looks like you did not perltidy anything.

t/mojo/certs/selfsigned.crt
```diff
@@ -0,0 +1,32 @@
+-----BEGIN CERTIFICATE-----
```
kraih: There should be instructions in `t/mojo/ioloop_tls.t` for how to regenerate all certs. But are you sure the existing certs don't already cover this case?

t/mojo/user_agent_tls.t
```diff
@@ -99,4 +100,22 @@ is $ua->ioloop->stream($tx->connection)->handle->get_cipher, 'AES256-SHA',
...
+$daemon = Mojo::Server::Daemon->new(
+  app    => app,
+  ioloop => Mojo::IOLoop->singleton,
+  silent => 1
+);
+$listen
+  = 'https://127.0.0.1'
+  . '?cert=t/mojo/certs/selfsigned.crt'
+  . '&key=t/mojo/certs/selfsigned.key';
+$port = $daemon->listen([$listen])->start->ports->[0];
```
kraih: Is this really a new test case? Or is it just doing the same as one of the others?

kraih:
You forgot `Mojolicious::Command::get` and `ojo`.

kraih:
How is the CA situation after this change? On which operating systems are system certs used by default? Where do we need to configure and use something like `Mozilla::CA`? How should we document the situation?

lib/Mojo/UserAgent.pm
```diff
@@ -503,8 +504,7 @@ L<Mojo::UserAgent> implements the following attributes.
   $ua    = $ua->ca('/etc/tls/ca.crt');
 
 Path to TLS certificate authority file used to verify the peer certificate,
-defaults to the value of the C<MOJO_CA_FILE> environment variable. Also
-activates hostname verification.
+defaults to the value of the C<MOJO_CA_FILE> environment variable.
```
kraih: There are many more places that need this change.

lib/Mojo/UserAgent.pm
kraih: And be a boolean attribute of course. With `-k/--insecure` flag in `Mojolicious::Command::get`.

kraih: And it appears the actual attribute accessor does not have a test yet.

kraih:
We usually require a deprecation period for breaking changes, and this is going to break some applications. But this change does fall under the security exception in our rules, so we are a little more flexible.

kraih:
Btw. I do welcome competing pull requests for this. If anyone else has more time to finish a polished proposal, please do. Right now this issue has momentum in the community, we shouldn't risk letting it go stale and forgotten.

kraih:
And for the record, after this patch has been applied on openSUSE Tumbleweed and macOS all requests to HTTPS sites with **valid** certificates do no longer work by default.
```
$ mojo get https://mojolicious.org
SSL connect attempt failed error:14090086:SSL routines:ssl3_get_server_certificate:certificate verify failed
 at lib/Mojolicious/Command/get.pm line 76.
```

jes:
Many thanks for fixing this!

I wasn't ignoring your feedback, I was going to get to it next time I worked on this. Glad it's all sorted now :)

PR #1233, opened 2018-06-29
add cookie jar to get command

s1037989:
### Summary
This adds a cookie-jar option to the Mojolicious `get` command, similar to `curl`'s cookie-jar.

### Motivation
This allows being able to call `get` multiple times and follow thru a sequence, for example logging into a website.

### References
None.

### DIFF
--- a/lib/Mojolicious/Command/get.pm
+++ b/lib/Mojolicious/Command/get.pm
@@ -9,6 +9,11 @@ use Mojo::URL;
 use Mojo::UserAgent;
 use Mojo::Util qw(decode encode getopt);
 use Scalar::Util 'weaken';
+use Mojo::Cookie::Request;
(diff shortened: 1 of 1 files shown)


s1037989:
That makes sense, thanks for considering!

CandyAngel:
You could use Mojo::UserAgent::CookieJar::Role::Persistent[1] for loading a Netscape-formatted cookie file on startup for persistence in commands. Saving is explicit though, so you'd either need to either generate it outside of your application (wget/curl) or write a specific command to log in and then save the cookie jar.

[1] I wrote this module and currently use it for several projects

kraih:
I'll close this pull request since it is highly unlikely to pass the vote and can't be accepted for security concerns alone.

Issue #1236, opened 2018-07-11
Mojo::File::spurt: Wide character in syswrite


dk:
* Mojolicious version: 7.87
* Perl version: v5.26.1
* Operating system: ubuntu

### Steps to reproduce the behavior
perl -MMojo::File -le 'Mojo::File->new(q(a.txt))->spurt(qq(\x{100}))'

### Expected behavior
a.txt written with 2 bytes c4 80

### Actual behavior
get error: Wide character in syswrite at /home/dk/perl5/perlbrew/perls/perl-5.26.1/lib/5.26.1/x86_64-linux/IO/Handle.pm line 483.
a.txt is zero bytes

pink-mist:
Mojo::File::spurt is documented to take bytes, but ```qq(\x{100})``` is a character. Encode it to your favourite encoding first.

dk:
>     Write all data at once to the file.

Doesn't seem like bytes to me

pink-mist:
https://metacpan.org/pod/Mojo::File#spurt
```$path = $path->spurt($bytes);```

dk:
Yes, write all bytes from the scalar.

dk:
Wow, just "closed the issue" as a non-issue? That's mature :D

dk:
Well it might be correct, but the documentation "Write all data at once to the file" doesn't say anything about the byte/character distinction, and $bytes can be understood as whatever bytes are there in the scalar, which I would expect. So that's at least a case to either fix the documentation or the attitude.

kraih:
I don't see this discussion end constructively, therefore i'm locking this issue.

PR #1237, opened 2018-07-11
Describe variable naming convention in documentation examples

pink-mist:
### Summary
Spelling out the already established variable naming conventions in example documentation

### Motivation
It seems people aren't reading enough into the variable names used

### References
Most recently there was #1236 but I'm sure if pressed I could find other instances where clarifying this convention would be useful

### DIFF
--- a/lib/Mojolicious/Guides.pod
+++ b/lib/Mojolicious/Guides.pod
@@ -57,17 +57,14 @@ you automatically, whenever they are used.
 =item Variable names
 
 Where possible, the documentation will try to use variable names in example code
-that describes whatever type of data should be passed through to or is returned
-from methods and functions. For instance C<$bytes> or C<$chars> to distinguish
-whether it is encoded bytes or decoded characters, C<$bool> to indicate whether
-a value only indicates true or false, the use of C<$c> to denote a
-L<Mojolicious::Controller> object, or C<$app> to denote the application object.
+that describes the type of data that the API will use. For instance C<$bytes> or
+C<$chars> to distinguish whether it is encoded bytes or decoded characters,
+C<$bool> if the value just indicates true or false, C<$c> to denote a
+L<Mojolicious::Controller> object, or C<$app> to denote the
+L<application|Mojolicious> object.
 
 The distinction between C<$bytes> and C<$chars> is especially important because
-Perl itself only considers them strings, so if an interface is designed to take
-bytes it can get into trouble if given characters, and if an interface is
-designed to take characters it can cause something to be double-encoded if given
-an already encoded string of bytes.
+Perl itself only considers them strings.
 
 =back
 
--- a/lib/Mojolicious/Guides.pod
+++ b/lib/Mojolicious/Guides.pod
@@ -54,6 +54,21 @@ L<features|feature> are enabled, even if examples don't specifically mention it.
 Some modules, like L<Mojo::Base> and L<Mojolicious::Lite>, will enable them for
 you automatically, whenever they are used.
 
+=item Variable names
+
+Where possible, the documentation will try to use variable names in example code
+that describes whatever type of data should be passed through to or is returned
+from methods and functions. For instance C<$bytes> or C<$chars> to distinguish
+whether it is encoded bytes or decoded characters, C<$bool> to indicate whether
+a value only indicates true or false, the use of C<$c> to denote a
+L<Mojolicious::Controller> object, or C<$app> to denote the application object.
+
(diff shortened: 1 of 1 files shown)


Grinnz:
I think it's a good idea. I also think it's worth mentioning that `$bool` is another one with specific meaning.

Grinnz:
Maybe a bit could be added to explain that the distinction between bytes and characters is because Perl makes no distinction in its strings, as many people tend to be confused by this.

s1037989:
I like this.  It took me a little bit of time to recognize and get used to this pattern.  Once I discovered it and learned to use this pattern to my advantage it became an absolute blessing.  Certainly, it would have reduced my learning curve to some degree had this pattern / convention been explicit in the guides as proposed here.

lib/Mojolicious/Guides.pod
```diff
@@ -54,6 +54,21 @@ L<features|feature> are enabled, even if examples don't specifically mention it.
 Some modules, like L<Mojo::Base> and L<Mojolicious::Lite>, will enable them for
 you automatically, whenever they are used.
 
+=item Variable names
+
+Where possible, the documentation will try to use variable names in example code
+that describes whatever type of data should be passed through to or is returned
+from methods and functions. For instance C<$bytes> or C<$chars> to distinguish
+whether it is encoded bytes or decoded characters, C<$bool> to indicate whether
+a value only indicates true or false, the use of C<$c> to denote a
```
jhthorsen: I think that instead of "true" and "false", I think you should do `L<true|Mojo::JSON/true>` or something like that to indicate what kind of "true" and "false" we are talking about.

lib/Mojolicious/Guides.pod
```diff
@@ -54,6 +54,21 @@ L<features|feature> are enabled, even if examples don't specifically mention it.
...
+L<Mojolicious::Controller> object, or C<$app> to denote the application object.
```
jhthorsen: Maybe say "denote the L<application|Mojolicious> object" or "denote the L<Mojolicious> application object"?

jhthorsen:
I also think this is a good idea 👍

lib/Mojolicious/Guides.pod
Grinnz: That's not the kind of true and false we're talking about though. It's just Perl boolean context true and false.

lib/Mojolicious/Guides.pod
pink-mist: not sure what you mean with the
>or "denote the L application object"

lib/Mojolicious/Guides.pod
pink-mist: It was changed from saying "truthiness" to "true or false"; like Grinnz says it's just Perl boolean context, not JSON::PP::Boolean

lib/Mojolicious/Guides.pod
```diff
@@ -54,6 +54,21 @@ L<features|feature> are enabled, even if examples don't specifically mention it.
...
+
+The distinction between C<$bytes> and C<$chars> is especially important because
+Perl itself only considers them strings, so if an interface is designed to take
+bytes it can get into trouble if given characters, and if an interface is
+designed to take characters it can cause something to be double-encoded if given
+an already encoded string of bytes.
```
kraih: This section is getting very long, which might be counter productive. Since many people will just go TLDR and skip it.

kraih:
Afraid i have to give this a 👎 as it stands now. I liked the version i originally saw yesterday, but since then it has grown a lot and is now simply too long for being a small hint for how to better read the docs. Maybe there's a better place for it, but i don't think the conventions section should explain Perl quirks like bytes/chars with this much detail.

lib/Mojolicious/Guides.pod
firnsy: You could replace this section with a shortened form:
> The distinction between C<$bytes> and C<$chars> is especially important because Perl itself only considers them strings.

kiwiroy:
In an attempt to be both brief and clear while leaving the reader opportunity to learn of more detail (Encode seemed more approachable than perlunicode or perlebcdic).

```perldoc
For brevity and clarity, example variables will reflect the type of data the API
uses. For instance, C<$bytes> or C<$chars> to distinguish whether it is encoded
bytes or decoded characters in a Perl L<string|Encode>, C<$bool> if the value
just indicates true or false, C<$c> to denote a L<Mojolicious::Controller>
object, or C<$app> to denote the L<application|Mojolicious> object.
```

kraih:
Closed by #1240.

PR #1246, opened 2018-08-01
 Add max-age support to Mojolicious::Sessions

srchulo:
Proposed fix for https://github.com/kraih/mojo/issues/1244.

### Summary
Adding the equivalent max-age directive alongside any expires directives in cookies.

### Motivation
To avoid issues with mismatching client/server clocks.

### References
https://github.com/kraih/mojo/issues/1244

I can test this later tonight before submitting if this looks okay.

### DIFF
--- a/lib/Mojolicious/Sessions.pm
+++ b/lib/Mojolicious/Sessions.pm
@@ -45,14 +45,18 @@ sub store {
   # Generate "expires" value from "expiration" if necessary
   my $expiration = $session->{expiration} // $self->default_expiration;
   my $default    = delete $session->{expires};
-  $session->{expires} = $default || time + $expiration
+  
+  my $time = time;
+  $session->{expires} = $default || $time + $expiration
     if $expiration || $default;
+  my $max_age = $session->{expires} - $time if $session->{expires};
 
   my $value = b64_encode $self->serialize->($session), '';
   $value =~ y/=/-/;
   my $options = {
     domain   => $self->cookie_domain,
     expires  => $session->{expires},
+    max_age => $max_age,
     httponly => 1,
     path     => $self->cookie_path,
     secure   => $self->secure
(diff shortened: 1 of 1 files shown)


srchulo:
So I've tested this change and it works as expected when expires is set, and when expires is undef.

kraih:
Unfortunately we can't consider this patch until it is finished.

lib/Mojolicious/Sessions.pm
```diff
@@ -45,14 +45,18 @@ sub store {
   # Generate "expires" value from "expiration" if necessary
   my $expiration = $session->{expiration} // $self->default_expiration;
   my $default    = delete $session->{expires};
-  $session->{expires} = $default || time + $expiration
+  
+  my $time = time;
+  $session->{expires} = $default || $time + $expiration
     if $expiration || $default;
+  my $max_age = $session->{expires} - $time if $session->{expires};
```
kraih: Don't do `my $foo = ... if ...`.

srchulo:
Finished in what way?

kraih:
Unfinished in the sense that the code has not been cleaned up and is still untested.

lib/Mojolicious/Sessions.pm
srchulo: Okay. What should I do instead? A traditional if?

my $max_age;
if ($session->{expires}) {
    $max_age = $time if $session->{expires};
}

srchulo:
Got it. I was looking for Sessions test but couldn't find one. Could you point me to the file?

And will the clean up be feedback from you and other owners of the mojo repo?

kraih:
And i should point out that i've given this pull request a `work in progress` tag. As long as that's active this pull request will not be considered (and mostly be left alone by everyone). If what you're looking for is help with your patch, please close the pull request and discuss your branch on IRC or the mailing-list. https://mojolicious.org/perldoc/Mojolicious/Guides/Contributing#CONTRIBUTING-CODE

lib/Mojolicious/Sessions.pm
kraih: Try to find the shortest solution that's still idiomatic Perl.

yusrideb:
83eb843
I have testing that code with look like clean code (maybe is clean code), and worked.
but I'm still hesitant to determine the right test file.

Result in the browser : 
mojolicious-session-max_age

kraih:
Looking around the Alexa Top 500. It appears that very few sites actually use max-age for their session cookies.

srchulo:
I feel like the more valuable question is how many browsers support it, which is basically all modern browsers.

kraih:
I have to strongly disagree. Why so many of the biggest sites consider max-age not valuable enough to use is an important question that needs to be answered before we can consider this change.

srchulo:
Well there are a lot of more modern parts of the web that many sites don't use-- I don't think that determines whether it should be used or not. 

I'm okay if it's decided this isn't right for Mojolicious, I just think the way to look at it is based on its own merits whether it makes sense or not, not what other sites are doing.

yusrideb:
Why so many biggest sites is not consider max-age ?
I assume they use not default session from framework or languages, they created new session mechanism

kraih:
Given the lack of information and uncertain state of the patch. I don't think it makes sense to keep this pull request open as a "work in progress" any longer. Please open a new one should the situation change. And feel free to discuss it in advance on IRC or the mailing-list.

srchulo:
Okay, maybe I will discuss it on the IRC.

Just one question though so I understand, what is the "uncertain state"? Why other sites in the top 500 aren't using it?

kraih:
No, i was referring to the lack of actual improvements made to the patch.

PR #1252, opened 2018-08-17
docfix Mojo::Base->tap allows assigning to $_

wrog:
### Summary
fix description of Mojo::Base->tap to make clear that you can assign to $_ to switch a method chain to different object

### Motivation
it was something that hadn't occurred to me that this was possible until I looked at the code, and the use of the for loop strongly suggests to me that this was deliberately intended behavior ("local $_ = ..." being the more obvious thing to do if you don't care about the aliasing) -- but even if not, it's way useful and worth mentioning.

I also suspect this is an issue that will be coming up more often with Promises given that those typically involve method chains where the object being referenced is constantly changing

### DIFF
--- a/lib/Mojo/Base.pm
+++ b/lib/Mojo/Base.pm
@@ -291,17 +291,22 @@ pass it either a hash or a hash reference with attribute values.
   $object = $object->tap('some_method', @args);
 
 Tap into a method chain to perform operations on an object within the chain
-(also known as a K combinator or Kestrel). The object will be the first argument
-passed to the callback, and is also available as C<$_>. The callback's return
-value will be ignored; instead, the object (the callback's first argument) will
-be the return value. In this way, arbitrary code can be used within (i.e.,
-spliced or tapped into) a chained set of object method calls.
+(also known as a K combinator or Kestrel). The object will be passed to the
+callback as its first argument and will also be available as C<$_>, which,
+in turn, is aliased to the value to be ultimately returned from C<tap> (the
+callback's own return value being ignored), i.e., the object itself, unless
+the callback assigns some other value to C<$_>. In this way, arbitrary code
+can be used within (i.e., spliced or tapped into) a chained set of object
+method calls.
 
   # Longer version
   $object = $object->tap(sub { $_->some_method(@args) });
 
   # Inject side effects into a method chain
   $object->foo('A')->tap(sub { say $_->foo })->foo('B');
+
+  # Perform surgery and continue with a different object
+  $object->tap(sub { ($_, $_->{foo}} = ($_->{foo}, $_->munge()); })->bar;
 
 =head2 with_roles
 


kraih:
No, this was not intended to be a feature. There are no tests for it after all. So, if you wanted to make this a feature you would need to add tests first. But i'll vote 👎 on this proposal because i think it looks rather ugly and is not valuable enough.

Grinnz:
I also disagree that this should be a feature. It will lead only to confusion if you use it.

kiwiroy:
`Mojo::Collection` uses this pattern too for `each`, `first`, `grep`, etc... which can be tempting, though I'm not advocating, to use on large lists to avoid a chain to `map` which creates a new obj.

```
## 9 not 1
perl -MMojo::Collection=c -le 'c(1..5)->each(sub { $_ = 9 if $_ == 1; })->join("\n")->say'
9
2
3
4
5
## only odd numbers get kept and doubled
perl -MMojo::Collection=c -le 'c(1..5)->grep(sub { my $res = $_ % 2; $_ *= 2; $res; })->join("\n")->say'
2
6
10
```

wrog:
I completely agree it's ugly.  But the whole point of facilities like tap and monkey_patch, as I understand it, is for when one is stuck in a situation where the API is missing something and the alternatives are uglier.  Once consensus emerges as to what the new method[s] should look like and get implemented, the various stopgaps can be cleaned up, but until then things need to work...

... and for that situation I would think it's better to err on the side of flexibility [i.e., letting people do what they need to do]

(or I should just use monkey_patch; hm...)

Grinnz:
The point of tap is to apply API functions to the object so that you can continue chaining on the same object afterward, not to make an even more confusing expression. Monkey patch is not really related and whether it's a good idea is highly situational.

Grinnz:
Also consider roles to bundle up a set of changes to an existing class without globally modifying the class (applying a role creates a new class, and you can apply them to an object which creates the class then reblesses the object). https://metacpan.org/pod/Role::Tiny

wrog:
I'll note the particular use case I'm looking at is a promise chain
```
->then(...)
->then(... if($cond) { $p->resolve($value); die; } ... )
->then(...)
->then(...)
->tap(sub{$_ = $p->resolve($_);})
->then(...)
->catch(...)->wait
```
and yes, it would be much cleaner to be able to do
```
->then(...)
->then(...  $p->throw($value) if($cond);  ... )
->then(...)
->then(...)
->settles($p)
->then(...)
->catch(...)->wait
```
but I'm still waffling on the naming issues, 

and until I have an actual coherent proposal that can be adopted, I need duct tape.

Grinnz:
You can do this with `->catch(sub { $_ })` unless I'm missing something?

kraih:
Ok, i don't see this proposal getting the support it needs to pass the vote. Before you send a new pull request for `Mojo::Promise` additions, please discuss the changes on IRC or the mailing-list first.

PR #1270, opened 2018-10-08
Add log shortcut.

marcusramberg:
This PR adds a log shortcut to Mojolicious::Controller,
and updates guides and helpers to use it

Writing $self->app->log over and over feels tedious, this
PR makes Mojolicious more fun to use ;)

As discussed on IRC.

PR #1271, opened 2018-10-08
Add log alias

marcusramberg:
Add a log alias helper to Mojolicious::Plugin::DefaultHelpers,
and update guides and helpers to use it

Logging often happens from the controller, and this saves you from
repetitive $c->app->log calls.

This is a updated version of #1270

### DIFF
--- a/lib/Mojolicious/Controller.pm
+++ b/lib/Mojolicious/Controller.pm
@@ -403,8 +403,8 @@ L<Mojolicious::Controller> implements the following attributes.
 A reference back to the application that dispatched to this controller, usually
 a L<Mojolicious> object.
 
-  # Use application logger
-  $c->app->log->debug('Hello Mojo');
+  # Use application user agent
+  my $tx = $c->app->ua->get('mojolicious.org');
 
   # Generate path
   my $path = $c->app->home->child('templates', 'foo', 'bar.html.ep');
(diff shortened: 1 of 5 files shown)


kraih:
Don't use `$c->log->*` in core Mojo code, it's much much slower than `$c->helpers->log->*`, and can break if someone has a log method in their controller.

marcusramberg:
> Don't use `$c->log->*` in core Mojo code, it's much much slower than `$c->helpers->log->*`, and can break if someone has a log method in their controller.

Good point. Made more sense when it was just a shortcut instead of a helper. Pushed a fixup commit to remove that part now.

kraih:
My vote will be neutral on this. Only reason i can think of against it would be the possibility of us wanting to use a log helper in the future for logging with request ids.

jberger:
I've always wanted the controller's logger to be a little smarter. I'd like it to have things like contextual information, request id and other things. This would be convenient but limit that future option. I'm pretty sure I discussed this at the hack day in Oslo. Anyway, that said, if the convenience is the pure benefit, I wont block it. Guess I'm neutral on the assumption that (and @kraih just posted) that the idea of per-controller logging is at least considered before accepting this proposal.

jhthorsen:
@jberger: I'm not sure if I follow. Could you tell a bit more about what you would add to the log helper to get what you want? An example log line might be enough to get me up to speed.

marcusramberg:
I guess this helper could later be extended to give the logger more
metadata if it supports it?

kraih:
We might want to use nested helpers like `log.debug` for logging with request id.

kraih:
Looks like the discussion went stale. Given the arguments we've collected so far, i'll have to give this a 👎.

kraih:
Since this proposal did not reach the required votes, i'll close it for now.

PR #1273, opened 2018-10-09
avoid "uninitialized value" warning in select_field

tim-2:
### Summary
passing an undefined start value to select_field TagHelper throws an "uninitialized value" warning

> Use of uninitialized value in list assignment at .../site_perl/5.28.0/Mojolicious/Plugin/TagHelpers.pm line 137.

### Motivation
consistency with other TagHelpers which do not produce a warning when passing undefined start values

Test script:
```
use Mojolicious::Lite;

get '/' => sub {
    my $c = shift;

    $c->param(txt => undef);
    $c->param(cb  => undef);
    $c->param(sel => undef);
} => 'index';

app->start;

__DATA__
@@ index.html.ep
%= text_field 'txt'
%= check_box 'cb'
%= select_field 'sel' => [qw(foo bar)]
```

### DIFF
--- a/lib/Mojolicious/Plugin/TagHelpers.pm
+++ b/lib/Mojolicious/Plugin/TagHelpers.pm
@@ -134,7 +134,7 @@ sub _option {
 sub _select_field {
   my ($c, $name, $options, %attrs) = (shift, shift, shift, @_);
 
-  my %values = map { $_ => 1 } @{$c->every_param($name)};
+  my %values = map { $_ => 1 } grep { defined } @{$c->every_param($name)};
 
   my $groups = '';
   for my $group (@$options) {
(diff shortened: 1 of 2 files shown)


kraih:
Can't you modify one of our existing tests to trigger this case easily too?

kraih:
Always squash your commits.

kraih:
You might have to elaborate on what motivated you to make this change. Because nobody on IRC understands this pull request.

tim-2:
I often set start values for input/select fields which come from a database result or another parameter which can be undef.

Example:

Controller:
```
my $names = $db->select('user', 'name')->arrays->flatten;
$self->param(names => $names);
```

where `$names` returns the array
`['name1', undef, 'name2']`

Template:
`%= select_field names => [...], multiple => undef`

PR #1274, opened 2018-10-11
Added WOFF2 MIME type. Added new media types for woff and woff2

kberov:
### Summary
WOFF2 file format is supported by all major browsers. More and more people start using it because it is widely supported format and it is lighter (has better compression). This commit adds support for it and adds new media types for woff and woff2 which are currently W3C recommendations

### Motivation
Instead of developers adding these media types over and over in their applications, they will have them ready for use. Now almost every application uses a font at least for icons. The preferred format is woff2.

### References
https://en.wikipedia.org/wiki/Web_Open_Font_Format
https://www.w3.org/TR/WOFF2/#IMT
https://www.w3.org/TR/2012/REC-WOFF-20121213/


kraih:
Please don't make `Changes` entries yourself.

lib/Mojolicious/Types.pm
```diff
@@ -115,7 +116,8 @@ L<Mojolicious::Types> manages MIME types for L<Mojolicious>.
   svg      -> image/svg+xml
   txt      -> text/plain;charset=UTF-8
   webm     -> video/webm
-  woff     -> application/font-woff
+  woff     -> application/font-woff, font/woff
+  woff2    -> application/font-woff2, font/woff2
```
kraih: The formatting is inconsistent.

kraih:
Why are you adding `application/font-woff2`? Where does that come from?

kraih:
What do browsers actually support now?

lib/Mojolicious/Types.pm
kberov: How it should look like? Please give an example. Thanks.

kberov:
> application/font-woff2
I found that this is the preferred type for Firefox which I use.
Example request made by Firefox 62:
```
GET http://example.com:3001/fonts/bukyvede-regular-webfont.woff2

Host: example.com:3001
User-Agent: Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:62.0) Gecko/20100101 Firefox/62.0
Accept: application/font-woff2;q=1.0,application/font-woff;q=0.9,*/*;q=0.8
Accept-Language: bg,en;q=0.8,en-US;q=0.5,fr;q=0.3
Accept-Encoding: identity
Referer: http://example.com:3001/css/example.css
....
```

And Chromium:
```
GET /fonts/bukyvede-regular-webfont.woff2 HTTP/1.1
Host: local.xn--b1arjbl.xn--90ae:3001
Connection: keep-alive
Pragma: no-cache
Cache-Control: no-cache
Origin: http://example.com:3001
User-Agent: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Ubuntu Chromium/69.0.3497.81 Chrome/69.0.3497.81 Safari/537.36
DNT: 1
Accept: */*
Referer: http://example.com:3001/css/example.css
Accept-Encoding: gzip, deflate
Accept-Language: bg-BG,bg;q=0.9,en-US;q=0.8,en;q=0.7,fr;q=0.6,cs;q=0.5,ru;q=0.4
```
To which Mojolicious replies again with:
`Content-Type: application/font-woff2`

This media type (application/font-woff2) according to Wikipedia is deprecated. And the current recommendation suggests font/woff2, But Firefox uses it as its first preference.

kraih:
>
> This media type (application/font-woff2) according to Wikipedia is
> deprecated. And the current recommendation suggests font/woff2, But Firefox
> uses it as its first preference.
>
Please be more precise, i do not see `application/font-woff2` mentioned
anywhere in the linked Wikipedia page.

kberov:
> This media type (application/font-woff2) according to Wikipedia is deprecated. And the current recommendation suggests font/woff2, But Firefox uses it as its first preference.
> Please be more precise, i do not see "application/font-woff2" mentioned anywhere in the linked Wikipedia page.

Yes, sorry.  "application/font-woff2" is not mentioned. Only "application/font-woff". But "application/font-woff2" is not mentioned as recommendation either.

Anyway. Firefox sends this in its `Accept` header. That is why I added it. Hope this is enough argument.

kraih:
Afraid that does not convince me, so i'll have to vote 👎 on the change as it stands.

kberov:
> What do browsers actually support now?

Firefox and Chrome support all 4 variants for media types that we provide now.
```
woff  => ['application/font-woff',  'font/woff']
woff2 => ['application/font-woff2', 'font/woff2']
```
I mean specifically, that:

1. When Chrome passes a header 'Accept: */*' and we reply with  `Content-Type: application/font-woff2`, the browser uses the font provided by us. 
2. When Chrome passes a header 'Accept: */*' and we reply with `Content-Type: font/woff2`, the browser uses the font provided by us. 
3. When Firefox passes `Accept: application/font-woff2;q=1.0,application/font-woff;q=0.9,*/*;q=0.8` and we reply with `Content-Type: application/font-woff2`, the browser uses the font provided by us. 
4. When Firefox passes `Accept: application/font-woff2;q=1.0,application/font-woff;q=0.9,*/*;q=0.8` and we reply with  `Content-Type: font/woff2`, the browser uses the font provided by us. 

Previously we served the file with `Content-type:text/plain` (the default) and the browsers were falling back to use what they have as font on the system they run and were trying to find characters among the available fonts on the client system.

kraih:
We went with a different solution. https://github.com/mojolicious/mojo/commit/33b39b0c96610eeaa5bb9fc6a767bd922309ae9c

kberov:
@kraih, I would n't say it is much different, because this was my first suggestion on IRC, but @Grinnz suggested [backward](https://freenode.logbot.info/mojo/20181011#c1762036) compatibility which in fact were `application/font-woff2` and `application/font-woff`. It works for me too and as I tested and answered you about the behavior of Firefox and Chrome. I don't know however how Edge behaves.

PR #1278, opened 2018-10-19
Review request - attempt at Mojo::DynamicMethods

shadowcat-mst:
### Summary

I created a thing called Mojo::DynamicMethods and use it to kill AUTOLOAD

### Motivation

sri nerd sniped me by telling me to kill AUTOLOAD

### References

I hate you all

### DIFF
--- a/lib/Mojo/DynamicMethods.pm
+++ b/lib/Mojo/DynamicMethods.pm
@@ -86,7 +86,9 @@ To opt your class into dynamic dispatch, simply pass the -dispatch flag:
 
   use Mojo::DynamicMethods -dispatch;
 
-and implement L</BUILD_DYNAMIC> as a method in your class.
+and implement L</BUILD_DYNAMIC> as a method in your class, making sure that
+the key you use to lookup methods in C<$dyn_methods> is the same thing you
+pass as C<$ref> to register.
 
 =head1 FUNCTIONS
 
@@ -99,32 +101,6 @@ L<Mojo::DynamicMethods> provides the following function, which is not exported:
 Registers the method C<$name> as eligible for dynamic dispatch for C<$class>,
 and sets C<$code> to be looked up for C<$name> by reference C<$ref> in a
 dynamic method contructed by L</BUILD_DYNAMIC>.
-
-=head1 METHODS
-
-L<Mojo::DynamicMethods> requires you to implement the following method on
-any class participating in dynamic dispatch:
-
-=head2 BUILD_DYNAMIC
-  
-  sub BUILD_DYNAMIC {
-    my ($class, $method, $dyn_methods) = @_;
-    return sub {
-      my $self    = shift;
-      my $dynamic = $dyn_methods->{$self}{$method};
-      return $self->$dynamic(@_) if $dynamic;
-      my $package = ref $self;
-      Carp::croak qq{Can't locate object method "$method" via package "$package"};
-    };
-  }
-
-Then later:
-
-  Mojo::DynamicMethods::register $class, $self, $name, $code;
-
-Must return a subroutine suitable for installation under name C<$method>
-which will look up the dynamic method coderef in the supplied C<$dyn_methods>
-hashref using the same reference passed to L</register>.
 
 =head1 SEE ALSO
 
--- a/lib/Mojo/DynamicMethods.pm
+++ b/lib/Mojo/DynamicMethods.pm
@@ -8,7 +8,8 @@ use Scalar::Util 'weaken';
 fieldhash my %Dyn_Methods;
 
 sub import {
-  my ($flag, $caller) = (shift // '', caller);
+  my ($flag, $caller) = ($_[1] // '', caller);
+  return unless $flag eq '-dispatch';
 
   my $dyn_pkg = "${caller}::_DynamicMethods";
   monkey_patch $dyn_pkg, 'can', sub {
--- a/lib/Mojo/DynamicMethods.pm
+++ b/lib/Mojo/DynamicMethods.pm
@@ -4,6 +4,7 @@ use Mojo::Base -strict;
 use Mojo::Util qw(class_to_path monkey_patch);
 use Hash::Util::FieldHash 'fieldhash';
 use Scalar::Util 'weaken';
+no warnings 'once'; # once warnings get confused by the stash lookups
 
 fieldhash my %Dyn_Methods;
 
@@ -34,7 +35,7 @@ sub register {
 
   my $dyn_pkg = "${target}::_DynamicMethods";
   monkey_patch($dyn_pkg, $name, $target->BUILD_DYNAMIC($name, \%Dyn_Methods))
-    unless $dyn_pkg->can($name);
+    unless do { no strict 'refs'; *{"${dyn_pkg}::${name}"}{CODE} };
   $Dyn_Methods{$object}{$name} = $code;
 }
 
--- a/lib/Mojo/DynamicMethods.pm
+++ b/lib/Mojo/DynamicMethods.pm
@@ -1,0 +1,40 @@
+package Mojo::DynamicMethods;
+use Mojo::Base -strict;
+
+use Mojo::Util qw(class_to_path monkey_patch);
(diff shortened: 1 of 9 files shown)


kraih:
String evals and the tricky garbage collection code have me a little worried. The performance gains are amazing, but i don't think it's worth the cost in maintainability. Especially when compared to your other solution, which works almost as well.
```diff
diff --git a/lib/Mojolicious/Controller.pm b/lib/Mojolicious/Controller.pm
index dcd200c74..abee5da5b 100644
--- a/lib/Mojolicious/Controller.pm
+++ b/lib/Mojolicious/Controller.pm
@@ -19,6 +19,9 @@ my %RESERVED = map { $_ => 1 } (
   qw(namespace path status template text variant)
 );

+{package Mojolicious::Controller::_Helpers};
+unshift our @ISA, 'Mojolicious::Controller::_Helpers';
+
 sub AUTOLOAD {
   my $self = shift;

@@ -29,9 +32,26 @@ sub AUTOLOAD {
   # Call helper with current controller
   Carp::croak qq{Can't locate object method "$method" via package "$package"}
     unless my $helper = $self->app->renderer->get_helper($method);
+
+  Mojo::Util::monkey_patch 'Mojolicious::Controller::_Helpers', $method, sub {
+    my $self = shift;
+    Carp::croak
+      qq{Can't locate object method "$method" via package "${\ref $self}"}
+      unless my $helper = $self->{helpers}{$method}
+      ||= $self->app->renderer->get_helper($method);
+    return $self->$helper(@_);
+  };
+
   return $self->$helper(@_);
 }

+sub can {
+  my ($self, $method, @rest) = @_;
+  return undef unless my $can = $self->SUPER::can($method, @rest);
+  my $h = ${Mojolicious::Controller::_Helpers}{$method}{CODE};
+  return $h && $h eq $can ? undef : $can;
+}
+
 sub continue { $_[0]->app->routes->continue($_[0]) }

 sub cookie {
```

kraih:
Actually, i think we can avoid most of the downsides of the other solution by moving the cache to the app instance.
```diff
diff --git a/lib/Mojolicious/Controller.pm b/lib/Mojolicious/Controller.pm
index dcd200c74..149992cb2 100644
--- a/lib/Mojolicious/Controller.pm
+++ b/lib/Mojolicious/Controller.pm
@@ -19,6 +19,9 @@ my %RESERVED = map { $_ => 1 } (
   qw(namespace path status template text variant)
 );

+{package Mojolicious::Controller::_Helpers};
+unshift our @ISA, 'Mojolicious::Controller::_Helpers';
+
 sub AUTOLOAD {
   my $self = shift;

@@ -29,9 +32,27 @@ sub AUTOLOAD {
   # Call helper with current controller
   Carp::croak qq{Can't locate object method "$method" via package "$package"}
     unless my $helper = $self->app->renderer->get_helper($method);
+
+  Mojo::Util::monkey_patch 'Mojolicious::Controller::_Helpers', $method, sub {
+    my $self = shift;
+    my $app  = $self->app;
+    Carp::croak
+      qq{Can't locate object method "$method" via package "${\ref $self}"}
+      unless my $helper = $app->{helpers}{$method}
+      ||= $app->renderer->get_helper($method);
+    return $self->$helper(@_);
+  };
+
   return $self->$helper(@_);
 }

+sub can {
+  my ($self, $method, @rest) = @_;
+  return undef unless my $can = $self->SUPER::can($method, @rest);
+  my $h = ${Mojolicious::Controller::_Helpers}{$method}{CODE};
+  return $h && $h eq $can ? undef : $can;
+}
+
 sub continue { $_[0]->app->routes->continue($_[0]) }

 sub cookie {
```

shadowcat-mst:
If "kill AUTOLOAD entirely" is no longer the goal, my first attempt was definitely simpler, and while "moving the cache to the app" is only going to work for controllers, that's probably the thing that most needs optimising anyway.

Though I'd note the GC thing is fairly idiomatic in stuff that uses weak references this way (it's essentially an unrolling of Hash::Util::FieldHash's idhash stuff that moves the effort somewhere that isn't as hot a path), and I'd be happy to comment it up - and the string eval could be got rid of by basically copy-pasting the croak and having the builder coderef passed to DynamicMethods->import.

So I *could* achieve "less crack, full implementation" - but in terms of *performance* so far as users will notice it, sri's adjusted version of my first attempt, plus perhaps duplicating that approach into Validation since that's also per-request calls, would work out fine.

Which I think leaves my considered opinion as "I still basically like both my patches, each for the purpose it was implemented for, and it's mostly a question of which purpose the Mojo team considers more important". Happy to work on tweaked versions of either if people want.

shadowcat-mst:
Branch updated to strip the string eval

lib/Mojo/DynamicMethods.pm
```diff
@@ -0,0 +1,60 @@
...
+  }
+}
+
+sub register_dynamic {
+  my ($target, $object, $name, $code) = @_;
+  state %dyn_methods;
+  {
+    state %done;
+    unless ($done{$target}{$name}) {
+      our %Builders;
```
kraih: You forgot one `our %Builders`.

shadowcat-mst:
Took sri's mojofication and then moved registration to support direct ->add_helper because users.

shadowcat-mst:
Hokay, I think that's unit tests and POD at least ish done. Comments welcome.

lib/Mojo/DynamicMethods.pm
```diff
@@ -0,0 +1,133 @@
...
+Registers the method C<$name> as eligible for dynamic dispatch for C<$class>,
+and sets C<$code> to be looked up for C<$name> by reference C<$ref> in a
+dynamic method contructed by L</BUILD_DYNAMIC>.
+
+=head1 METHODS
+
+L<Mojo::DynamicMethods> requires you to implement the following method on
+any class participating in dynamic dispatch:
+
+=head2 BUILD_DYNAMIC
```
kraih: We only document methods that the class actually contains like this.

kraih:
`Mojolicious::Validator::new` needs documentation added, like the other [overloaded constructors](https://github.com/mojolicious/mojo/blob/master/lib/Mojo/Log.pm#L241) we have. And i guess the `BUILD_DYNAMIC` methods need to be hidden from `Test::Pod::Coverage`.

kraih:
Please review and vote @jberger, @marcusramberg and @jhthorsen.

shadowcat-mst:
I converted the AUTOLOAD sections to BUILD_DYNAMIC sections. Not sure if that's quite right but it seemed like a good starting point.

shadowcat-mst:
My gods, it's full of green.

Apologies for forgetting how to POD.

marcusramberg:
I think this looks great, and am 👍 to this change.

shadowcat-mst:
@kraih I kept it as capitalised because it's still morally a global but lowercase does look better so I've changed it, thanks.

I definitely think ::_Dynamic looks better than ::_DynamicMethods did in stack traces etc. but I'm about out of opinions at this point

shadowcat-mst:
@kraih asked about possible downsides, including "new leaks".

So, given I'm using the fieldhash trick to make coderefs go out of scope as soon as the object they're eligible to be called on does, I don't *think* I've created any new leaks - the only leak I can think of is where one of those coderefs closes over the thing they're a helper for, which while unfortunate would already be a leak on master AFAICT.

The other thing this code will do is blow the perl5 VM method cache every time it registers a new method name, but in any remotely long lived process that shouldn't make any real difference - if you registered a gensym()ed method name on every request then *maybe* it would be detectable but I really wouldn't expect it to be significant on its own, let alone compared to the performance gains already inherent in this patchset.

I guess a theoretical disadvantage is that if somebody subclassed one of the -dispatch classes and then implemented their own AUTOLOAD, they'd no longer have that called in the case of an already-registered dynamic method name, but that's into jenga style crack stacking so I'm not convinced we should care.

kraih:
The code does look pretty clean now, and i don't see any obvious risks or downsides anymore, so i vote 👍 too.

Issue #1256, opened 2018-09-07
Mojo::IOLoop::Stream memory hog


leonerd:
* Mojolicious version: 7.29
* Perl version: 5.14.2 threaded
* Operating system: Debian Linux x86

### Steps to reproduce the behavior

The `{buffer}` field of a `Mojo::IOLoop::Stream` stores pending bytes for output. Once written, they are `substr()`ed out of the buffer. This means that the memory usage of the actual buffer SV is never actually returned to the malloc pool.

This has the problem that if e.g. a temporary network stall happens, large buffers are created that use up lots of memory, which perl never gets around to reusing even after the buffers are flushed.

### Expected behavior

Once the underlying `_write` method has successfully flushed the contents of the buffer, it should be `undef`ed to return the actual byte buffer back to the malloc pool, so it can be reused.

### Actual behavior

Because currently it doesn't get `undef`ed, any allocated byte storage remains, consuming memory that the rest of the process can't use.

leonerd:
It is important to actually `undef` the buffer - merely assigning `""` into it won't suffice:

```
eval: my $buffer = "A"x1E6; $buffer = ""; Dump $buffer
SV = PV(0x564e1a222bf0) at 0x564e1a24eef8
  REFCNT = 2
  FLAGS = (POK,pPOK)
  PV = 0x7fb180c50010 ""\0
  CUR = 0
  LEN = 1000002
```

Here, the buffer itself is still using a million bytes of RAM (the LEN field)

```
eval: my $buffer = "A"x1E6; undef $buffer; $buffer = ""; Dump $buffer
SV = PV(0x564e1a222bf0) at 0x564e1a24eef8
  REFCNT = 2
  FLAGS = (POK,IsCOW,pPOK)
  PV = 0x564e1a25ebb0 ""\0
  CUR = 0
  LEN = 10
  COW_REFCNT = 1
```

The old buffer has been returned to the malloc pool, and a new one created at only 10 bytes long.

jberger:
So we've added this change to hopefully address the problem. We haven't come to a good way to test it in the test suite, so for a first check @leonerd can you see if this at least is effective?

leonerd:
Yeah, testing this kind of thing is always rather tricky. But the code looks to be doing the right thing on a visual inspection at least.

I can deploy it and see how it behaves, though due to the rare nature of the problem in practice it may be a while before it turns up again.

jberger:
[...]

PR #1279, opened 2018-10-22
Fix 5 warnings on Perl 5.10

preaction:
### Summary
These changes fix 4 warnings in high-traffic code (every use of a Mojo::DOM selector for tags and potentially multiple for every write to a Mojo::IOLoop::Stream), and 1 in less-used code (validations)

For example, on Perl 5.10 this command results in 3 warnings (2 from Mojo::IOLoop::Stream, one from Mojo::DOM::CSS): `perl -Mojo -E'say g("http://google.com")->dom->at("body")'`

### Motivation
Removing these warnings makes it easier to see real issues

### References
3 of the warnings were introduced by the memory freeing done for #1256. This change has been checked using the same tests in that ticket to ensure that the memory is still freed.

### DIFF
--- a/lib/Mojo/DOM/CSS.pm
+++ b/lib/Mojo/DOM/CSS.pm
@@ -120,7 +120,9 @@ sub _compile {
     # Tag
     elsif ($css =~ /\G((?:$ESCAPE_RE\s|\\.|[^,.#:[ >~+])+)/gco) {
       my $alias = (my $name = $1) =~ s/^([^|]*)\|// && $1 ne '*' ? $1 : undef;
-      my $ns = length $alias ? $ns{$alias} // return [['invalid']] : $alias;
+      my $ns = defined $alias && length $alias
+        ? $ns{$alias} // return [['invalid']]
+        : $alias;
       push @$last, ['tag', $name eq '*' ? undef : _name($name), _unescape($ns)];
     }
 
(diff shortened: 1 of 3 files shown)


kraih:
:-1: We do [not optimize](https://mojolicious.org/perldoc/Mojolicious/Guides/FAQ#Which-versions-of-Perl-are-supported-by-Mojolicious) for Perl 5.10. And this pull request does make some of the code more complicated. I will count this as a vote for us to stop supporting Perl 5.10 completely, which is something we've been planning to do for some time now.

kraih:
:-1: We do not specifically optimize for [Perl 5.10](https://mojolicious.org/perldoc/Mojolicious/Guides/FAQ#Which-versions-of-Perl-are-supported-by-Mojolicious). And this pull request makes some code more complicated. I will count this as a vote to stop supporting Perl 5.10 completely, which is something we've been planning for some time now.

kraih:
The core team has discussed this a little, and there is actually a lot of code in Mojo that already depends on Perl 5.14 to work correctly. So using anything older is a very very bad idea and therefore it makes absolutely no sense to silence any warnings that might encourage a user to rethink their choice of Perl version.

preaction:
As an author of a bunch of projects that rely on Mojolicious, I have been chasing support for the same Perl versions as Mojolicious. What version of Perl should I target?

PR #1280, opened 2018-10-24
switch DynamicMethods to a pre-cached super

shadowcat-mst:
Doing this (ala Class::Method::Modifiers) lets me avoid C3 mro trickery to avoid unexpected linearisation errors - means e.g,. people doing

  use Mojo::Base 'Mojolicious::Controller';

twice won't get an error they don't understand. Note that while the example I gave is something that was wrong but accidentally works, it seems to me too easy to get the same helper using class in isa twice and this makes that situation much more predictable


marcusramberg:
Is it possible to write a simple test for the problem this resolves too? In general I'm 👍 tho, seems better to not have to involve next.

PR #1283, opened 2018-10-31
Simplify code

KES777:
### Summary
Do things similar. Like Minion [does](https://metacpan.org/source/SRI/Minion-9.07/lib/Mojolicious/Plugin/Minion/Admin.pm#L14)


kraih:
As discussed on IRC, make the requested improvements.

PR #1284, opened 2018-10-31
Add options to PODRenderer to allow for customisation

jjatria:
### Summary
These changes add options to the PODRenderer plugin to make it possible to adjust the module that the the browser loads by default; the route under which the browser is made available; and the template to use for rendering the HTML.

These make it possible to customise the browser so it fits the style of an existing project, including any possible branding that may be necessary. It also makes it possible to set a landing page that makes sense for that particular project (much like Mojolicious::Guides does for Mojolicious).

### Motivation
I would like to be able to customise the documentation browser so it fits the style of my project.

### References
I suggested this change [on the Mojolicious Telegram channel](https://t.me/perlmojo/12717), and was told having code to look at would make the decision process easier.

### Implementation details
I've tried to make the code fit the style in the rest of the distribution, but I'd be happy to make any changes that are deemed necessary.

I'm not sure if the handling of the `route` option (accepting either a string or a route object) is a desirable feature or if it adds too much complexity for too little benefit, so I'd be OK with making changes to this.

Likewise, moving the `_perldoc` sub into the `any` call was done because I needed access to the config values, but if there is a better way to do this, I'd be happy to change it.

### DIFF
--- a/lib/Mojolicious/Plugin/PODRenderer.pm
+++ b/lib/Mojolicious/Plugin/PODRenderer.pm
@@ -8,11 +8,19 @@ use Mojo::File 'path';
 use Mojo::URL;
 use Pod::Simple::XHTML;
 use Pod::Simple::Search;
+use Scalar::Util 'blessed';
 
 sub register {
   my ($self, $app, $conf) = @_;
 
   my $preprocess = $conf->{preprocess} || 'ep';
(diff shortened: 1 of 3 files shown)


kraih:
I'm afraid the plugin has a very narrow focus regarding features, only things we need in Mojo core will get added. You're welcome to fork the module, as the description says.

Issue #1285, opened 2018-11-03
IPC for subprocesses


kraih:
This comes up every now and then on IRC and i think it would make sense for `Mojo::IOLoop::Subprocess` to have basic support for progress updates. That means a second pipe connecting the two processes, a new method and a new event.
```perl
my $subprocess = Mojo::IOLoop->subprocess(sub {
  my $subprocess = shift;
  sleep 3;
  $subprocess->progress({percentage => 30});
  sleep 3;
  $subprocess->progress({percentage => 60, arbitrary => 'data'});
  sleep 3;
  $subprocess->progress({percentage => 90});
}, sub {
  my ($subprocess, @results) = @_;
  say 'Done!';
});
$subprocess->on(progress => sub {
  my ($subprocess, @data) = @_;
  say $data[0]->{percentage};
});
```
This is mostly rather trivial to implement, but one problem to consider is that progress update data structures might be bigger than the limit for atomic writes. Therefore it might be necessary to a) throw an error if the serialized data structure ends up being bigger, or b) support partial reads and therefore bigger data structures.

akarelas:
What do you think on my take on this? (has test, but no docs)

https://github.com/akarelas/mojo/commit/3be18b8c41fd50698936e176f0eca071f3f31416

PR #1286, opened 2018-11-06
added progress notifications to Mojo::IOLoop::Subprocess

akarelas:
### Summary
Added 'progress' method and event to Mojo::IOLoop::Subprocess, in accordance to kraih's specifications. Includes test, but no documentation just yet.

### Motivation
This pull-request comes after an issue for its implementation was written in the mojo project

### References
issue #1285

### DIFF
--- a/lib/Mojo/IOLoop/Subprocess.pm
+++ b/lib/Mojo/IOLoop/Subprocess.pm
@@ -28,8 +28,8 @@ sub _start {
 
   # Pipe for subprocess communication
   return $self->$parent("Can't create pipe: $!")
-    unless pipe(my $reader, my $writer);
-  $writer->autoflush(1);
+    unless pipe(my $reader, $self->{writer});
+  $self->{writer}->autoflush(1);
 
   # Child
   return $self->$parent("Can't fork: $!")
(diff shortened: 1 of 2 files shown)


lib/Mojo/IOLoop/Subprocess.pm
```diff
@@ -10,6 +10,7 @@ use Storable;
 has deserialize => sub { \&Storable::thaw };
 has ioloop      => sub { Mojo::IOLoop->singleton }, weak => 1;
 has serialize   => sub { \&Storable::freeze };
+has _progress_writer => undef, weak => 1;
```
kraih: We don't use private attribute accessors. Why do you need to weaken the reference?

lib/Mojo/IOLoop/Subprocess.pm
```diff
@@ -55,6 +62,32 @@ sub _start {
...
+      while (1) {
+        my ($length) = $progress_buffer =~ /\A(\d+)\-/;
+        defined $length or last;
+        my $snippet_size = length($length) + 1 + $length;
+        length($progress_buffer) >= $snippet_size or last;
+        my $snippet = substr $progress_buffer, 0, $snippet_size, '';
+        $snippet =~ s/\A$length\-//;
+        my $args = $self->deserialize->($snippet);
+        $self->emit(progress => @$args);
+      }
```
kraih: This block is in need of some golfing, it seems way too verbose. 😉

lib/Mojo/IOLoop/Subprocess.pm
akarelas: I weakened the reference thinking that would allow the pipe to disappear once `->run` completes (but before `$subprocess` goes out of scope, and thus save on resources. Is there something wrong with that thinking?

lib/Mojo/IOLoop/Subprocess.pm
kiwiroy: `pos` can replace the `length($length) + 1`
```perl
        my ($length, $s) = (($progress_buffer =~ m/\A(\d+)\-/gc), pos($progress_buffer));
        defined $length or last;
        length($prog_buffer) >= ($length += $s) or last;
        my $args = $self->deserialize->(substr substr($progress_buffer, 0, $length, ''), $s);
        $self->emit(progress => @$args);
```

kraih:
Squash your commits.

lib/Mojo/IOLoop/Subprocess.pm
```diff
@@ -26,23 +26,30 @@ sub _start {
...
+  # Pipe for subprocess communication about progress
+  return $self->$parent("Can't create progress pipe: $!")
+    unless pipe(my $progress_reader, $self->{progress_writer});
+  $self->{progress_writer}->autoflush(1);
+
   # Child
   return $self->$parent("Can't fork: $!")
     unless defined(my $pid = $self->{pid} = fork);
   unless ($pid) {
+    close $progress_reader;
```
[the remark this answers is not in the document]
akarelas: I don't know the answer to that, to be honest. I just followed the example in perldoc about pipes.

marcusramberg:
lgtm 👌

PR #1298, opened 2018-12-06
Add support for the SameSite cookie attribute

dylanwh:
### Summary
This PR adds support for the SameSite cookie attribute to Mojo::Cookie::Response and Mojolicious::Sessions.

### Motivation
SameSite prevents the browser from sending cookies along with cross-site requests. The main goal is mitigate the risk of cross-origin information leakage. It also provides some protection against cross-site request forgery attacks. 

## Links

https://tools.ietf.org/html/draft-west-first-party-cookies-07

### DIFF
--- a/lib/Mojo/Cookie/Response.pm
+++ b/lib/Mojo/Cookie/Response.pm
@@ -129,7 +129,13 @@ Max age for cookie.
 =head2 samesite
 
   my $samesite = $cookie->samesite;
-  $cookie       = $cookie->samesite('Lax');
+  $cookie      = $cookie->samesite('Lax');
+  # or
+  $cookie      = $cookie->samesite('Strict');
+
+SameSite flag, prevents the browser from sending cookies along with cross-site requests.
+The B<Strict> option blocks all cross-site cookies, even for top-level navigation (such as clicking on links).
+B<Lax> will only block cross-site cookies for cross-site resource requires, such as images.
 
 =head2 path
 
--- a/lib/Mojo/Cookie/Response.pm
+++ b/lib/Mojo/Cookie/Response.pm
@@ -4,9 +4,9 @@ use Mojo::Base 'Mojo::Cookie';
 use Mojo::Date;
 use Mojo::Util qw(quote split_cookie_header);
 
-has [qw(domain expires host_only httponly max_age path secure)];
+has [qw(domain expires host_only httponly max_age path secure samesite)];
 
-my %ATTRS = map { $_ => 1 } qw(domain expires httponly max-age path secure);
+my %ATTRS = map { $_ => 1 } qw(domain expires httponly max-age path secure samesite);
 
 sub parse {
   my ($self, $str) = @_;
@@ -55,6 +55,9 @@ sub to_string {
 
   # "Max-Age"
   if (defined(my $max = $self->max_age)) { $cookie .= "; Max-Age=$max" }
+
+  # "Same-Site"
+  if (defined(my $samesite = $self->samesite)) { $cookie .= "; SameSite=$samesite" }
 
   return $cookie;
 }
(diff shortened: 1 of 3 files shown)


lib/Mojo/Cookie/Response.pm
```diff
@@ -4,9 +4,9 @@ use Mojo::Base 'Mojo::Cookie';
 use Mojo::Date;
 use Mojo::Util qw(quote split_cookie_header);
 
-has [qw(domain expires host_only httponly max_age path secure)];
+has [qw(domain expires host_only httponly max_age path secure samesite)];
```
kraih: Order not alphabetically.

lib/Mojo/Cookie/Response.pm
```diff
@@ -4,9 +4,9 @@ use Mojo::Base 'Mojo::Cookie';
...
 
-my %ATTRS = map { $_ => 1 } qw(domain expires httponly max-age path secure);
+my %ATTRS = map { $_ => 1 } qw(domain expires httponly max-age path secure samesite);
```
kraih: Order again.

lib/Mojo/Cookie/Response.pm
```diff
@@ -123,6 +126,17 @@ cookie.
...
+=head2 samesite
+
+  my $samesite = $cookie->samesite;
+  $cookie      = $cookie->samesite('Lax');
+  # or
+  $cookie      = $cookie->samesite('Strict');
+
+SameSite flag, prevents the browser from sending cookies along with cross-site requests.
+The B<Strict> option blocks all cross-site cookies, even for top-level navigation (such as clicking on links).
+B<Lax> will only block cross-site cookies for cross-site resource requires, such as images.
```
kraih: Lines seem too long.

lib/Mojo/Cookie/Response.pm
```diff
@@ -123,6 +126,17 @@ cookie.
 
 Max age for cookie.
 
+=head2 samesite
+
+  my $samesite = $cookie->samesite;
+  $cookie      = $cookie->samesite('Lax');
+  # or
```
kraih: This is inconsistent with the rest of the docs. We just show calling conventions at the top, then a description, and then more detailed examples with descriptive comments.

t/mojo/cookie.t
```diff
@@ -443,4 +443,8 @@ like $@, qr/Method "parse" not implemented by subclass/, 'right error';
 eval { Mojo::Cookie->to_string };
 like $@, qr/Method "to_string" not implemented by subclass/, 'right error';
 
+$cookies = Mojo::Cookie::Response->parse(
+  'foo=bar; Path=/; Expires=Tuesday, 09-Nov-99 23:12:40 GMT; SameSite=Lax');
+is $cookies->[0]->samesite, 'Lax', 'right samesite';
```
kraih: A `to_string` test is missing.

lib/Mojolicious/Sessions.pm
```diff
@@ -4,7 +4,7 @@ use Mojo::Base -base;
 use Mojo::JSON;
 use Mojo::Util qw(b64_decode b64_encode);
 
-has [qw(cookie_domain secure)];
+has [qw(cookie_domain secure samesite)];
```
kraih: Attribute has no documentation or tests. Should it be activated by default too?

t/mojo/cookie.t
kraih: And i suppose these tests should be placed a little higher up, maybe next to a related topic like `httponly`.

lib/Mojolicious/Sessions.pm
dylanwh: > Attribute has no documentation or tests. Should it be activated by default too?
This is a tricky question. It would provide the best default security to default to Lax for the session cookie, but it also will potentially break working applications.

I will add documentation for this but it's definitely a maintainer-question on what the default should be.

heikojansen:
FWIW, there was also #1057 which might be used as inspiration, may it be negative or positive ;-)

kraih:
Looks like this pull request went stale after all.

dylanwh:
I'll reopen once I fix it up. We had a few fires this week, sorry. :(

PR #1300, opened 2018-12-13
adjust Mojo::Pointer to better handle invalid pointers

### LINKED ISSUE #1296: Mojo::JSON::Pointer should return undef for all invalid paths
* Mojolicious version: 8.04

By my reading of https://tools.ietf.org/html/rfc6901, json pointers should always start with `/`. If they don't, the pointer is invalid, and data should not be returned.  However, Mojo::JSON::Pointer returns the entire document when provided a path that lacks a leading `/`, which is the same as if `''` had been passed, and causes operator confusion.

This patch demonstrates the issue; the first test fails (the entire document is returned), but it should behave the same as the second (passing) test, returning undef when an unresolvable pointer is passed:

```
diff --git a/t/mojo/json_pointer.t b/t/mojo/json_pointer.t
index 0a34579c5..503eea45a 100644
--- a/t/mojo/json_pointer.t
+++ b/t/mojo/json_pointer.t
@@ -90,6 +90,8 @@ my $hash = {
 };
 $pointer = Mojo::JSON::Pointer->new($hash);
 is_deeply $pointer->get(''), $hash, 'empty pointer is whole document';
+is_deeply $pointer->get('foo'), undef, 'pointer without leading slash is undef';
+is_deeply $pointer->get('/bar'), undef, 'pointer to non-existent path is undef';
 is_deeply $pointer->get('/foo'), ['bar', 'baz'], '"/foo" is "["bar", "baz"]"';
 is $pointer->get('/foo/0'), 'bar', '"/foo/0" is "bar"';
 is $pointer->get('/'),      0,     '"/" is 0';
```

```
ok 39 - empty pointer is whole document
not ok 40 - pointer without leading slash is undef
#   Failed test 'pointer without leading slash is undef'
#   at t/mojo/json_pointer.t line 93.
#     Structures begin differing at:
#          $got = HASH(0x7f81cba40670)
#     $expected = undef
ok 41 - pointer to non-existent path is undef
```

karenetheridge:
While the pointers '' and '/' refer to the entire document, any other
pointer without a leading slash is invalid, and should return undef
rather than the document.

closes #1296.

### DIFF
--- a/lib/Mojo/JSON/Pointer.pm
+++ b/lib/Mojo/JSON/Pointer.pm
@@ -12,7 +12,8 @@ sub _pointer {
   my ($self, $contains, $pointer) = @_;
 
   my $data = $self->data;
-  return $contains ? 1 : $data unless $pointer =~ s!^/!!;
+  return $contains ? 1 : $data if $pointer eq '';
+  return $contains ? 0 : undef unless $pointer =~ s!^/!!;
   for my $p (length $pointer ? (split '/', $pointer, -1) : ($pointer)) {
     $p =~ s!~1!/!g;
     $p =~ s/~0/~/g;
--- a/t/mojo/json_pointer.t
+++ b/t/mojo/json_pointer.t
@@ -15,6 +15,8 @@ ok $pointer->new({foo => {bar => undef}})->contains('/foo/bar'),
 # "contains" (string)
 $pointer = Mojo::JSON::Pointer->new('works');
 ok $pointer->contains(''), 'contains ""';
+ok !$pointer->contains('foo'), 'pointer without leading slash -> false';
+ok !$pointer->contains('/bar'), 'pointer to non-existent path -> false';
 ok !$pointer->contains('/'),    'does not contain "/"';
 ok !$pointer->contains('/foo'), 'does not contain "/foo"';
 ok $pointer->new('0')->contains(''), 'contains ""';
@@ -90,6 +92,8 @@ my $hash = {
 };
 $pointer = Mojo::JSON::Pointer->new($hash);
 is_deeply $pointer->get(''), $hash, 'empty pointer is whole document';
+is_deeply $pointer->get('foo'), undef, 'pointer without leading slash is undef';
+is_deeply $pointer->get('/bar'), undef, 'pointer to non-existent path is undef';
 is_deeply $pointer->get('/foo'), ['bar', 'baz'], '"/foo" is "["bar", "baz"]"';
 is $pointer->get('/foo/0'), 'bar', '"/foo/0" is "bar"';
 is $pointer->get('/'),      0,     '"/" is 0';


kraih:
If you want this patch to be considered please provide a good reason (as i said on IRC before). This is a new feature, not a bug fix, your pull request description is a bit misleading.

karenetheridge:
I find it very confusing that one gets a different result for a bad pointer depending on whether it has a leading slash or not.  In the tests, neither `'foo'` and `'/bar'` correspond to a location in the document, but they return different results.

kraih:
The RFC is much more vague than you make it sound.
>  This specification does not define how errors are handled.  An
>   application of JSON Pointer SHOULD specify the impact and handling of
>  each type of error.
>
>  For example, some applications might stop pointer processing upon an
> error, while others may attempt to recover from missing values by
>  inserting default ones.

karenetheridge:
To be more clear on my specific usecase: I think it is wrong that Test::Mojo's `->json_is('foo')` would return a "pass" when it is obviously a bad pointer. This has led to incorrectly-passing tests in my code.

Would it be preferred to modify Test::Mojo instead, so that runtime uses of json pointers are not affected?

jberger:
I keep thinking about this and while I know we don't typically validate inputs, this feels more like silently doing the wrong thing than just not validating. While the RFC is maddeningly vague, it does clearly state that a pointer that doesn't start with '/' (or '') is an error. I could support returning undef (as in not matching), throwing an error (though I don't think this is consistent, so probably not) or treating it as if there was an implicit leading slash. Returning the document on error doesn't seem appropriate.

karenetheridge:
These paragraphs:

> This specification does not define how errors are handled.  **An application of JSON Pointer SHOULD specify the impact and handling of each type of error.**

> For example, some applications might stop pointer processing upon an error, while others may attempt to recover from missing values by inserting default ones.

...do leave it open to do whatever we want on an error condition, including defaulting to returning the whole document. However, that behaviour should be documented, so if it's preferred, I can prepare a doc patch that explains this. I'd like to understand why that behaviour is preferred though, other than speed (there is one string comparison to `''` between the two implementations).

kraih:
The Mojolicious project defaults to choosing the simplest solution unless there are good reasons not to (such as a contradicting spec).

jhthorsen:
I don't understand how you'll get an invalid pointer unless you have a typo. Such a typo should be easy to fix and I not sure if it should be Mojo's responsibility.

When that is said, I think the diff is so small and the confusion might be huge, so I'm leading toward being positive.

How about his diff:

```
return length $pointer ? undef : $contains ? 1 : $data unless $pointer =~ s!^/!!;
```

kraih:
So at the very least the implementation could be more elegant and efficient. So i think it's best we stop the vote here and reject the pull request for now.

PR #1301, opened 2018-12-14
Add support for SameSite cookies

dylanwh:
### Summary
This PR adds support for the SameSite cookie attribute to Mojo::Cookie::Response and Mojolicious::Sessions.

### Motivation
SameSite prevents the browser from sending cookies along with cross-site requests. The main goal is mitigate the risk of cross-origin information leakage. It also provides some protection against cross-site request forgery attacks. 

This security implications of this are comparable to the httponly flag, and as such Mojolicious::Sessions sets SameSite=Lax by default. I do not understand the use case for SameSite=Strict (it seems to break normal browsing behavior)

### References

* https://tools.ietf.org/html/draft-west-first-party-cookies-07
* #1298 
* #1057

### DIFF
--- a/lib/Mojolicious/Sessions.pm
+++ b/lib/Mojolicious/Sessions.pm
@@ -152,7 +152,6 @@ SameSite attribute for session cookies.
 This attribute is used to control when the browser is allowed to send cookies
 for cross-site requests.
 
-
 =over 4
 
 =item C<"Lax">
--- a/lib/Mojolicious/Sessions.pm
+++ b/lib/Mojolicious/Sessions.pm
@@ -9,6 +9,7 @@ has cookie_name        => 'mojolicious';
 has cookie_path        => '/';
 has default_expiration => 3600;
 has deserialize        => sub { \&Mojo::JSON::j };
+has samesite           => 'Lax';
 has serialize          => sub { \&Mojo::JSON::encode_json };
 
 sub load {
@@ -55,7 +56,7 @@ sub store {
     expires  => $session->{expires},
     httponly => 1,
     path     => $self->cookie_path,
-    samesite => 'Lax',
+    samesite => $self->samesite,
     secure   => $self->secure,
   };
   $c->signed_cookie($self->cookie_name, $value, $options);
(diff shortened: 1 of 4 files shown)


lib/Mojolicious/Sessions.pm
```diff
@@ -55,7 +55,8 @@ sub store {
     expires  => $session->{expires},
     httponly => 1,
     path     => $self->cookie_path,
-    secure   => $self->secure
+    samesite => 'Lax',
```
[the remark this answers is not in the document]
dylanwh: I didn’t look for the tests yet. I will make it configurable and add tests.

kraih:
Did this pull request go stale again?

kraih:
If you want this to be considered a work in progress pull request please set a deadline. https://github.com/mojolicious/mojo/commit/5cfd5e6b5226646eeb9eb73ec7e15d91f6d1898e

dylanwh:
In lieu of some tagging, I'll have this done by before 2018-12-18 UTC.

kraih:
You're supposed to add the deadline to the pull request description. So comment is perfectly fine too.

lib/Mojolicious/Sessions.pm
```diff
@@ -140,6 +142,38 @@ A callback used to deserialize sessions, defaults to L<Mojo::JSON/"j">.
...
+
+Cookies will never be sent for cross-site requests. Even when the user follows a
+link to the application, the session cookies will not be sent.
+
+=item C<undef>
+
+Cookies will be sent for all cross-site requests. This is the traditional
+behavior of cookies.
+
+=back
```
kraih: We normally link to the specs in cases like this, instead of trying to write our own explanations.

dylanwh: Ok. I was wondering if this was the case. Though it's N-level deep in the spec to understand how this impacts experience.

kraih: `Mojo::Headers` for example has a lot of these. https://mojolicious.org/perldoc/Mojo/Headers#origin

dylanwh:
I guess https://tools.ietf.org/html/draft-west-first-party-cookies-07 is close to canonical... even if it has expired, chrome and firefox have no intention of reverting SameSite :-)

kraih:
Unfortunately this pull request expired.

PR #1304, opened 2018-12-21
Made debug, exception and not_found pages responsive

jhthorsen:
### Summary
This PR makes the debug, exception and not_found pages mobile friendly. See screenshots below for a quick overview of the changes.

* Menu links is hidden under a hamburger toggle on small screens.
* The debug pages on mobile will allow the pre/table elements to scroll horizontally.
* The images on not_found and exception pages are in center on mobile.
* Should lib/Mojolicious/resources/templates/mojo/perldoc.html.ep be deleted?

### Motivation
The motivation for this PR is to make Mojolicious more mobile friendly. It will also make the diff for  https://github.com/mojolicious/mojolicious.org/pull/6 smaller.

### References
* https://github.com/mojolicious/mojolicious.org/pull/6

not found production
exception production
exception development
not found menu open development

### DIFF
--- a/lib/Mojolicious/resources/templates/mojo/menubar.html.ep
+++ b/lib/Mojolicious/resources/templates/mojo/menubar.html.ep
@@ -67,21 +67,21 @@
       #mojobar-brand picture:after {
         content: "\2261";
         color: #f0f9fc;
-        font-size: 46px;
-        line-height: 16px;
+        font-size: 36px;
         text-align: center;
         width: 28px;
-        height: 28px;
+        height: 46px;
+        line-height: 46px;
+        vertical-align: middle;
         position: absolute;
-        top: 12px;
+        top: 0;
         right: 1rem;
         transition: transform 0.26s ease;
         display: block;
       }
       #mojobar.open #mojobar-brand picture:after {
         content: "\2715";
-        font-size: 26px;
-        line-height: 32px;
+        font-size: 30px;
         transform: rotate(180deg);
       }
       #mojobar-links {
@@ -150,10 +150,11 @@
       $('html, body').animate({scrollTop: offset.top - mojobarHeight}, 1);
     }
   }
-  document.querySelector("#mojobar-brand").addEventListener("click", function (e) {
-    if (window.innerWidth > 900) return;
-    e.preventDefault();
-    document.querySelector("#mojobar").classList.toggle("open");
+  $("#mojobar-brand").on("click", function (e) {
+    if (window.innerWidth <= 900) {
+      e.preventDefault();
+      $("#mojobar").toggleClass("open");
+    }
   });
   $(window).on('load', function () {
     if (window.location.hash) {
--- a/lib/Mojolicious/resources/templates/mojo/menubar.html.ep
+++ b/lib/Mojolicious/resources/templates/mojo/menubar.html.ep
@@ -50,6 +50,71 @@
     }
     #mojobar-links form { display: inline }
     .animated { transition: all 0.25s ease }
+    @media (max-width: 900px) {
+      #mojobar.open {
+        overflow: visible;
+        box-shadow: none;
+        transform: translateY(0) !important; /* prevent for being hidden on scroll */
+      }
+      #mojobar-brand {
+        line-height: 46px;
+        padding: 0 1rem;
+        display: block;
+        overflow: visible;
+      }
+      #mojobar-brand img {
+        vertical-align: middle;
+      }
+      #mojobar-brand picture:after {
+        content: "\2261";
+        color: #f0f9fc;
+        font-size: 46px;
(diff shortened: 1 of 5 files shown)


lib/Mojolicious/resources/templates/mojo/menubar.html.ep
```diff
@@ -86,6 +150,11 @@
       $('html, body').animate({scrollTop: offset.top - mojobarHeight}, 1);
     }
   }
+  document.querySelector("#mojobar-brand").addEventListener("click", function (e) {
+    if (window.innerWidth > 900) return;
```
kraih: I don't like the if without block, is that really considered good JavaScript style?

lib/Mojolicious/resources/templates/mojo/menubar.html.ep
```diff
@@ -86,6 +150,11 @@
       $('html, body').animate({scrollTop: offset.top - mojobarHeight}, 1);
     }
   }
+  document.querySelector("#mojobar-brand").addEventListener("click", function (e) {
```
kraih: The rest of the file uses jQuery for everything, shouldn't it be consistent?

lib/Mojolicious/resources/templates/mojo/menubar.html.ep
```diff
@@ -50,6 +50,70 @@
...
+        line-height: 46px;
+        padding: 0 1rem;
+        display: block;
+        overflow: visible;
+      }
+      #mojobar-brand img {
+        vertical-align: middle;
+      }
+      #mojobar-brand picture:after {
+        content: "\2261";
```
kraih: How portable is this?

lib/Mojolicious/resources/templates/mojo/menubar.html.ep
jhthorsen: I'm not sure which JavaScript style we're going for, but I don't mind this style, in the same way as I like `return unless $whatever` in perl.

Nevertheless, I'll change it in a new commit.

lib/Mojolicious/resources/templates/mojo/menubar.html.ep
jhthorsen: I agree it should be consistent. If you want, I'll make a new PR later on where we avoid loading jQuery, but for now I'll update this to use jQuery.

lib/Mojolicious/resources/templates/mojo/menubar.html.ep
jhthorsen: I've tested it through browserstack.com on the following devices/browsers:

* Iphone XS
* Iphone 8
* Google Pixel 2
* Galaxy S6
* Windows Phone 8.1
* Internet Explorer 9
* Firefox 60
* Google Chrome 60

It does not work in IE8, but I don't consider that a problem, since the target is small screen devices and not many users are on IE8 - Especially not developers.

The appearance is also a bit different, so if we want it to be pixel perfect, then I have to replace it with an image/icon.

jhthorsen: What is the target for these internal pages? I mean, do we care about IE9 and older? I was hoping we could drop support for anything that is used less than 0.25%.

There's a nice slider on the settings page on https://caniuse.com/ that can give a quick guide of what needs to be supported.

Browser usage 0.25%

jhthorsen:
What is a "Bootstrap solution"? Does that mean depending on all of the CSS from bootstrap or are you talking specifically about the breakpoints they use?

There’s two ways to set breakpoints: One way is to do it based on the device, which is the default in Bootstrap. Those values however are sort of random values, since there’s so many screen widths nowadays - especially if you take rotation of screens into consideration. The other way to set a breakpoint is to look at the content you have and see when it has to change layout. That’s what I choose here, but I also rounded it up to a value that is easier to remember.

kraih:
One of our spin-off projects [already uses](https://github.com/mojolicious/minion#minion-) Bootstrap very successfully to provide a responsive UI. (And [another plugin](https://github.com/mojolicious/mojo-status#mojolicious-plugin-status-) in the mojolicious org does too...) I'm afraid your arguments have not convinced me that your solution is a better choice, it just doesn't feel right.

jhthorsen:
Ok. Thanks for looking over the changes.

lib/Mojolicious/resources/templates/mojo/menubar.html.ep
jhthorsen: I’m reading the styling guide for Airbnb which has 80k stars, and they don’t mind skipping the block:

https://github.com/airbnb/javascript/blob/master/README.md#blocks--braces

But they also like cuddled elses, so I’m not sure if they can be trusted 😂

PR #1305, opened 2018-12-25
Mojo::IOLoop::Client: no NDN until next_tick

iamb:
as is done with blocking dns, delay creation of the NDN object (and
importantly, its sockets) until the loop runs.

adds a test to check that no handles are added to the reactor if a stream is
created without the loop running. i'm making the assumption that the next_tick
done with blocking dns is intending what i'm testing for -- no handles in the
reactor without actually running the loop

i first i thought the issue was with Net::DNS::Native. there's some more details at: https://github.com/olegwtf/p5-Net-DNS-Native/pull/5

not creating these objects without the loop running also avoids some warnings from NDN about destroying objects early.

### DIFF
--- a/lib/Mojo/IOLoop/Client.pm
+++ b/lib/Mojo/IOLoop/Client.pm
@@ -45,21 +45,24 @@ sub connect {
     if !NNR || $args->{handle} || $args->{path};
 
   # Non-blocking name resolution
-  $NDN //= Net::DNS::Native->new(pool => 5, extra_thread => 1);
-  my $handle = $self->{dns} = $NDN->getaddrinfo($address, _port($args),
-    {protocol => IPPROTO_TCP, socktype => SOCK_STREAM});
-  $reactor->io(
-    $handle => sub {
-      my $reactor = shift;
+  $reactor->next_tick(sub {
+    my $reactor = shift;
+    $NDN //= Net::DNS::Native->new(pool => 5, extra_thread => 1);
+    my $handle = $self->{dns} = $NDN->getaddrinfo($address, _port($args),
+      {protocol => IPPROTO_TCP, socktype => SOCK_STREAM});
+    $reactor->io(
+      $handle => sub {
+        my $reactor = shift;
 
-      $reactor->remove($self->{dns});
-      my ($err, @res) = $NDN->get_result(delete $self->{dns});
-      return $self->emit(error => "Can't resolve: $err") if $err;
+        $reactor->remove($self->{dns});
+        my ($err, @res) = $NDN->get_result(delete $self->{dns});
+        return $self->emit(error => "Can't resolve: $err") if $err;
 
-      $args->{addr_info} = \@res;
-      $self->_connect($args);
-    }
-  )->watch($handle, 1, 0);
+        $args->{addr_info} = \@res;
+        $self->_connect($args);
+      }
+    )->watch($handle, 1, 0);
+  });
 }
 
 sub _cleanup {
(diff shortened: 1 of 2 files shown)


lib/Mojo/IOLoop/Client.pm
```diff
@@ -45,21 +45,23 @@ sub connect {
...
-
-      $reactor->remove($self->{dns});
-      my ($err, @res) = $NDN->get_result(delete $self->{dns});
-      return $self->emit(error => "Can't resolve: $err") if $err;
-
-      $args->{addr_info} = \@res;
-      $self->_connect($args);
-    }
-  )->watch($handle, 1, 0);
+  $reactor->next_tick(sub {
```
Grinnz: You should add `my $reactor = shift;` here so that the callback does not close over the reactor it's being added to

t/mojo/stream_while_stopped.t
```diff
@@ -0,0 +1,12 @@
+use Mojo::Base -strict;
+
+use Test::More;
+use Mojo::IOLoop;
+
+Mojo::IOLoop->client({address => 'localhost'}, sub {});
+
+ok(!exists Mojo::IOLoop->singleton->reactor->{io}, 'no handles created');
```
kraih: We don't mess with private instance data even for tests.

kraih: And why does this deserve a whole new test file?

lib/Mojo/IOLoop/Client.pm
iamb: good catch, thank you

t/mojo/stream_while_stopped.t
```diff
@@ -0,0 +1,12 @@
+use Mojo::Base -strict;
+
+use Test::More;
+use Mojo::IOLoop;
+
+Mojo::IOLoop->client({address => 'localhost'}, sub {});
```
kraih: Is this actually portable?

kraih:
Closing this pull request for now since it has significant flaws and was not declared as a work in progress.

t/mojo/stream_while_stopped.t
kraih: On second thought, the whole test seems completely pointless since it doesn't even verify that `Net::DNS::Native` is installed.

kraih: To properly verify it works the same with and without you need two test files (along the lines of poll/ev tests), and find a scenario that doesn't rely on private instance data. And it would prolly have to be online tests, so the tests can be used to cover a few more interesting DNS cases.

iamb:
Just for clarity's sake:

Summary: Here's a deficiency (I claim) I found in Mojolicious. It has inconsistent behavior when an ioloop client connects with block vs. non-blocking DNS, and puts the user in a situation where they may unintentionally leak resources. The user shouldn't be depending on undocumented implementation details, but there's a small change that can improve the situation and avoid the problem.

Details: The current behavior of Mojo::IOLoop::Client->connect with blocking DNS is "don't do things til the loop starts." That's explicit and intentional as far as I can tell from the project history, though it's not documented or tested. Yes, this test should have just been left off. That was a poorly-done last second attempt to prevent the PR being ignored and immediately closed. I intended to respond to comments, and in fact was in the process of typing when it was closed. The "don't do things til ioloop starts" is currently untested as far as I can see, and I find no way to test it by project standards. I will not be able refactor the code as would be necessary to fix that.

I would have attempted to discuss this issue first, but for so small a change a PR with an explanation was the most efficient by a long shot. I don't have endless time to devote to your project.

If you want to drop the test and use this, great. If you want to just make the change yourself because you agree with my reasoning, great. If you choose not to take it, that's ok too, I'll just work around it in my code. I would prefer to see Mojolicious improved. The current situation has caused me real-world problems, and cost me a lot of time trying to track down.

PR #1306, opened 2018-12-29
Add Mojo::Promise handling to the eval command

jberger:
I find the eval command to be invaluable for one-off usage and testing, especially while developing. As I've been writing more promise apis and now as I'm writing fewer blocking apis due to the existence of Mojo::AsyncAwait, I'm finding the usage of promises in the eval command to be cumbersome to the point of not using it. With this change, one that I doubt would affect any usage in the wild (who wants to dump the stringified promise?), we can make a very comfortable interface for one-off testing of promise apis and make for easy examples and demos for newcomers or answer-seekers on irc.

### DIFF
--- a/lib/Mojolicious/Command/eval.pm
+++ b/lib/Mojolicious/Command/eval.pm
@@ -2,6 +2,7 @@ package Mojolicious::Command::eval;
 use Mojo::Base 'Mojolicious::Command';
 
 use Mojo::Util 'getopt';
+use Scalar::Util 'blessed';
 
 has description => 'Run code against application';
 has usage       => sub { shift->extract_usage };
@@ -16,7 +17,13 @@ sub run {
   my $app = $self->app;
   no warnings;
   my $result = eval "package main; sub app; local *app = sub { \$app }; $code";
-  return $@ ? die $@ : $result unless defined $result && ($v1 || $v2);
+  die $@ if $@;
+  if (blessed $result && $result->isa('Mojo::Promise')) {
+    my $err;
+    $result->then(sub { $result = shift }, sub { $err = shift })->wait;
+    die $err if $err;
+  }
+  return $result unless defined $result && ($v1 || $v2);
   $v2 ? print($app->dumper($result)) : say $result;
 }
 
@@ -48,7 +55,9 @@ Mojolicious::Command::eval - Eval command
 
 =head1 DESCRIPTION
 
-L<Mojolicious::Command::eval> runs code against applications.
+L<Mojolicious::Command::eval> runs code against applications. If the result is
+a L<Mojo::Promise>, it will wait until the promise is fulfilled or rejected and
+the result is returned.
 
 This is a core command, that means it is always enabled and its code a good
 example for learning to build new commands, you're welcome to fork it.
--- a/lib/Mojolicious/Command/eval.pm
+++ b/lib/Mojolicious/Command/eval.pm
@@ -1,8 +1,8 @@
 package Mojolicious::Command::eval;
 use Mojo::Base 'Mojolicious::Command';
 
+use Mojo::Promise;
 use Mojo::Util 'getopt';
-use Scalar::Util 'blessed';
 
 has description => 'Run code against application';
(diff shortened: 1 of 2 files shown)


lib/Mojolicious/Command/eval.pm
```diff
@@ -49,6 +56,7 @@ Mojolicious::Command::eval - Eval command
 =head1 DESCRIPTION
 
 L<Mojolicious::Command::eval> runs code against applications.
+If the result is a L<Mojo::Promise>, it will wait until the promise is completed and the result is returned.
```
jberger: Before the PR was opened @kraih commented on this line on a commit https://github.com/mojolicious/mojo/commit/4eaf550044e988635d41f4ce0349a2c18acf6b9d#r31802449

jberger: Do you mean just my added sentence or the existing one as well?

jberger: Oh! Do you just mean the literal line formatting? Yeah, I'll fix that, sorry.

lib/Mojolicious/Command/eval.pm
```diff
@@ -16,7 +17,13 @@ sub run {
   my $app = $self->app;
   no warnings;
   my $result = eval "package main; sub app; local *app = sub { \$app }; $code";
-  return $@ ? die $@ : $result unless defined $result && ($v1 || $v2);
+  die $@ if $@;
+  if (blessed $result && $result->isa('Mojo::Promise')) {
```
jhthorsen: Would it be to generic to use the same logic as inside Mojo::Promise instead?

https://github.com/mojolicious/mojo/blob/fb702a7a7618c82fff57b1a0d7d49ac71eb8ee7d/lib/Mojo/Promise.pm#L97

(Guess the documentation has to be updated as well then)

kraih:
I never had a need for this, so not sure how to vote.

lib/Mojolicious/Command/eval.pm
```diff
@@ -48,7 +55,9 @@ Mojolicious::Command::eval - Eval command
 
 =head1 DESCRIPTION
 
-L<Mojolicious::Command::eval> runs code against applications.
+L<Mojolicious::Command::eval> runs code against applications. If the result is
+a L<Mojo::Promise>, it will wait until the promise is completed and the result
```
kraih: The documentation for `Mojo::Promise::wait` says "fulfilled or rejected" instead of "completed".

jberger: I can make this change the next time I'm at a computer

jberger:
> I never had a need for this, so not sure how to vote.

I had only rarely until a new project at $work. Since async functions necessarily return promises and since they nearly obviate the need for blocking apis I now have an app for which the model is entirely promises. I think once aa becomes more common this will be the case more and more often, it feels very natural

lib/Mojolicious/Command/eval.pm
jberger: As of the second commit in this PR, arbitrary then-ables are now handled.

jberger:
@marcusramberg I know you're on vacation, but if you get a second, I'd appreciate a review/vote

jberger:
Per the new PR policy, I'm calling a vote on this PR. @mojolicious/core

kraih:
Neutral vote from me.

marcusramberg:
I'm also in favor of this change 👍

s1037989:
What do you think about adding a Usage example?

```
Usage: APPLICATION eval [OPTIONS] CODE

  ./myapp.pl eval 'say app->ua->get("/")->result->body'
  ./myapp.pl eval 'say for sort keys %{app->renderer->helpers}'
  ./myapp.pl eval -v 'app->home'
  ./myapp.pl eval -V 'app->renderer->paths'
  ./myapp.pl eval -V '
```

I don't know if this is a good example, but something like:
```
  ./myapp.pl eval -v 'app->ua->get_p("https://mojolicious.org")->then(sub{shift->result->dom->at("title")->text})'
```

Issue #716, opened 2014-12-11
Net::DNS::Native and Mojolicious on Linux


olegwtf:
As we noticed yesterday `Net::DNS::Native` with perl not linked with pthreads still may crash on Linux: https://rt.cpan.org/Public/Bug/Display.html?id=100834

Today I found a way to reproduce this problem with the following server

``` Perl
use Mojolicious::Lite;
use Mojo::IOLoop;

get '/' => sub {
        my $self = shift;

        Mojo::IOLoop->timer(rand(1) => sub {
                $self->render(json => [qw/a b c d e d/]);
        });
};

app->start;
```

and client

``` Perl
use strict;
use Mojo::UserAgent;
use Mojo::IOLoop;

my $ua = Mojo::UserAgent->new;
my @links = ("http://data-flow.ru:3000") x 5000;

while (@links) {
        Mojo::IOLoop->delay->new->steps(
                sub {
                        my @l = splice @links, 0, 50;
                        my $delay = shift;
                        $ua->get($_ => $delay->begin) for @l;
                },
                sub {
                        warn "+50";
                }
        )->wait
}
```

where you need to change data-flow.ru to your own domain (but not localhost).
This client segfaults 10 times out of 10

And the fix is as simple as

``` diff
 use strict;
+use Net::DNS::Native;
 use Mojo::UserAgent;
 use Mojo::IOLoop;
```

And all the problem caused by ... my favorite `IO::Socket::IP`. Look. Mojolicious loads `IO::Socket::IP` [here](https://metacpan.org/source/SRI/Mojolicious-5.68/lib/Mojo/IOLoop/Client.pm#L5) and `IO::Socket::IP` has [this](https://metacpan.org/source/PEVANS/IO-Socket-IP-0.34/lib/IO/Socket/IP.pm#L39) construction which executes `getprotobyname()` at compile time. Looks ok, but this loads non thread safe DNS part of libc (because our perl not linked with pthreads). And only after that Mojolicious [loads](https://metacpan.org/source/SRI/Mojolicious-5.68/lib/Mojo/IOLoop/Client.pm#L13) `Net::DNS::Native`, which can't replace non-thread safe DNS part of libc with thread safe even with [RTLD_GLOBAL](https://metacpan.org/source/OLEG/Net-DNS-Native-0.13/lib/Net/DNS/Native.pm#L23) flag.
So, you can comment mentioned line of `IO::Socket::IP` and all will work good
[...]

PR #1308, opened 2018-12-31
Load Net::DNS::Native before IO::Socket::IP

Grinnz:
### Summary
With the current load order Net::DNS::Native is never used in one setup because it fails to load due to IO::Socket::IP defining non thread safe symbols.

### Motivation
Loading it first seems to fix the issue.

### References
```
***********************************************************************
Some package defined non thread safe symbols which Net::DNS::Native uses internally
Please make sure you are not placed loading of modules like IO::Socket::IP
before this one and not called functions like getaddrinfo(), gethostbyname(),
inet_aton() before loading of Net::DNS::Native
************************************************************************
```

### DIFF
--- a/lib/Mojo/IOLoop/Client.pm
+++ b/lib/Mojo/IOLoop/Client.pm
@@ -1,6 +1,13 @@
 package Mojo::IOLoop::Client;
 use Mojo::Base 'Mojo::EventEmitter';
 
+# Non-blocking name resolution requires Net::DNS::Native
+# This must be loaded before IO::Socket::IP
+use constant NNR => $ENV{MOJO_NO_NNR}
+  ? 0
+  : eval { require Net::DNS::Native; Net::DNS::Native->VERSION('0.15'); 1 };
+my $NDN;
+
 use Errno 'EINPROGRESS';
 use IO::Socket::IP;
 use IO::Socket::UNIX;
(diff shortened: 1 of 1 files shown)


Grinnz:
I can't think of any reasonable way to test this.

Grinnz:
Through further testing both IO::Socket::IP and Mojo::IOLoop::TLS must be loaded after, probably due to Mojo::IOLoop::TLS loading IO::Socket::SSL.

Grinnz:
```
15:30:13 <Grinnz> https://github.com/mojolicious/mojo/issues/716 some context
15:31:18 <Grinnz> i believe it's because this perl was built without pthreads
15:31:29 <Grinnz> which was not default until ...
15:31:52 <Grinnz> 5.22
```

kraih:

marcusramberg:
I'm not sure if optimizing for older Perls is worth it. Are there any downsides to the change?

Grinnz:
The downsides are the introduced inconsistency in code layout, and that it can only be a partial solution, the setup can still break randomly (by break I mean it just won't use NDN) but at least Mojo::IOLoop::Client can load it in some cases.

Grinnz:
For the record, I've worked around this in my code by loading Net::DNS::Native before anything from Mojo.

jhthorsen:
I think since it will randomly break, then I'm negative to this change. I think a documentation update would be better.

jberger:
As I mentioned on IRC, and agreeing with others, I don't think that because it cannot really address the issue I don't think this change is merited. I agree with @jhthorsen that a documentation patch would be better however I really wonder if this is an upstream issue with NDN and IO::Socket::*

kraih:
Thanks all, with at least two negative votes this pull request cannot pass the vote.

leonerd:
As author of `IO::Socket::IP` - is there anything for me to do with this, or that I should know about?

Grinnz:
Only what is in the referenced issue #716.

Issue #1310, opened 2019-01-06
Extending the Mojo::Promise API


kraih:
This comes up a lot on IRC, so i'll open this issue to collect ideas and votes for extensions to the [Mojo::Promise API](https://metacpan.org/pod/Mojo::Promise). So far we have limited the API mostly to what [JavaScript Promises](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) contain, which is not much at all. The small API does have the advantage that it is easy to learn, but very often Mojolicious users now end up making their own `Mojo::Promise::Role::*` extensions or even switching to [Future](https://metacpan.org/pod/Future) for more features. So, what are the most commonly used features that you would like to see added?

kraih:
Personally, i very much like [bluebird promises](http://bluebirdjs.com/docs/api-reference.html) and there are two features i would like to see added. A [timeout](http://bluebirdjs.com/docs/api/timeout.html) and a [map](http://bluebirdjs.com/docs/api/promise.map.html) method.
```perl
# The timeout method would create a promise that rejects in 5 seconds with the error "Timeout!"
Mojo::Promise->race($promise1, $promise2, Mojo::Promise->timeout(5, 'Timeout!'))->then(sub {...});

# The map method would allow for 5 async operations to run at the same time here
Mojo::Promise->map(\@urls, sub { $url->get_p($_) }, {concurrency => 5})->then(sub {...});
```

Grinnz:
Some more options could come from the available convergent future constructors from Future (`wait_any` is analogous to Mojo::Promise `race`, and `needs_all` is analogous to Mojo::Promise `all`). `wait_all` would be problematic because it unconditionally succeeds with the component promises and relies on being able to introspect the results, so I won't propose an analogue here.

`needs_any` would be similar to the current `all`, but acts in the opposite manner. It resolves immediately with the first promise that resolves, or rejects once all promises have rejected. The use case would be similar to `race` but when you want to try each promise until one is successful.

    Mojo::Promise->needs_any(
      $ua->get_p('http://mirror1.com/foo'),
      $ua->get_p('http://mirror2.com/foo'),
    )->then(sub { ... })->catch(sub { ... });

The name should of course be something better matching the existing API.

niczero:
[...]

PR #1311, opened 2019-01-06
Add Mojo::Promise->timeout

jhthorsen:
### Summary
Adding a timeout method to avoid having to wrap Mojo::IOLoop->timeout(...) in a promise.

### Motivation
To make Mojo::Promise simpler to use in a generic usecase.

### References
https://github.com/mojolicious/mojo/issues/1310#issuecomment-451748766

### DIFF
--- a/lib/Mojo/Promise.pm
+++ b/lib/Mojo/Promise.pm
@@ -63,6 +63,13 @@ sub then {
   return $new;
 }
 
+sub timeout {
+  my ($self, $after, $err)
+    = (ref $_[0] ? shift : shift->new, @_, 'Promise timeout');
+  $self->ioloop->timer($after => sub { $self->reject($err) });
+  return $self;
+}
+
 sub wait {
   my $self = shift;
   return if (my $loop = $self->ioloop)->is_running;
@@ -347,6 +354,15 @@ L<Mojo::Promise> object resolving to the return value of the called handler.
       return "This is bad: $reason[0]";
     }
   );
+
+=head2 timeout
+
+  my $promise = $promise->timeout(5 => 'Timeout!');
+  my $promise = $promise->timeout(5);
+
+Create a new L<Mojo::Promise> that will be rejected after a given amount of
+time in seconds. A default rejection reason will be provided, unless specified
+as input.
 
 =head2 wait
 
--- a/t/mojo/promise.t
+++ b/t/mojo/promise.t
@@ -134,7 +134,7 @@ is_deeply \@results, ['pass'], 'promise resolved';
 
 # Clone
 my $loop = Mojo::IOLoop->new;
-$promise  = Mojo::Promise->new(ioloop => $loop)->resolve('failed');
+$promise = Mojo::Promise->new(ioloop => $loop)->resolve('failed');
 $promise2 = $promise->clone;
 (@results, @errors) = ();
 $promise2->then(sub { @results = @_ }, sub { @errors = @_ });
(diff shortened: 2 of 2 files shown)


lib/Mojo/Promise.pm
```diff
@@ -63,6 +63,12 @@ sub then {
   return $new;
 }
 
+sub timeout {
+  my ($self, $after, $err) = (blessed $_[0] ? shift->clone : shift->new, @_);
```
kraih: Other methods use `ref` instead of `blessed`.

kraih: The use of `clone` here is actually inconsistent (and there appears to be no test for it).

kraih: There should also be a default error message, so `Mojo::Promise->timeout(5)` works too.

jhthorsen: What do you want the default error message to be? "Timeout"..?

kraih: "Promise timeout" would seem reasonable, no?

jhthorsen:
> Why not make this more clean?

`shift // 'Promise timeout'` is not the same, since if someone wants to pass in `undef` as $err, then that cannot be done. If you don't want that to be supported, then I'll change it.

kraih:
Is there a use case for `undef` as error message?

jhthorsen:
The only thing I can think of is if you’re using finally() instead of then/catch.

t/mojo/promise.t
```diff
@@ -181,6 +181,20 @@ Mojo::IOLoop->one_tick;
...
+# Timeout
+(@errors, @results) = @_;
+$promise = Mojo::Promise->timeout(0.25 => 'Timeout1');
+$promise2 = Mojo::Promise->new->timeout(0.025 => 'Timeout2');
+$promise3
+  = Mojo::Promise->race($promise, $promise2)->then(sub { @results = @_ })
+  ->catch(sub { @errors = @_ })->wait;
+is_deeply \@results, [], 'promises not resolved';
+is_deeply \@errors, ['Timeout2'], 'promise rejected';
+
```
kraih: Extra newline.

jhthorsen: I added a newline to indicate a new test case. Should I remove the newline or add a new heading, such as `# Timeout with default catch message`?

Grinnz:
I think it's reasonable for `->timeout(5 => undef)` to result in the default timeout message instead of an undefined one.

jhthorsen:
> I think it's reasonable for `->timeout(5 => undef)` to result in the default timeout message instead of an undefined one.

I disagree. I also don't see when someone would pass in `undef` as the last argument, unless that is intentional. I do however think `timeout(5)` is a good case for a default value to be passed on to the rejection handlers.

t/mojo/promise.t
kraih: Yes.

jhthorsen: My question was "should I remove the newline OR add a new heading". I'm not sure what you're answering "Yes" to.

jhthorsen: I added a new heading, since it's a new test case. I think that makes sense.

lib/Mojo/Promise.pm
```diff
@@ -348,6 +355,14 @@ L<Mojo::Promise> object resolving to the return value of the called handler.
     }
   );
 
+=head2 timeout
+
+  my $new = Mojo::Promise->timeout(5 => 'Timeout!');
+  my $new = $promise->timeout(5 => 'Timeout!');
```
jhthorsen: Should the default rejection be documented? If so, how is the best way to do that?

kraih: At the very least a `my $new = $promise->timeout(5);` here.

jhthorsen:
@mojolicious/core: I'm calling for a vote.

marcusramberg:
I also find this reasonable and a good addition 👍

jberger:
:+1:

jhthorsen:
I guess the vote passed.

Thanks everyone :)

PR #1314, opened 2019-01-29
Add get methods to Mojo::Promise

Grinnz:
### Summary
Add a `get` method to Mojo::Promise for retrieving results in a synchronous manner.

### Motivation
The `get` method in Future is widely used to similarly await the results of a Future. The `eval` core command can handle promises much more simply using this method.

### References
https://github.com/mojolicious/mojo/issues/1310#issuecomment-452119631, https://metacpan.org/pod/Mojo::Promise::Role::Get, https://metacpan.org/pod/Future#get

### DIFF
--- a/lib/Mojo/Promise.pm
+++ b/lib/Mojo/Promise.pm
@@ -1,6 +1,7 @@
 package Mojo::Promise;
 use Mojo::Base -base;
 
+use Carp 'croak';
 use Mojo::IOLoop;
 use Scalar::Util 'blessed';
 
@@ -40,6 +41,20 @@ sub finally {
 
   return $new;
 }
+
+sub get_all {
+  my $self = shift;
+  croak q{'get' cannot be called when the event loop is running}
+    if $self->ioloop->is_running;
+  my ($res, $err);
+  my $p = $self->then(sub { $res = [@_] }, sub { $err = [@_] });
+  $p->wait until $res or $err;
+  return $res if $res;
+  my $reason = $err->[0] // 'Promise rejected';
+  ref($reason) || $reason =~ m/\n\z/ ? die $reason : croak $reason;
+}
+
+sub get_one { shift->get_all->[0] }
 
 sub race {
   my ($class, @promises) = @_;
@@ -175,6 +190,12 @@ Mojo::Promise - Promises/A+
     warn "Something went wrong: $err";
   })->wait;
 
+  # Get results synchronously or throw exception if rejected
+  my $mojo = get('https://mojolicious.org');
+  my $cpan = get('https://metacpan.org');
+  my $responses = Mojo::Promise->all($mojo, $cpan)->get_all;
+  say $_->[0]->res->code for @$responses;
+
 =head1 DESCRIPTION
 
 L<Mojo::Promise> is a Perl-ish implementation of
@@ -282,6 +303,21 @@ reason.
     my @value_or_reason = @_;
     say "We are done!";
   });
+
+=head2 get_all
+
+  my $results = $promise->get_all;
+
+Start L</"ioloop"> and stop it again once the promise has been fulfilled or
+rejected, then return the results in an array reference if it was fulfilled, or
+throw an exception with the reason if it was rejected. An exception will be
+thrown if the L</"ioloop"> is already running.
+
+=head2 get_one
+
+  my $result = $promise->get_one;
+
+Similar to L</"get_all">, but returns the first result value.
 
 =head2 race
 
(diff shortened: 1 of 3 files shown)


lib/Mojo/Promise.pm
```diff
@@ -41,6 +42,19 @@ sub finally {
...
+  my ($self) = @_;
+  croak q{'get' cannot be called when the event loop is running}
+    if $self->ioloop->is_running;
+  my (@result, $rj);
+  $self->then(sub { @result = @_ }, sub { $rj++; @result = @_ })->wait;
+  if ($rj) {
+    my $reason = $result[0] // 'Promise was rejected';
+    ref($reason) || $reason =~ m/\n\z/ ? die $reason : croak $reason;
+  }
+  return wantarray ? @result : $result[0];
```
jhthorsen: Why use `$rj`? How about this:

```
my ($err, $res);
$self->then(sub { $res = [@_] }, sub { $err = [@_] })->wait;
return wantarray ? @$res : $res->[0] if $res;
$err->[0] //= 'Promise was rejected';
ref($err->[0]) || $err->[0] =~ m/\n\z/ ? die $err->[0] : croak $err->[0];
```

jhthorsen: I'm glad we saved a couple of lines though 👍

lib/Mojo/Promise.pm
```diff
@@ -41,6 +42,19 @@ sub finally {
...
 }
 
+sub get {
+  my ($self) = @_;
+  croak q{'get' cannot be called when the event loop is running}
+    if (my $loop = $self->ioloop)->is_running;
+  my ($res, $err);
+  $self->then(sub { $res = [@_]; $loop->stop },
+    sub { $err = [@_]; $loop->stop });
+  $loop->start;
```
Grinnz: Opinion: should this be `$loop->start until $res or $err;` so that it is not interrupted by something else stopping the loop?

jhthorsen:
> I replaced the call to `wait` with just directly stopping the loop as I don't think the extra promise step is worth the slightly shorter code.

I'm 👎 to this change. I don't see the value with this extra code.

Grinnz:
Look at the source of `wait`. It is an entire extra promise step that is unnecessary since we are already adding both resolve and rejection handlers.

jhthorsen:
I understand your argument @Grinnz, but it's just not worth it to me.

kraih:
The code is a lot more ugly than i expected, so i guess i'm 👎 on this implementation now too.

kraih:
To be more precise, things i did not expect were `wantarray`, `die/croak` hacks, and manual loop starting/stopping. If consensus is that those things are really necessary i believe this feature is better off in a role on CPAN.

kraih:
After some discussion on IRC i'm afraid i have come to the conclusion that i have to veto this proposal because it uses `wantarray`. 🙅‍♂️

Grinnz:
Instead of wantarray, there are now two methods, `get_one` and `get_all`. I am not attached to the names, and some other ideas I have are `get_first`, `get_every`, `get_reason`, and `get_reasons`. Please suggest alternatives you prefer.

lib/Mojo/Promise.pm
```diff
@@ -283,6 +304,24 @@ reason.
...
+the L</"ioloop"> is already running.
+
+=head2 get_one
+
+  my $result = $promise->get_one;
+
+Start L</"ioloop"> and stop it again once the promise has been fulfilled or
+rejected, then return the first result if it was fulfilled, or throw an
+exception with the reason if it was rejected. An exception will be thrown if
+the L</"ioloop"> is already running.
```
kraih: Don't duplicate the same text, just link to `get_all` and document the differences. Like other similar methods do.

jhthorsen:
I’m not very positive anymore.

I think we should just add get() that returns the first item. Optimize for what I hope is the most common use case.

Grinnz:
I don't believe it is possible to say either form is more common, `get` is often used with convergent Futures in Future where you need all of the results. But like you said, always returning an arrayref is not optimal for the other common cases.

jhthorsen:
Once I figure out how, I’m going to suggest making a breaking change, where the promise-callbacks can only take one argument. So by adding get() that always returns the first argument will encourage writing code, where the callbacks only take one argument - I think.

kraih:
Looks like this proposal as it stands will not pass a vote.

kraih:
Honestly, i think a simple `Mojo::Promise::get` that only returns the first value would cover most common use cases.

marcusramberg:
I guess there's no reason to leave this open.

PR #1315, opened 2019-01-29
Wait for promise to resolve regardless of loop being stopped

Grinnz:
### Summary
Ensure that `wait` will keep running the loop until the promise is no longer pending.

### Motivation
Calling `wait` can result in a still-pending promise even if the loop is not running, as anything may stop the loop. This small change ensures that it will not return while the promise is pending, which is the documented behavior and user expectation. This is also important for the proposed `get` method(s), to ensure it does not return before the results are available.

### References
Prior art: the implementations of `await` and `get` in Future event loop subclasses will loop until the Future is no longer pending.


kraih:
Wanted to merge this PR, but there are now conflicts.

Grinnz:
Rebased

PR #1319, opened 2019-02-08
Make using a unknwon type in routes fatal

2shortplanks:
### Summary
When creating routes that use placeholders that have types, the routes should throw an exception when matching instead of simply failing to match anything if a route type used does not exist.

### Motivation
This is a hard to debug problem where a typo in a placeholder type can result in routes not working in a subtle way

### References

IRC 2019-02-08
>&lt;markfowler> So, what the logic behind having a route that uses a type that doesn't exist resulting in a route that never matches
>12:34 PM rather than going bang
>12:37 PM &lt;kraih> ?
>12:38 PM &lt;markfowler> Mark Fowler So If I do $r->get("/foo/<bar:a_type_i_have_not_declared")
>12:38 PM Then I get an unmatchable route, not some sort of immediate error
>12:38 PM &lt;kraih> implementation detail, feel free to propose improvements

### DIFF
--- a/lib/Mojolicious/Routes/Pattern.pm
+++ b/lib/Mojolicious/Routes/Pattern.pm
@@ -113,7 +113,7 @@ sub _compile {
     # Placeholder
     else {
       if ($value =~ /^(.+)\Q$start\E(.+)$/) {
-        ($value, $part) = ($1, _compile_req($types->{$2} // '?!'));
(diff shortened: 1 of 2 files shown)


kraih:
I feared you would take a shortcut like this. That's why i warned you on IRC that a real solution might require very big changes. Had you added higher level tests you would have noticed that the compilation of routes is deferred until the route matches the first time. So this solution doesn't really work.

Issue #1320, opened 2019-02-10
Feature Request: Mojo::Collection - shift, unshift, push, pop


srchulo:
It could be nice to have these as methods on Mojo::Collection. I had an example recently where I had to break the nice flow of `->`:

```
my $array = $collection->arrays->flatten->to_array;
unshift @$array, 'item';
```

It would be great to be able to do something like:

`my $array = $collection->arrays->flatten->unshift('item')->to_array;`

kiwiroy:
There is this pattern:
```perl
my $array = $collection->arrays->flatten->tap(sub { unshift @$_, 'item' })->to_array;
```

srchulo:
ah, that is better than what I was doing before, thanks! I still think it would make sense to have common operations like `shift`, `unshift`, `push`, and `pop` be actual methods on the Collection.

s1037989:
I've also had situations where I've needed to break the nice flow and have often wished for this.

kiwiroy:
The current methods on `Mojo::Collection` all return copies of the collection with the transformation applied, whereas the request above appears like an **i**n place transformation is desired, though that might be side effect of the current situation.

For [example](https://github.com/kiwiroy/mojo/compare/collection-array-funcs) here is `unshift`/`unshift_i`, since the examples above use `unshift`.

kraih:
I'm vetoing this feature for now, since the two proposed solutions had very substantial flaws. Please try thinking more about what the actual use cases are for each method you're adding before sending more pull requests. Maybe try having a discussion about it on IRC.

kraih:
If you're sure you've found the right semantics, please provide two independent use cases for each method that you believe to be strong.

kraih:
Does not look like there is much support for this feature.

kraih:
Doesn't look like this proposal got any more support. Thanks everyone.

PR #1321, opened 2019-02-11
adding pop, push, shift, unshift

kiwiroy:
### Summary
Adding `pop`, `push`, `shift`, and `unshift` as chain able methods to `Mojo::Collection`. Qualifying `shift` makes this quite a noisey PR.

### Motivation
There were a few positive expressions made to the feature request raised by @srchulo.

### References
#1320

### DIFF
--- a/lib/Mojo/Collection.pm
+++ b/lib/Mojo/Collection.pm
@@ -9,13 +9,12 @@ use Scalar::Util 'blessed';
 
 our @EXPORT_OK = ('c');
 
-sub TO_JSON { [@{shift()}] }
+sub TO_JSON { [@{$_[0]}] }
 
 sub c { __PACKAGE__->new(@_) }
 
 sub compact {
-  my $self = shift;
-  return $self->new(grep { defined && (ref || length) } @$self);
+  return $_[0]->new(grep { defined && (ref || length) } @{$_[0]});
 }
 
 sub each {
@@ -27,7 +26,7 @@ sub each {
 }
 
 sub first {
-  my ($self, $cb) = (shift, shift);
+  my ($self, $cb) = (CORE::shift, CORE::shift);
   return $self->[0] unless $cb;
   return List::Util::first { $_ =~ $cb } @$self if ref $cb eq 'Regexp';
   return List::Util::first { $_->$cb(@_) } @$self;
@@ -36,7 +35,7 @@ sub first {
 sub flatten { $_[0]->new(_flatten(@{$_[0]})) }
 
 sub grep {
-  my ($self, $cb) = (shift, shift);
+  my ($self, $cb) = (CORE::shift, CORE::shift);
   return $self->new(grep { $_ =~ $cb } @$self) if ref $cb eq 'Regexp';
   return $self->new(grep { $_->$cb(@_) } @$self);
 }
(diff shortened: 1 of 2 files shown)


lib/Mojo/Collection.pm
```diff
@@ -9,12 +9,12 @@ use Scalar::Util 'blessed';
 
 our @EXPORT_OK = ('c');
 
-sub TO_JSON { [@{shift()}] }
+sub TO_JSON { [@{CORE::shift()}] }
```
jhthorsen: I’m not sure what is best to read, but maybe many of these CORE::shift calls could be replaced by $_[0]?

lib/Mojo/Collection.pm
```diff
@@ -45,32 +45,47 @@ sub join {
...
+  my $self = CORE::shift;
   @_ = (@_, @$self);
   goto &List::Util::reduce;
 }
 
 sub reverse { $_[0]->new(reverse @{$_[0]}) }
 
+sub shift {
+  my $self = CORE::shift;
+  return $self->slice(1 .. $#$self);
```
jhthorsen: Doing this extra method call seems a bit slow. How about doing the array slice directly instead?

jhthorsen:
I guess you could argue that they won’t work as the CORE:: methods, since they return a number, that change is ok by me, since I don’t think the return value if very often used.

Grinnz:
I have to agree that I would expect these methods to mutate the original collection, as an exceptional case.

kiwiroy:
Agreed, `CORE::shift` is definitely on the hard to read side of the line, so I've made that change where it makes sense. `pop`, `slice` and `unshift` can be golfed further, but `slice` in particular loses legibility, at least for me:

```patch
diff --git a/lib/Mojo/Collection.pm b/lib/Mojo/Collection.pm
index ba36a3091..2342a272b 100644
--- a/lib/Mojo/Collection.pm
+++ b/lib/Mojo/Collection.pm
@@ -61,8 +61,7 @@ sub pop {
 }
 
 sub push {
-  my $self = CORE::shift;
-  return $self->new(@$self, @_);
+  return $_[0]->new(@{$_[0]}, @_[1 .. $#_]);
 }
 
 sub reduce {
@@ -82,8 +81,7 @@ sub shuffle { $_[0]->new(List::Util::shuffle @{$_[0]}) }
 sub size { scalar @{$_[0]} }
 
 sub slice {
-  my $self = CORE::shift;
-  return $self->new(@$self[@_]);
+  return $_[0]->new(@{$_[0]}[ @_[1 .. $#_] ]);
 }
 
 sub sort {
@@ -112,8 +110,7 @@ sub uniq {
 }
 
 sub unshift {
-  my $self = CORE::shift;
-  return $self->new(@_, @$self);
+  return $_[0]->new(@_[1 .. $#_], @{$_[0]});
 }
 
 sub with_roles { CORE::shift->Mojo::Base::with_roles(@_) }
```

Method call to `slice` removed.

Should `Mojo::Collection::Role::Mutating` be a thing?

kraih:
Agreed, creating new collections seems wrong to me. 👎

marcusramberg:
👎 would also want these to mutate the existing collection, not return new ones.

marcusramberg:
Sorry, but this PR did not pass the vote.

PR #1323, opened 2019-02-14
adding pop, push, shift, unshift - mutating

kiwiroy:
### Summary
Adding `pop`, `push`, `shift`, and `unshift` as chain able methods to `Mojo::Collection`.

### Motivation
There were a few positive expressions made to the feature request raised by @srchulo and an aversion to creating new collections in the implementations contained in the previous PR.

### References
* #1320 
* #1321

### DIFF
--- a/lib/Mojo/Collection.pm
+++ b/lib/Mojo/Collection.pm
@@ -9,13 +9,12 @@ use Scalar::Util 'blessed';
 
 our @EXPORT_OK = ('c');
 
-sub TO_JSON { [@{shift()}] }
+sub TO_JSON { [@{$_[0]}] }
 
 sub c { __PACKAGE__->new(@_) }
 
 sub compact {
-  my $self = shift;
-  return $self->new(grep { defined && (ref || length) } @$self);
+  return $_[0]->new(grep { defined && (ref || length) } @{$_[0]});
 }
 
 sub each {
@@ -27,7 +26,7 @@ sub each {
 }
 
 sub first {
-  my ($self, $cb) = (shift, shift);
+  my ($self, $cb) = (CORE::shift, CORE::shift);
   return $self->[0] unless $cb;
   return List::Util::first { $_ =~ $cb } @$self if ref $cb eq 'Regexp';
   return List::Util::first { $_->$cb(@_) } @$self;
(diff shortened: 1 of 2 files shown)


lib/Mojo/Collection.pm
```diff
@@ -88,18 +102,23 @@ sub sort {
   return $self->new(@sorted);
 }
 
-sub tap { shift->Mojo::Base::tap(@_) }
+sub tap { CORE::shift->Mojo::Base::tap(@_) }
```
jhthorsen: Isn't it better with just `sub tap { Mojo::Base::tap(@_) }`?

lib/Mojo/Collection.pm
```diff
@@ -88,18 +102,23 @@ sub sort {
...
   return $self->new(grep { !$seen{$_ // ''}++ } @$self);
 }
 
-sub with_roles { shift->Mojo::Base::with_roles(@_) }
+sub unshift {
+  CORE::unshift @{$_[0]}, @_[1 .. $#_];
+  $_[0];
+}
+
+sub with_roles { CORE::shift->Mojo::Base::with_roles(@_) }
```
jhthorsen: Isn't it better with just `sub tap { Mojo::Base::with_roles(@_) }`?

lib/Mojo/Collection.pm
```diff
@@ -304,6 +323,32 @@ C<$b> will always be set to the next element in the collection.
 
 Create a new collection with all elements in reverse order.
 
+=head2 pop
+
+  # [1, 2, 3, 4]
+  my $new = c(1 .. 5)->pop;
```
jhthorsen: I think this is very confusing that the variable is called `$new`. Sounds like it's not mutating, even though the documentation says so.

kraih:
Agreed, the implementation is too ugly and not consistent with Mojo code style. 👎

jhthorsen:
@kraih: Could you comment inside the code where you want it to be changed? Is it as simple as doing something like this?

```
# Alt 1.
sub pop {
  my $self = shift;
  pop(@$self);
  return $self;
}

# Alt 2.
sub pop { $_[0]->tap(sub { pop @$_ }) }
```

I guess alternative two is most compact, but probably super slow..?

Also, do you disagree with my comment below @kraih?

> Isn't it better with just `sub tap { Mojo::Base::with_roles(@_) }`?

kraih:
@jhthorsen I commented outside because i don't like any of it. Not even the documentation is consistent.

jhthorsen:
Thanks for the clarification.

I'm leaning toward that this functionality isn't really needed.

kraih:
`shift` and `pop` methods that remove an element from the collection and return the invocant make absolutely no sense and will cause endless confusion in the future. Therefore I'm afraid i have to veto this feature.

PR #1325, opened 2019-02-18
Add map class method to Mojo::Promise

jberger:
Add a map method that allows delaying of portions of a set of promises so as to limit concurrency.

### DIFF
--- a/lib/Mojo/Promise.pm
+++ b/lib/Mojo/Promise.pm
@@ -41,6 +41,31 @@ sub finally {
   return $new;
 }
 
+sub map {
+  my ($class, $options) = (shift, ref $_[0] eq 'HASH' ? shift : {});
+  my ($cb, @items) = @_;
+
+  my @start = map { $_->$cb } splice @items, 0,
+    $options->{concurrency} // @items;
+  my $proto = $class->resolve($start[0]);
+
+  my (@trigger, @wait);
+  for my $item (@items) {
+    my $p = $proto->clone;
+    push @trigger, $p;
+    push @wait, $p->then(sub { local $_ = $item; $_->$cb });
+  }
+
+  my @all = map {
+    $proto->clone->resolve($_)->then(
+      sub { shift(@trigger)->resolve if @trigger; @_ },
+      sub { @trigger = (); $proto->clone->reject($_[0]) },
+    )
+  } (@start, @wait);
+
+  return $class->all(@all);
+}
+
 sub race {
   my ($class, @promises) = @_;
   my $new = $promises[0]->clone;
@@ -289,6 +314,33 @@ reason.
     my @value_or_reason = @_;
     say "We are done!";
   });
+
+=head2 map
+
+  my $promise = Mojo::Promise->map(sub { ... }, @items);
+  my $promise = Mojo::Promise->map({concurrency => 3}, sub { ... }, @items);
+
+Apply a function that returns a L<Mojo::Promise> to each item in a list of
+items while optionally limiting concurrency. Returns a L<Mojo::Promise> that
+collects the results in the same manner as L</all>.
+
+Note that if any item's promise is rejected, any remaining items which have not
+yet been mapped will not be.
+
+  Mojo::Promise->map({concurrency => 3}, sub { $ua->get_p($_) }, @urls)
+    ->then(sub{ say $_->[0]->res->dom->at('title')->text for @_ });
+
+These options are currently available:
+
+=over 2
+
+=item concurrency
+
+  concurrency => 3
+
+The maximum number of items are in progress at the same time.
+
+=back
 
 =head2 race
 
--- a/lib/Mojo/Promise.pm
+++ b/lib/Mojo/Promise.pm
@@ -122,11 +122,12 @@ sub _finally {
 
 sub _settle {
   my ($self, $status) = (shift, shift);
-  $self = $self->new unless ref $self;
+  my $thenable = blessed $_[0] && $_[0]->can('then');
+  $self = $thenable ? $_[0]->clone : $self->new unless ref $self;
 
   $_[0]->then(sub { $self->resolve(@_); () }, sub { $self->reject(@_); () })
     and return $self
-    if blessed $_[0] && $_[0]->can('then');
+    if $thenable;
 
   return $self if $self->{result};
 
--- a/t/mojo/promise.t
+++ b/t/mojo/promise.t
@@ -280,4 +280,14 @@ is_deeply \@results, [], 'promise not resolved';
 is_deeply \@errors, [1], 'correct errors';
 is_deeply \@started, [1, 2, 3], 'only initial batch started';
 
+my $ok;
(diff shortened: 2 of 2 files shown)


jberger:
N.B. this draft PR still needs spit-and-polish. The style isn't fully consistent, the test isn't very "mojo-like" and should be expanded anyway, and of course there is no documentation. That said, the code should be sound and someone else can help this over the finish line or I'll get back to it eventually.

jberger:
I've now pushed some preliminary documentation. It is meant to use as a starting point and to prevent test failures due to missing documentation.

lib/Mojo/Promise.pm
```diff
@@ -290,6 +315,37 @@ reason.
...
+
+=over 2
+
+=item concurrency
+
+The maximum number of items are in progress at the same.
+
+=back
+
+C<concurrency> which specifies
```
Grinnz: This line looks accidental

jberger: It is, thanks

jberger:
Calling for a vote on this @mojolicious/core

marcusramberg:
Looks like a great addition. 👍 from me.

lib/Mojo/Promise.pm
```diff
@@ -290,6 +315,36 @@ reason.
...
+
+=back
+
+  my $promise = Mojo::Promise->map(
+    {concurrency => 3},
+    sub { $ua->get_p($_) },
+    @urls,
+  )->then(sub{
+    say $_->[0]->res->dom->at('title')->text for @_
+  });
```
kraih: This example is out of place.

t/mojo/promise.t
```diff
@@ -239,4 +239,37 @@ Mojo::IOLoop->one_tick;
...
+  sub {
+    my $n = $_;
+    push @started, $n;
+    Mojo::Promise->resolve->then(sub { Mojo::Promise->reject($n) });
+  },
+  1 .. 5
+)->then(sub { @results = @_ }, sub { @errors = @_ })->wait;
+is_deeply \@results, [], 'promise not resolved';
+is_deeply \@errors, [1], 'correct errors';
+is_deeply \@started, [1, 2, 3], 'only initial batch started';
```
kraih: There is no test without concurrency setting.

lib/Mojo/Promise.pm
```diff
@@ -290,6 +315,36 @@ reason.
...
+
+Note that if any item's promise is rejected, any remaining items which have not
+yet been mapped will not be.
+
+These options are currently available:
+
+=over 2
+
+=item concurrency
+
```
kraih: There's a `concurrency => 3` missing here, like other option lists have.

kraih:
What happens if the map closure returns something that is not a promise?

jhthorsen:
> What happens if the map closure returns something that is not a promise?

+1. I think the code will simply fail, since you can’t clone that inside the for-loop.

lib/Mojo/Promise.pm
```diff
@@ -41,6 +41,31 @@ sub finally {
   return $new;
 }
 
+sub map {
+  my ($class, $options) = (shift, ref $_[0] eq 'HASH' ? shift : {});
+  my ($cb, @items) = @_;
+
```
kiwiroy: There are no tests for empty `@items` and no promise returning check, e.g.
`return $class->reject('nothing to map') unless @items;`

kraih:
@marcusramberg, @jhthorsen I don't understand why you vote in favour when there is clearly tests missing for quite a few cases. So i'm afraid i have to veto this for now.

jberger:
I will address the aforementioned issues when I have the time, for now I guess this is on hold

jberger:
BTW, I would have appreciated this level of review when I asked for it, many times, while it was in the draft phase and before I called for a vote. Just so happens I had more time then.

jhthorsen:
@kraih: I meant +1 on your comment, not on the PR.

@jberger: I’m sorry, I was late with my feedback.

lib/Mojo/Promise.pm
jberger: As this just delegates to Mojo::Promise::all I assumed that this would just be a pass-through.

jberger:
Good catch on upgrading non-promise values. I've added that code and added a test for that case (which also tests without the concurrency option).

lib/Mojo/Promise.pm
jberger: We have methods in other classes which have a more involved example after the body of the documentation. I can't seem to find one with options too however. Where should we put such and example in that case?

lib/Mojo/Promise.pm
```diff
@@ -41,6 +41,31 @@ sub finally {
...
+  my @start = map { $class->resolve($_->$cb) } splice @items, 0,
```
jhthorsen: I don't think this will work, since resolve() on the class use the singleton ioloop and not the IOLoop from the item(s) you pass in.

jhthorsen: Maybe it's better to require that the map-function only return promises, since that's the requirement for the input to all()?

jberger: > Maybe it's better to require that the map-function only return promises, since that's the requirement for the input to all()?

That was my original intent.

kraih:
Looks like @jberger is confident he fixed all the issues, so i'm reopening votes for @mojolicious/core.

lib/Mojo/Promise.pm
```diff
@@ -290,6 +316,33 @@ reason.
...
+
+Apply a function that returns a L<Mojo::Promise> to each item in a list of
+items while optionally limiting concurrency. Returns a L<Mojo::Promise> that
+collects the results in the same manner as L</all>.
+
+Note that if any item's promise is rejected, any remaining items which have not
+yet been mapped will not be.
+
+  Mojo::Promise->map({concurrency => 3}, sub { $ua->get_p($_) }, @urls)
+    ->then(sub{ say $_->[0]->res->dom->at('title')->text for @_ });
```
jhthorsen: I think there’s a tiny glitch in the documentation here.

lib/Mojo/Promise.pm
```diff
@@ -97,11 +122,12 @@ sub _finally {
 
 sub _settle {
   my ($self, $status) = (shift, shift);
-  $self = $self->new unless ref $self;
+  my $thenable = blessed $_[0] && $_[0]->can('then');
+  $self = $thenable ? $_[0]->clone : $self->new unless ref $self;
```
jhthorsen: This is very nifty @jberger 🎉👍

lib/Mojo/Promise.pm
jberger: I'm not in a good position to check at the moment, but visually I don't see a glitch. What do you see @jhthorsen ?

jhthorsen: “sub{“ - missing a space.

Issue #1328, opened 2019-03-10
Feature request: support min and maybe max size for compress in Mojolicious::Renderer


srchulo:
In Molicious::Renderer there is support for compressing all dynamic content:

https://metacpan.org/pod/Mojolicious::Renderer#compress

It would be nice if compress supported a minimum size before compressing, and maybe a maximum.

kraih:
You're not explaining why you want this feature. Perhaps with some data backing up the request?

srchulo:
The reason is because compressing some smaller requests can actually make them larger, and also compressing requests below a certain size can be slower than not compressing at all. Here's a question on webmasters.stackexchange quoting some points from the Akamai CDN:

https://webmasters.stackexchange.com/questions/31750/what-is-recommended-minimum-object-size-for-gzip-performance-benefits

Here's a recommendation from Google mentioned there:

> Note that gzipping is only beneficial for larger resources. Due to the overhead and latency of compression and decompression, you should only gzip files above a certain size threshold; we recommend a minimum range between 150 and 1000 bytes. Gzipping files below 150 bytes can actually make them larger.

And here is why Akamai says that they only compress when a response is above 860 bytes:

> The reasons 860 bytes is the minimum size for compression is twofold: (1) The overhead of compressing an object under 860 bytes outweighs performance gain. (2) Objects under 860 bytes can be transmitted via a single packet anyway, so there isn't a compelling reason to compress them.

Nginx also supports a [minimum length before gzipping](https://nginx.org/en/docs/http/ngx_http_gzip_module.html#gzip_min_length), as does [Apache](http://www.thern.org/projects/mod_gzip) (see mod_gzip_minimum_file_size ).

kraih:
👍 Seems reasonable.

kraih:
Thanks, resolved by #1331.

PR #1330, opened 2019-03-11
#1328 support minimum size for compress option in Mojolicious::Renderer

mjemmeson:
### Summary
Adding an attribute to Mojolicious::Renderer to specify minimum size for compressing output.

### Motivation
Issue #1328 - feature request - efficiency in compression (small files not efficient to compress)

### References
https://github.com/mojolicious/mojo/issues/1328
and mentioned on IRC

### DIFF
--- a/lib/Mojolicious/Renderer.pm
+++ b/lib/Mojolicious/Renderer.pm
@@ -11,6 +11,7 @@ use Mojo::Util qw(decamelize encode gzip md5_sum monkey_patch);
 has cache   => sub { Mojo::Cache->new };
 has classes => sub { ['main'] };
 has [qw(compress default_handler)];
+has minimum_compress_size => '860';
 has default_format => 'html';
 has encoding       => 'UTF-8';
 has [qw(handlers helpers)] => sub { {} };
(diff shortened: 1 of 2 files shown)


lib/Mojolicious/Renderer.pm
```diff
@@ -123,12 +124,15 @@ sub respond {
   # Gzip compression
   my $res = $c->res;
   if ($self->compress) {
-    my $headers = $res->headers;
-    $headers->append(Vary => 'Accept-Encoding');
-    my $gzip = ($c->req->headers->accept_encoding // '') =~ /gzip/i;
-    if ($gzip && !$headers->content_encoding) {
-      $headers->content_encoding('gzip');
-      $output = gzip $output;
+    my $byte_length = do { use bytes; length($output) };
```
kraih: `$output` can only be bytes here, so that's unnecessary.

lib/Mojolicious/Renderer.pm
```diff
@@ -123,12 +124,15 @@ sub respond {
...
+    if ($byte_length >= $self->minimum_compress_size) {
```
kraih: This is very verbose, why not just return?

lib/Mojolicious/Renderer.pm
```diff
@@ -288,6 +292,15 @@ Try to negotiate compression for dynamically generated response content and
 C<gzip> compress it automatically, defaults to false. Note that this attribute
 is EXPERIMENTAL and might change without warning!
 
+=head2 minimum_compress_size
+
+  my $byte_length = $renderer->minimum_compress_size;
+  $renderer = $renderer->minimum_compress_size( $byte_length );
+
+Return or set a minimum size in bytes below which it is not efficient to compress
+the output. Default is 860 bytes, following the behaviour of Akamai CDN. Note that
```
kraih: Lets not advertise Akamai, they don't give us free stuff. 😉

lib/Mojolicious/Renderer.pm
```diff
@@ -11,6 +11,7 @@ use Mojo::Util qw(decamelize encode gzip md5_sum monkey_patch);
 has cache   => sub { Mojo::Cache->new };
 has classes => sub { ['main'] };
 has [qw(compress default_handler)];
+has minimum_compress_size => '860';
```
kraih: For consistency this should be `min_compress_size`.

lib/Mojolicious/Renderer.pm
kraih: The description is also inconsistent with similar attributes.

kraih:
Thanks, resolved by #1331.

PR #1331, opened 2019-03-11
Add minimum size to renderer compression

CandyAngel:
### Summary
Implements a minimum size of the output for compression via an attribute

### Motivation
Compression can be counter-productive on small inputs both in terms of output size and CPU time.

### References
Issue #1328

### DIFF
--- a/lib/Mojolicious/Renderer.pm
+++ b/lib/Mojolicious/Renderer.pm
@@ -14,6 +14,7 @@ has [qw(compress default_handler)];
 has default_format => 'html';
 has encoding       => 'UTF-8';
 has [qw(handlers helpers)] => sub { {} };
+has min_compress_size => '860';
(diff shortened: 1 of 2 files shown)


lib/Mojolicious/Renderer.pm
```diff
@@ -14,6 +14,7 @@ has [qw(compress default_handler)];
 has default_format => 'html';
 has encoding       => 'UTF-8';
 has [qw(handlers helpers)] => sub { {} };
+has min_compress_size => '860';
```
srchulo: nit: small thing, but why make this a string instead of an int? When comparing in the `>=` below, Perl will then make a number scalar value in addition to the string one. It's only three bytes, so not like it really matters, but why not create it as a number if that's how we intend to use it?

PR #1338, opened 2019-03-19
Silence "Ambiguous Scalar" Warnings

DabeDotCom:
### Summary
Silence `Warning: Use of "scalar" without parentheses is ambiguous at template path/to/template.html.ep line XXX` messages

### Motivation
I included a module that (like Mojolicious itself) enabled `use warnings` in its calling context.  Unfortunately, however, whenever that template included an `%= expression` (or `<%= expression %>`) it would issue a warning to STDERR.

I had originally looked into wrapping the "scalar + $value" like `scalar( + $value )`, and/or `scalar do { + $value }`, but that didn't work when the expression spanned multiple lines.  :-/

Then I tried adding just the `do {` to the opening statement, and the `}` before the semicolon, but *that* didn't work for, e.g., `%= $block->(--$i) if $i` — so I ended up just punting and prepending `no warnings 'ambiguous';` before the whole assignment.

"If it's stupid — but it works — it isn't stupid..."  ;-D

### DIFF
--- a/lib/Mojo/Template.pm
+++ b/lib/Mojo/Template.pm
@@ -213,11 +213,11 @@ sub _compile {
 
       # Escaped
       if (!$multi && ($op eq 'escp' && !$escape || $op eq 'expr' && $escape)) {
-        $blocks[-1] .= "\$_O .= _escape scalar + $value";
+        $blocks[-1] .= "no warnings 'ambiguous'; \$_O .= _escape scalar + $value";
       }
 
       # Raw
-      elsif (!$multi) { $blocks[-1] .= "\$_O .= scalar + $value" }
+      elsif (!$multi) { $blocks[-1] .= "no warnings 'ambiguous'; \$_O .= scalar + $value" }
 
       # Multi-line
       $multi = !$next || $next->[0] ne 'text';
--- a/t/mojo/template.t
+++ b/t/mojo/template.t
@@ -1189,4 +1189,12 @@ like $@, qr/invalid encoding/, 'right error';
 $mt = Mojo::Template->new(escape => sub { '+' . $_[0] });
 is $mt->render('<%== "hi" =%>'), '+hi', 'right escaped string';
 
+# No 'Use of "scalar" without parentheses is ambiguous' warnings
+$mt     = Mojo::Template->new;
+$output = $mt->render(<<'EOF');
+% use warnings FATAL => 'ambiguous';
+%= "ABC"
+EOF
+is $output, "ABC\n", 'no "ambiguous scalar" warning';
+
 done_testing();


Grinnz:
I don't think this is the best fix either, this warning is important for many instances which the user might run into themselves.

Grinnz:
What version of Perl are you seeing these warnings on?

DabeDotCom:
> this warning is important for many instances which the user might run into themselves

I do believe, however, that it's limited in scope to just the template, at least...  (Though there might be a better way to disable the warning once at the beginning of the whole `$_O` construction, rather than every time an expression is added.)

> What version of Perl are you seeing these warnings on?

I just tested it w/ Mojolicious 8.12 under Perl 5.24.0 (though I first noticed it under 5.18.)

`perl -MMojo::Template -E 'say Mojo::Template->new->render(q{<% use warnings; %><%= "ABC" %>})'`

jhthorsen:
Why can't we just do this?

```
@@ -213,11 +213,11 @@ sub _compile {
 
       # Escaped
       if (!$multi && ($op eq 'escp' && !$escape || $op eq 'expr' && $escape)) {
-        $blocks[-1] .= "\$_O .= _escape scalar + $value";
+        $blocks[-1] .= "\$_O .= _escape scalar(+ $value)";
       }
 
       # Raw
-      elsif (!$multi) { $blocks[-1] .= "\$_O .= scalar + $value" }
+      elsif (!$multi) { $blocks[-1] .= "\$_O .= scalar(+ $value)" }
```

DabeDotCom:
@jhthorsen  If *only* it were that easy!  ;-D

Like I said in my original message, there's no guarantee that the entire statement will be self-contained in a single <%= template %> expression...  It could just be part of a larger "begin/end" block, or it could have trailing conditionals, etc.

For example:

**`print Mojo::Template->new->render("<% use warnings; %>Two + Two = <%= 4 if 0 %>");`**

gets compiled to:

```
use warnings; $_O .= "Two\ \+\ Two\ \=\ ";$_O .= scalar +  4 if 0 ;$_O .= "\n";
```

But putting parentheses around `scalar(+ $value)` turns that into:

```
use warnings; $_O .= "Two\ \+\ Two\ \=\ ";$_O .= scalar(+  4 if 0 );$_O .= "\n";
```

which is a syntax error; the `if 0` needs to be OUTSIDE the parens.

Like I said, I also tried using, like, `do { $value }` and stuff like that, instead, but things still went haywire if the template spanned multiple lines:

```
<%= join " " =>
    $user->{'firstname'},
    $user->{'lastname'},
    $user->{'age'},
%>
```

etc.

Conveniently, `t/mojo/template.t` tests a whole slew of edge cases [Yay TDD!], and so far, my `no warnings 'ambiguous'` hack^H^H^H^H _solution_ is the only way I've found to satisfy them all...  :-}

kraih:
There is just no way we prepend `no warnings 'ambiguous';` to every expression.

jhthorsen:
@DabeDotCom: Sorry for my idiotic comment. I have no idea how I could run my suggested patch and _not_ notice the failing tests further up in my terminal 😞

kraih:
There's probably a smart way to generate better Perl with `Mojo::Template`.

marcusramberg:
I'm gonna go ahead and close this, since it seems to have gone stale, and will not be approved.

PR #1342, opened 2019-04-12
four forms instead of three since -role got added

elmar:
Small documentation fix.  This was overlooked when -role was added.


jberger:
thanks

PR #1343, opened 2019-04-24
return result of GetOptionsFromArray in getopt().

reneeb:
That way the users of getopt can check whether the option parsing was successful or not.

### Summary

Currently it's not possible to check e.g. in a ::Command whether the option parsing was successful or not. But if you want to exit as early as possible this is necessary.

### Motivation

GetOptionsFromArray already has such a return value but it is not returned in getopt() as the Getopt::Long::Configure call is the last statement in the subroutine.

### DIFF
--- a/lib/Mojo/Util.pm
+++ b/lib/Mojo/Util.pm
@@ -148,8 +148,10 @@ sub getopt {
   my ($array, $opts) = map { ref $_[0] eq 'ARRAY' ? shift : $_ } \@ARGV, [];
   my $save = Getopt::Long::Configure(qw(default no_auto_abbrev no_ignore_case),
     @$opts);
-  GetOptionsFromArray $array, @_;
+  my $result = GetOptionsFromArray $array, @_;
   Getopt::Long::Configure($save);
+
+  return $result;
 }
 
 sub gunzip {
--- a/t/mojo/util.t
+++ b/t/mojo/util.t
@@ -142,6 +142,26 @@ is_deeply $array, ['stuff'], 'right structure';
   is_deeply \@ARGV,    ['test'],   'right structure';
 }
 
+{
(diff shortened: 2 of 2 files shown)


jberger:
Overall I think it makes sense to return the result, meets with user expectations. For the tests, I wouldn't suggest testing the exact value but rather ok and ok not. Just check the boolean status.

kraih:
Agreed with @jberger. Also, what are the use cases for this feature?

jberger:
Since we are explicitly (ie, documented as) wrapping Getopt::Long I don't especially think we need to consider use cases when conforming to the expectations of the library we wrap. If we wanted to have some other documented behavior then I'd think we'd need use-cases.

glauschwuffel:
> Agreed with @jberger. Also, what are the use cases for this feature?

The use case is this: When I implement an additional Mojo command I want to check if argument parsing succeeded or not. If the user provided wrong arguments I want to fail as early as possible because continuing makes no sense (and might be actually harmful).

kraih:
I wanted to merge this PR for the next release, but looks like it hasn't been fixed yet.

reneeb:
I've replaced the is() checks with ok(). Anything else missing?

Tekki:
> Anything else missing?

Maybe an example in the docs?

reneeb:
Will add a paragraph or two to the docs today...

kraih:
No, the documentation is fine, but the tests need descriptions to be consistent.

jberger:
I've made a branch that takes these commits, adds one to normalize the tests, and rebases against master (as of this comment). https://github.com/mojolicious/mojo/tree/getopt_return

jberger:
I've merged #1351 which encompasses and obviates this PR.

PR #1348, opened 2019-05-02
Fall back to normal constructor if hash style arguments are provided

marcusramberg:
### Summary
Fall back to old behavior if more than one argument or hashref is given as argument

### Motivation
Maintain old behavior

### References
As discussed on IRC.

### DIFF
--- a/lib/Mojo/Promise.pm
+++ b/lib/Mojo/Promise.pm
@@ -2,6 +2,7 @@ package Mojo::Promise;
 use Mojo::Base -base;
 
 use Mojo::IOLoop;
+use Mojo::Util 'deprecated';
 use Scalar::Util 'blessed';
 
 has ioloop => sub { Mojo::IOLoop->singleton }, weak => 1;
@@ -67,6 +68,10 @@ sub map {
 }
 
 sub new {
+  if (@_ > 2 or ref($_[1]) eq 'HASH') {
+    deprecated 'Mojo::Promise::new with attributes is DEPRECATED';
+    return shift->SUPER::new(@_);
+  }
   my $self = shift->SUPER::new;
   shift->(sub { $self->resolve(@_) }, sub { $self->reject(@_) }) if @_;
   return $self;


marcusramberg:
As discussed on IRC we should probably either add a deprecation notice to this or change clone and the test back to use the constructor arguments instead of the setter. I prefer the latter.

kraih:
What are the use cases for this feature?

marcusramberg:
@kraih We do have a lot of constructors with special behavior for single argument. Like

``` lib/Mojo/URL.pm:55:1:sub new { @_ > 1 ? shift->SUPER::new->parse(@_) : shift->SUPER::new }```

**Comment by marcusramberg** [comment]
My main concern is maintaining backwards compat, so I'm fine with adding this with a deprecation notice as well to avoid breakage until upstreams have had time to adjust.

**Comment by kraih** [comment]
@marcusramberg That's not the same at all. There is no attribute handling in `Mojo::URL::new`, which is very special.

**Comment by Grinnz** [comment]
What's the status of this? Downstream modules are still broken in the meantime :(

**Comment by jhthorsen** [comment]
I'm 👍 on merging this, but maybe we should probably also add a deprecation warning at the same time if the `if` kicks in.

**Comment by Grinnz** [comment]
@marcusramberg Could you add a deprecation notice to the PR?

**Comment by marcusramberg** [comment]
@Grinnz done. @mojolicious/core can we have a couple of more votes?

File: lib/Mojo/Promise.pm [hunk 1]
```diff
@@ -3,6 +3,7 @@ use Mojo::Base -base;
 
 use Mojo::IOLoop;
 use Scalar::Util 'blessed';
+use Mojo::Util 'deprecated';
```
kraih: Out of order.

lib/Mojo/Promise.pm
```diff
@@ -67,6 +68,10 @@ sub map {
 }
 
 sub new {
+  if (@_ > 2 or ref($_[1]) eq 'HASH') {
+    deprecated 'Mojo::Promise::new with attributes is deprecated';
```
kraih: Inconsistent message.

PR #1359, opened 2019-06-03
Fix misplaced and missing paragraph in get command documentation

CandyAngel:
The documentation for the get command has a missing "attributes" paragraph and instead had another form of description. Stock attributes paragraph was added and second form of description was merged with the original description.


CandyAngel:
After finding the commit that introduced this error, I noticed my original PR wasn't achieving the goal of that commit. The description has now been replaced, instead of merged.

CandyAngel:
@mojolicious/core assemble! Your votes are required.

kraih:
👍

Issue #1350, opened 2019-05-10
Feature Request: Splice for Mojo::Collection


srchulo:
I think this could be a useful method. It would remove the elements from the existing collection, and return a new collection with those elements. I'd be happy to help with a pull request and tests if this seems like a good idea.

https://perldoc.perl.org/functions/splice.html

srchulo:
Currently I get around this like this:

`my $spliced_c = c(splice @$c, $offset, $length);`

s1037989:
There are times when I'd like to chain splice as well! :+1:

CandyAngel:
As you have to hold a reference to the original collection to use the leftovers, splice would have to always be the first method in the chain. For example, this is the same as ``slice`` as the flattened collection is lost and $original is unchanged.
```
my $original = c([0, 1], [0, 2]);
my $spliced = $original->flatten->splice(1, 2);
```
Unless you plan on using ``tap`` to store it, which could easily get lost in a long chain..
```
my $original = c([0, 1], [0, 2]);
my $spliced = $original->flatten->tap(sub { $original = $_ })->splice(1, 2);
```
Do you have an example where having splice available would make things clearer? Perhaps one where it isn't the first in the chain?

Creating a collection role for this (and your #1320 methods) is also an option to bear in mind if this isn't introduced to Mojo::Collection :)

jhthorsen:
I think splice() in a chain gets very awkward, so based on that I'm not very positive to adding this Mojo::Collection. I do wonder though if we could add a bit more logic to slice() so we don't get undef elements. The one-liner below prints "2" when I expect it to print "0":

```
perl -Mojo -le'my $c=c(1,2,3); print $c->slice(4, 5)->size'
```

I could be off topic here, but the code example @srchulo looks more like "slice" than "splice".

Grinnz:
I think that slice behavior is consistent with list slicing. @foo[4,5] will also return two undefs from an array without those indexes.

s1037989:
> I do wonder though if we could add a bit more logic to slice()

Speaking of slice(), I've often wanted the ability to slice the end, instead of the beginning.  For example, When I'm dealing with a large collection, I might want to do some testing and grab the first 100.

```
$coll->slice(0..99);
```

But sometimes, I need to focus more on the latter elements, so it'd be nice to slice the end.

```
$coll->slice(-99..-0);
```

Is there a way that slicing could grab to the end?  Perhaps by passing a callback if necessary to provide access to $_?

```
[...]

PR #1360, opened 2019-06-05
head and tail methods for Mojo::Collection

Grinnz:
### Summary
Methods for Mojo::Collection to create new collections of the beginning or end of the collection.

Algorithms for the methods were adapted from [List::Util::PP](https://metacpan.org/pod/List::Util::PP). We could just delegate to List::Util for XS functions but it would require List::Util 1.50.

### Motivation
It can be useful to operate on the beginning or end of the collection during a chain without calculating slice indexes, and resulting in no extra undef elements when going past the original collection indexes.

### References
https://github.com/mojolicious/mojo/issues/1350

### DIFF
--- a/lib/Mojo/Collection.pm
+++ b/lib/Mojo/Collection.pm
@@ -41,6 +41,13 @@ sub grep {
   return $self->new(grep { $_->$cb(@_) } @$self);
 }
 
+sub head {
+  my ($self, $size) = @_;
+  return $self->new(@$self) if $size > @$self;
+  return $self->new(@$self[0 .. ($size - 1)]) if $size >= 0;
+  return $self->new(@$self[0 .. ($#$self + $size)]);
+}
+
 sub join {
   Mojo::ByteStream->new(join $_[1] // '', map {"$_"} @{$_[0]});
 }
@@ -86,6 +93,13 @@ sub sort {
     $a->$cb($b);
   } @$self;
   return $self->new(@sorted);
+}
+
+sub tail {
+  my ($self, $size) = @_;
+  return $self->new(@$self) if $size > @$self;
+  return $self->new(@$self[($#$self - ($size - 1)) .. $#$self]) if $size >= 0;
+  return $self->new(@$self[(0 - $size) .. $#$self]);
 }
 
 sub tap { shift->Mojo::Base::tap(@_) }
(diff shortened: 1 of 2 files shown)


s1037989:
:+1: Love it!

kraih:
Like i said on IRC, it's pretty hard to justify adding so much code for a feature with such limited scope. Are there any good use cases related to `Mojo::DOM` or `Mojo::Pg`? Think i would be more ok with it if the methods were `List::Util` based one-liners and we were already depending on `List::Util` 1.50 for other reasons (like some of its other functions being used to improve `Mojo::Parameters` or similar).

jhthorsen:
I guess range() would be sort of a competing feature? https://github.com/mojolicious/mojo/commit/a24ae46f6bc5

Grinnz:
range and head/tail would accomplish largely the same things. I'm not sure I can say which is clearer or nicer to use.

marcusramberg:
I prefer #1361.

kraih:
I just did some tests, and the pair functions in `List::Util` are very very slow. So that option of bumping the dependency is ruled out.

kraih:
Since there have been no changes in some time, and the proposal is kinda complete. I'm just calling for a vote @mojolicious/core!

kraih:
The feature seems useful, but i don't think it's useful enough to require so much code. So, i'll be voting neutral with a slight tendency towards -1.

kraih:
Maybe worth mentioning that `Mojo::Collection::slice` is deprecated now. So there is no alternative at all in the `Mojo::Collection` API anymore.

jhthorsen:
Just as a reminder: https://github.com/mojolicious/mojo/pull/1361 was my alternative, but I closed it since there wasn’t any interest. I still think it’s a good idea to mimic JavaScript though.

jberger:
I am in favor of both this and/or #1361. Sorry I'm late to the vote, vacation got in the way

marcusramberg:
I think such a feature should exist, but guess I'm neutral to head/tail

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

kraih:
I've changed my vote to 👍, since there is a lot of interest in this functionality and i can't think of a more mojo-ish solution.

Grinnz:
Updated to require and delegate to List::Util 1.50.

PR #1361, opened 2019-06-05
Add range() to Mojo::Collection

jhthorsen:
### Summary
Method for Mojo::Collection to create new collections with the same API as `slice()` in JavaScript: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/slice

### Motivation
I often wanted `Mojo::Collection::slice()` to avoid returning `undef()` elements when specifying a range outside of the collection's size, and also allow specifying a range with negative indexes.

### References
This idea (finally) came to life from the discussions/PR below and conversations on IRC, earlier today.
https://github.com/mojolicious/mojo/issues/1350
https://github.com/mojolicious/mojo/pull/1360

### DIFF
--- a/lib/Mojo/Collection.pm
+++ b/lib/Mojo/Collection.pm
@@ -57,6 +57,16 @@ sub new {
   return bless [@_], ref $class || $class;
 }
 
+sub range {
+  my $self = shift;
+  my ($begin, $end)
+    = map { my $i = $_ < 0 ? @$self + $_ : $_; $i < 0 ? 0 : $i }
+    ($_[0] // 0, $_[1] // int @$self);
+  return $self->new if $begin >= @$self || $end == 0;
+  $end = @$self - 1 if $end >= @$self;
+  return $self->new(@$self[$begin .. $end]);
+}
+
 sub reduce {
   my $self = shift;
   @_ = (@_, @$self);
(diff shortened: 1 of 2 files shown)


marcusramberg:
This seems more versatile than head/tail, but it still lacks docs?

Grinnz:
Functionally this would be able to serve the same purpose as head and tail methods so I am in favor. But I prefer the head and tail methods as an API. The slice API this is based on is confusing to me, and not consistent with other Perl APIs.

jhthorsen:
We could make a similar API with `($offset, $len)`, but that would not allow you to do `(-4, -2)`, I think.

kraih:
I have to admit that i do find this API confusing too.

jhthorsen:
I also found the JavaScript API confusing in the beginning, but now I really like it.

Either way, I'm closing this PR because of lack of interest.

jhthorsen:
Based on the discussions on IRC, I'm reopening the PR.

jhthorsen:

kraih:
If you're going for the JavaScript compatibility angle the name needs to be `slice`.

Grinnz:
I am :+1: if #1360 does not succeed as it would then fill missing functionality. But compared to #1360 I am :-1: because the API is inconsistent with Perl.

kraih:
While i appreciate reusing common JavaScript APIs because most people are already familiar with them, i think #1360 is the more mojo-ish solution. 👎

kraih:
I believe the other proposal won over this.

PR #1370, opened 2019-06-20
Add -nst to .perltidyrc

Grinnz:
### Summary
The -nst option disables -st which is enabled by -pbp, which causes the tidied code to be written to STDOUT instead of a a new file suffixed with `.tdy`. This means only one file can be tidied at a time, and the -st option is also incompatible with the -b option, which allows updating files in place, a workflow well suited for git.

Instead of `perltidy lib/Some/File.pm > lib/SomeFile.pm.tdy`, now just do `perltidy lib/Some/File.pm` and multiple files can be specified in one command.

`perltidy -b lib/Some/File.pm` will update the file(s) in place and create lib/Some/File.pm.bak.

`perltidy -b -bext=/ lib/Some/File.pm` will update the file(s) in place and delete the backup file(s) as long as there were no errors.

### Motivation
To make it easier for contributors to run perltidy on their contributions.

### References
https://github.com/mojolicious/json-validator/pull/162

### DIFF
--- a/.perltidyrc
+++ b/.perltidyrc
@@ -11,3 +11,4 @@
 -sbt=2   # High square bracket tightness
 -wn      # Weld nested containers
 -isbc    # Don't indent comments without leading space
+-nst     # Don't output to STDOUT


Grinnz:

s1037989:
> To make it easier for contributors to run perltidy on their contributions.
I can say I've always been confused by how I need to run perltidy.  Granted, I don't use perltidy otherwise  :S
> perltidy -b lib/Some/File.pm will update the file(s) in place and create lib/Some/File.pm.bak.
It may be easy to do already, but I've not found explicit instructions for how to run perltidy for this project.  Even if this PR isn't accepted, just having a specific instruction on how specifically we should run perltidy would be helpful.  The simplicity above of how to use perltidy is great!

Grinnz:
I think it would be good to have a wiki page about invoking perltidy if nothing else.

Grinnz:
Note that -pbp also includes -se, which causes errors to be sent to STDERR instead of a filename.ERR file, but I don't think that impedes the workflows I described.

Grinnz:
I've updated/renamed the wiki page https://github.com/mojolicious/mojo/wiki/Perltidy to possibly be more helpful.

PR #1371, opened 2019-06-23
Update default SSL/TLS key to support TLS 1.2+

knowledgejunkie:
### Summary

This updates the default key used for local HTTPS server testing to
support TLS 1.2+. An RSA key of 4096 bits is created, with SHA256
digest support, with duration and Common Name as before.

### Motivation

As of Debian 10 "buster" the default system supported TLS level is 1.2,
and as a result, the current server.crt certificate using RSA:1024 with SHA1
digest support is not supported and will no longer work "out of the box".

With Apple, Google, Microsoft, and Mozilla announcing plans to deprecate
TLS 1.0 and 1.1 in early 2020, and Linux distributions updating their
baseline TLS support, it seems prudent to update this key so that it
continues to work for developers.

### References

https://groups.google.com/forum/#!topic/mojolicious/gjz-0uvUDLk
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=929675


jhthorsen:
I'm positive to updating the key to TLS 1.2+, but I'm negative to accept a new cert via a PR from someone who's not in the core team. Please correct me if I'm wrong in considering that a security concern.

kraih:
Agreed, someone from the team has to generate the new cert.

knowledgejunkie:
Ack. Would you like me to close this PR and open a regular issue (with the relevant tags)?

kraih:
@knowledgejunkie I suppose that would make sense, thanks.

kraih:
I've updated the certificate. https://github.com/mojolicious/mojo/compare/59b409c21343...2343209bdf7b

Issue #1295, opened 2018-11-26
Add proxy primitives


jberger:
When users use a Mojolicious app to proxy to and from some other resource, be it a [PSGI application mounted within](https://github.com/marcusramberg/Mojolicious-Plugin-MountPSGI/issues/8), streaming useragent responses to a web client directly, etc etc, the mechanisms available are not exactly matched to the task. 

We should provide a system that is like the nodejs pipe. This should be more that a method but rather a set of extension hooks that would allow piping between transactions.

kraih:
Regarding the API, on the highest level i think something like this might work.
```perl
get '/' => sub {
  my $c = shift;

  my $tx = $c->ua->build_tx(GET => 'http://mojolicious.org');
  $tx->res->content(body => sub {
    $c->res->headers->header('X-Hello' => 'Mojo!');
  });
  $tx->pipe($c->tx);
  $c->ua->start_p($tx)->then(sub {...});
};
```
Start-line line and headers would be rebuilt, but the response content can be piped right through.

jhthorsen:
Maybe there's some ideas to steal from https://github.com/nodejitsu/node-http-proxy ?

jberger:
More context on the PSGI reference is an IRC snippet:

```
2018-11-19T14:13:37 0 <jberger> kraih: is there any way to get or keep the original body of a multi part request?
2018-11-19T14:14:12 0 <kraih> no                                                              
2018-11-19T14:14:34 0 <jberger> hmmmm                                                         
2018-11-19T14:14:45 0 <kraih> i mean, you could add an event somewhere and store the raw body i suppose
2018-11-19T14:15:05 0 <jberger> the MountPSGI plugin has a problem with form data it seems
2018-11-19T14:15:33 0 <kraih> $tx->req->content body and then read events maybe               
2018-11-19T14:15:48 0 <jberger> yeah, that might be the best thing available                  
2018-11-19T14:16:00 0 <kraih> this is kind of related to the proxy optimization i wanted  
2018-11-19T14:16:54 0 <kraih> a way to pipe the message body from one transaction to another
2018-11-19T14:17:32 0 <jberger> related but maybe not exactly, since this end of the pipe doesn't have a transaction but a psgi env
2018-11-19T14:17:55 0 <kraih> it could still use the same primitives                          
2018-11-19T14:18:09 0 <kraih> pipe wouldn't just be one method, but a set of extension hooks
[...]

PR #1373, opened 2019-06-27
Add streaming proxy feature

kraih:
This is an implementation of the streaming proxy described in #1295.

### DIFF
--- a/lib/Mojo/Headers.pm
+++ b/lib/Mojo/Headers.pm
@@ -31,6 +31,12 @@ for my $header (keys %NAMES) {
   };
 }
 
+# Hop-by-hop headers
+my @HOP_BY_HOP = map {lc} (
+  qw(Connection Keep-Alive Proxy-Authenticate Proxy-Authorization TE Trailer),
+  qw(Transfer-Encoding Upgrade)
+);
+
 sub add {
   my ($self, $name) = (shift, shift);
 
@@ -50,7 +56,22 @@ sub append {
   return $self->header($name => defined $old ? "$old, $value" : $value);
 }
 
-sub clone { $_[0]->new->from_hash($_[0]->to_hash(1)) }
+sub clone {
+  my $self = shift;
+
+  my $clone = $self->new;
+  %{$clone->{names}} = %{$self->{names} // {}};
+  @{$clone->{headers}{$_}} = @{$self->{headers}{$_}}
+    for keys %{$self->{headers}};
+
+  return $clone;
+}
+
+sub dehop {
+  my $self = shift;
+  delete @{$self->{headers}}{@HOP_BY_HOP};
+  return $self;
+}
 
 sub every_header { shift->{headers}{lc shift} || [] }
 
@@ -385,6 +406,13 @@ L<RFC 6265|http://tools.ietf.org/html/rfc6265>.
   $headers = $headers->date('Sun, 17 Aug 2008 16:27:35 GMT');
 
 Get or replace current header value, shortcut for the C<Date> header.
+
+=head2 dehop
+
+  $heders = $headers->dehop;
+
+Remove hop-by-hop headers that should not be retransmitted. Note that this
+method is EXPERIMENTAL and might change without warning!
 
 =head2 dnt
 
--- a/lib/Mojolicious/Plugin/DefaultHelpers.pm
+++ b/lib/Mojolicious/Plugin/DefaultHelpers.pm
@@ -6,9 +6,10 @@ use Mojo::ByteStream;
 use Mojo::Collection;
 use Mojo::Exception;
 use Mojo::IOLoop;
+use Mojo::Promise;
 use Mojo::Util qw(dumper hmac_sha1_sum steady_time);
 use Time::HiRes qw(gettimeofday tv_interval);
-use Scalar::Util 'blessed';
+use Scalar::Util qw(blessed weaken);
 
 sub register {
   my ($self, $app) = @_;
@@ -40,6 +41,10 @@ sub register {
 
   $app->helper(dumper  => sub { shift; dumper @_ });
   $app->helper(include => sub { shift->render_to_string(@_) });
+
+  $app->helper('proxy.get_p'  => sub { _proxy_method_p('GET',  @_) });
+  $app->helper('proxy.post_p' => sub { _proxy_method_p('POST', @_) });
+  $app->helper('proxy.start_p' => \&_proxy_start_p);
 
   $app->helper("reply.$_" => $self->can("_$_")) for qw(asset file static);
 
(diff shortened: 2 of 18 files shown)


kraih:
The core functionality seems pretty solid, and so do the new low level methods and attributes (naming could use some bikeshedding probably).

What i do not quite like yet is the API of the helper, but i can't think of a better alternative. I originally planned to use the returned promise for response rewriting (adding headers and so on). Unfortunately the promise would have to be resolved right in between the `body` and `read` events of the forwarded response, which is impossible because resolving a promise requires another reactor tick. So instead i had to use a callback passed as an argument to `start_p`, which looks rather silly.

The helper API might be a blocker for this feature.

kraih:
This script illustrates the problem.
```perl
use Mojo::Base -strict;

use Mojo::EventEmitter;
use Mojo::IOLoop;
use Mojo::Promise;
use Mojo::Util 'dumper';

my $promise = Mojo::Promise->new;
my $events  = Mojo::EventEmitter->new;

# We want to use a promise to change the "foo" value
my $response = {foo => 'before'};
$promise->then(sub {
  $response->{foo} = 'after';
});

# The first event resolves the promise, and the second prints the "foo" value
$events->on(
  one => sub {
    $promise->resolve();
  }
);
$events->on(
  two => sub {
    say dumper $response;
  }
);

# Events happen in the right order, but the promise only resolves after the
# second event because it requires a reactor tick
$events->emit('one')->emit('two');
Mojo::IOLoop->one_tick;
```

kraih:
Having mentioned the negative, on the plus side the streaming proxy has been up to 25 times faster than previously used proxy implementations for Mojolicious. And response size does not matter at all, it can even be an infinite EventSource stream.
```perl
use Mojolicious::Lite;

my $UPSTREAM = 'http://127.0.0.1:3000';

# Old way (caching the whole response)
get '/proxy1' => sub {
  my $c = shift;

  $c->ua->get_p($UPSTREAM)->then(sub {
    my $tx = shift;
    $c->tx->res($tx->res);
    $c->rendered;
  });
};

# New way (streaming the response as chunks of data arrive)
get '/proxy2' => sub {
  my $c = shift;

  $c->proxy->get_p($UPSTREAM);
};

app->start;

```
Throttling works automatically, so if one direction is faster than the other the proxy will not cache more than 1mb (this is configurable) of data.

kraih:
I'm calling for a vote on this @mojolicious/core. If there are reasonable suggestions for fixing the API i still have Friday to work on the feature.

kraih:
This PR also contains some small optimizations for streaming responses in general. Found those bottlenecks while benchmarking the proxy helper.

lib/Mojo/Headers.pm
```diff
@@ -50,7 +56,22 @@ sub append {
   return $self->header($name => defined $old ? "$old, $value" : $value);
 }
 
-sub clone { $_[0]->new->from_hash($_[0]->to_hash(1)) }
+sub clone {
```
jhthorsen: Is the new clone() method faster than the old one or is there another reason for the rewrite?

lib/Mojolicious/Plugin/DefaultHelpers.pm
```diff
@@ -41,6 +42,9 @@ sub register {
   $app->helper(dumper  => sub { shift; dumper @_ });
   $app->helper(include => sub { shift->render_to_string(@_) });
 
+  $app->helper('proxy.get_p'   => sub { _proxy_method_p('GET', @_) });
```
jhthorsen: Why not just do this?

```
$app->helper('proxy.get_p' => sub { _proxy_start_p($_[0], shift->ua->build_tx(get => @_)) };
```

lib/Mojolicious/Plugin/DefaultHelpers.pm
```diff
@@ -161,6 +165,61 @@ sub _is_fresh {
...
+
+      my $write = $source_content->is_chunked ? 'write_chunk' : 'write';
+      $source_content->unsubscribe('read')->on(
+        read => sub {
+          $content->$write(pop) and $tx->resume;
+
+          # Throttle transparently when backpressure rises
+          return if $stream->can_write;
+          $source_stream->stop;
+          $stream->once(drain => sub { $source_stream->start });
```
jhthorsen: I think this throttling code is really nice 👍

lib/Mojolicious/Plugin/DefaultHelpers.pm
```diff
@@ -161,6 +165,61 @@ sub _is_fresh {
...
+    body => sub {
+      my $source_content = shift;
+
+      my $source_res = $source_tx->res;
+      my $res        = $tx->res;
+      my $content    = $res->content;
+      $res->code($source_res->code)->message($source_res->message);
+      my $headers = $source_res->headers->clone->dehop;
+      $content->headers($headers);
+      $c->$cb($source_tx) if $cb;
```
jhthorsen: I feel it's a bit awkward to pass in a callback. Could we (ab)use events on the transaction instead? Like `$source_tx->emit('proxy_start')` or something?

lib/Mojolicious/Plugin/DefaultHelpers.pm
```diff
@@ -161,6 +165,61 @@ sub _is_fresh {
...
+      $promise->resolve();
```
jhthorsen: What is the reasoning behind resolving the promise at this point? I wonder if it would make more sense to resolve the promise when the whole transaction is done.

lib/Mojo/Headers.pm
kraih: More than twice as fast.

lib/Mojolicious/Plugin/DefaultHelpers.pm
kraih: Because i might want to add `proxy->post_p` before the PR is merged.

lib/Mojolicious/Plugin/DefaultHelpers.pm
kraih: That event would be specific to a Mojolicious level experimental helper? How do you describe that event in the documentation?

lib/Mojolicious/Plugin/DefaultHelpers.pm
kraih: Consistency, and so the reject below cannot succeed.

lib/Mojolicious/Plugin/DefaultHelpers.pm
kraih: I like it too, but i wish `once` handlers were less expensive.

lib/Mojolicious/Plugin/DefaultHelpers.pm
jhthorsen: I thought about the same (issue about documentation), but I still think adding the specific documentation together with the helper is less strange than the callback. I also wonder if adding more hooks/events in the future could make more sense.

kraih: Yes i agree, if we can find a good justification for the event it would be much less awkward. But we have to find a justification that's not a high level helper.

kraih: The same way making promises actually work for this would be less awkward. But we can't, so here we are with the callback being the best choice at the moment.

jhthorsen: I'm glad you managed to get rid of $cb 👍

kraih:
Ok, the problem with the API has been resolved, this should be fine now. 😀

kraih:
The solution we settled on for the API problem was to just reuse the `body` event after the `start_p` call.
```perl
my $tx = $c->ua->build_tx(GET => 'http://mojolicious.org');
$c->proxy->start_p($tx);
$tx->res->content->once(body => sub {
  $c->res->headers->header('X-Proxy' => 'Mojo');
});
```
This gives us all the flexibility we need for rewriting HTTP responses.

kraih:
Since there is a lot of potential for bugs, i suggest we keep the feature experimental for some time.

jberger:
Looks good to me 👍 .

For posterity I want to leave a note that reminds us to document the timing (and values if any) of the resolution/rejection of the promise. For now, that can be deferred.

PR #1374, opened 2019-06-29
fix usage of setsid

skaji:
### Summary

Fix usage of setsid

### Motivation

POSIX::setsid returns -1, not undef on failure:
```
❯ perl -MPOSIX -le 'my $ret = POSIX::setsid; print "$ret: $!"'
-1: Operation not permitted
```

### References

N/A


kraih:
Thanks, applied.

PR #1377, opened 2019-07-08
Can string match on object as well

jhthorsen:
### Summary
This change allow `check()` in Mojo::Exception to run regex checks on objects that can stringify. 

### Motivation
@kraih asked for it in #mojo on freenode, and I think it makes a lot of sense to be able check if for example a Mojo::Exception object match certain regexes, since it's a very generic exception object.

I chose to check if the object is string overloaded, so we don't match against strings that look like "Foo::Bar=HASH(0x7f8ab5804248)".

### DIFF
--- a/lib/Mojo/Exception.pm
+++ b/lib/Mojo/Exception.pm
@@ -26,6 +26,7 @@ sub check {
 
   my ($default, $handler);
   my $is_obj = blessed $err;
+  my $can_re = !$is_obj || $is_obj && overload::Method($err, '""');
 CHECK: for (my $i = 0; $i < @spec; $i += 2) {
     my ($checks, $cb) = @spec[$i, $i + 1];
 
@@ -33,8 +34,8 @@ CHECK: for (my $i = 0; $i < @spec; $i += 2) {
 
     for my $c (ref $checks eq 'ARRAY' ? @$checks : $checks) {
       my $is_re = ref $c eq 'Regexp';
-      ($handler = $cb) and last CHECK if $is_obj  && !$is_re && $err->isa($c);
-      ($handler = $cb) and last CHECK if !$is_obj && $is_re  && $err =~ $c;
+      ($handler = $cb) and last CHECK if $is_obj && !$is_re && $err->isa($c);
+      ($handler = $cb) and last CHECK if $can_re && $err =~ $c;
     }
   }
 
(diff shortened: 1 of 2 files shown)


Grinnz:
I would not bother with the overload check. This will miss implied overloads from other operators.

jhthorsen:
@mojolicious/core: Calling for a vote.

jhthorsen:
> I would not bother with the overload check. This will miss implied overloads from other operators.

That is correct, but I'm guessing it's not a very common case.

kraih:
@Grinnz raises a good point. I keep going back and forth, but am really not sure which is better.

kraih:
Think we might actually have agreement on not checking overloads being the better choice.

lib/Mojo/Exception.pm
```diff
@@ -26,15 +26,16 @@ sub check {
...
   my ($default, $handler);
   my $is_obj = blessed $err;
+  my $can_re = !$is_obj || $is_obj && overload::Method($err, '""');
 CHECK: for (my $i = 0; $i < @spec; $i += 2) {
     my ($checks, $cb) = @spec[$i, $i + 1];
 
     ($default = $cb) and next if $checks eq 'default';
 
     for my $c (ref $checks eq 'ARRAY' ? @$checks : $checks) {
       my $is_re = ref $c eq 'Regexp';
```
karenetheridge: The proper way to check for regexes, that will work on all perl versions, is `re->can('is_regexp') ? re::is_regexp($_[0]) : ref($_[0]) eq 'Regexp'`

kraih: We don't have to support all Perl versions.

lib/Mojo/Exception.pm
```diff
@@ -26,15 +26,16 @@ sub check {
 
   my ($default, $handler);
   my $is_obj = blessed $err;
+  my $can_re = !$is_obj || $is_obj && overload::Method($err, '""');
```
kraih: Isn't that `$is_obj &&` redundant?

kraih:
I know a few people from the core team have argued about this, and it seems nobody is comfortable voting on it at all. To me that's a vote against and i consider this proposal failed. We'll do it without the overload check.

lib/Mojo/Exception.pm
karenetheridge: re::is_regexp became available in perl 5.9.5, so that would be preferrable to checking ref eq 'Regexp' in all cases then.

PR #1378, opened 2019-07-09
use re::is_regexp to detect regex refs

Grinnz:
### Summary
Use [is_regexp](https://perldoc.pl/re#is_regexp($ref)) to detect regex refs instead of checking for ref 'Regexp'.

### Motivation
Regex refs created by qr// are actually blessed into the 'Regexp' class. ref 'Regexp' can be fooled by blessing random things into 'Regexp', or by reblessing the regex ref into something else. is_regexp will always detect the underlying regex ref.

### References
https://github.com/mojolicious/mojo/pull/1377

### DIFF
--- a/lib/Mojo/Collection.pm
+++ b/lib/Mojo/Collection.pm
@@ -1,6 +1,7 @@
 package Mojo::Collection;
 use Mojo::Base -strict;
 
+use re 'is_regexp';
 use Carp 'croak';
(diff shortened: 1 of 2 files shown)


kraih:
Since @Grinnz forgot, i will call for a vote @mojolicious/core!

jberger:
:+1:

PR #1380, opened 2019-07-09
add private packages to no_index metadata

Grinnz:
### Summary
Prevent private packages from being indexed.

### Motivation
Packages starting with _ are not intended to be externally used or depended on, so they have no reason to be indexed on CPAN.

An alternative method is to put a newline after the 'package' keyword so PAUSE will ignore these packages, but this requires an exception for perltidy.

### DIFF
--- a/Makefile.PL
+++ b/Makefile.PL
@@ -18,9 +18,12 @@ WriteMakefile(
   META_MERGE   => {
     dynamic_config => 0,
     'meta-spec'    => {version => 2},
-    no_index       => {directory => ['examples', 't']},
-    prereqs        => {runtime => {requires => {perl => '5.010001'}}},
-    resources      => {
+    no_index       => {
+      directory => [qw(examples t)],
+      package   => [qw(Mojo::Exception::_Guard Mojo::Server::PSGI::_IO)],
+    },
+    prereqs   => {runtime => {requires => {perl => '5.010001'}}},
+    resources => {
       bugtracker => {web => 'https://github.com/mojolicious/mojo/issues'},
       homepage   => 'https://mojolicious.org',
       license    => ['http://www.opensource.org/licenses/artistic-license-2.0'],


Grinnz:
Calling for a vote from @mojolicious/core to gauge interest.

kraih:
I think this is the most indifferent i have ever felt about a patch. 😐

christopherraa:
Askin just so I understand: is this change proposed because CPAN can (or won't?) skip indexing `_`-prefixed packages?

Grinnz:
Correct, PAUSE currently does not care about _ prefixes on package names.

Grinnz:
The alternative approach does seem workable with perltidy, so it is opened at https://github.com/mojolicious/mojo/pull/1381

kraih:
After #1381, i suddenly feel much less indifferent about this one. 🤣

christopherraa:
From what I have understood so far the situation is this:

- it is assumed but not *enforced* that `_`-prefixed packages should not be indexed
- it does not present problems that they do get indexed
- the fix is either cosmetically dubious code, changes that requires perltidy-changes or the Makefile.PL-change

Given the above I would rather see the changes in the Makefile.PL if I had to choose, but overall I am :-1: to this change as I do not see **why** indexing `_`-prefixed packages is a problem. Feel free to enlighten me. :)

Grinnz:
It's not an explicit problem. It (being indexed on CPAN) could be considered an implicit indication that these packages can be used or depended on, which is not the case.

genio:
If you have any secondary packages within an actual module, you do not want to index those modules as you would not be able to use them directly from any application. So, it's best practice to hide inner packages/classes from PAUSE so that people don't try to do the wrong thing.

```perl
package Foo;
use strict;

{
  package Foo::Inner;
  use strict;
  sub whatever { return 42 }
}

sub new {
  my $self = bless {}, shift;
  $self->{_something} = "bar";
  return $self;
}

sub foo { return Foo::Inner::whatever }
1;
```

Now, in a script, I should be able to find `Foo` but not `FooInner`.

```perl
#!/usr/bin/env perl
use strict;
use warnings;
use Foo;
my $foo = Foo->new();
print $foo->foo; # yay
```

```perl
#!/usr/bin/env perl
use strict;
use warnings;
use Foo::Inner; # kaboom

print Foo::Inner::whatever;
```

Since PAUSE will index that by default, it will show up as something shipped in your distribution. You should hide these from PAUSE so people aren't going to try that second script example.

```perl
package # hide from PAUSE
  Foo::Inner;
```

Grinnz:
Note that #1381 will additionally hide the packages from the metacpan release page, while this PR will not.

christopherraa:
Thank you for a good example. That helped my understanding. Considering the example I'm neutral to this change as this does not require changes to code like shown in #1381 .

PR #1381, opened 2019-07-09
hide private packages from PAUSE indexer

Grinnz:
### Summary
Prevent private packages from being indexed. Alternative to #1380

### Motivation
Packages starting with _ are not intended to be externally used or depended on, so they have no reason to be indexed on CPAN.


kraih:
I prefer #1380 to this, even if it is the standard solution.

christopherraa:
I too would prefer #1380 , ref my last comment there.

Issue #1382, opened 2019-07-11
Feature Request : Adding method submit_form to mojo::useragent


b03cmans:
Feature Request : Adding method submit_form to Mojo::UserAgent 

like this method on https://metacpan.org/pod/WWW::Mechanize#$mech-%3Esubmit_form(-...-)  on WWW::Mechanize

which let you select a form from the previously fetched page, fill in its fields, and submit it. It combines the form_number/form_name/form_id, set_fields/with_fields and click methods into one higher level call .

kraih:
Unfortunately that can't work, since `Mojo::UserAgent` does not contain request state. Only the returned transaction object does.

b03cmans:
I see , is it possible to add such method to parse the result of Mojo::Message::Response and return a new request object which can be again  using Mojo::UserAgent ?

kraih:
Yes, that's more possible. On the `Mojo::DOM` layer you would teach [Mojo::DOM::val](https://mojolicious.org/perldoc/Mojo/DOM#val) to extract full forms (or add a new related method). And then probably add a method to `Mojo::Transaction::HTTP` that uses it to create a followup transaction.

kraih:
We've been trying to do the `Mojo::DOM` part in the past, but never found a good enough solution.

Tekki:
Would be great if we could replace LWP::UserAgent + HTML::Form with this.

Tekki:
@b03cmans + @kiwiroy, what do you think of that?
```
$dom->form('#form1')->val(%newvalues)->submit($ua, '#button1');
```
or
```
$form = FormClass->new($dom->at('#form1'))->val(%newvalues);
$ua->build_tx($form->submit('#button1'));
```

b03cmans:
@Tekki 

both looks good , supposing FormClass is repesenting the HTML Form elements something like HTML::Form ,it will be very nice .

kiwiroy:
I'd imagined
```
# Mojo::DOM::HTML::decode_form
$ua->build_tx(decode_form $dom->at('form'), $newvalues, '#button1');
```

Tekki:
The closer I look at it the more complicated it gets. First I wanted to propose an independent module. But this will not work. Just an example: We can't create the parameters for `build_tx` from a DOM because it doesn't contain the request URL.
If we have a FormClass module it should be built like this:
```
$form = $ua->get($url)->result->form('#form1');
```
Similar to `->result->json`.
[...]

PR #1383, opened 2019-07-15
Teach Mojo::DOM->val to extract full forms

kiwiroy:
### Summary
With this PR, `Mojo::DOM->val` can extract the `name` => `value` pairs from a `<form>` and return a hash reference of those. The attributes of the `<form>` are not returned, that might be another method's role.

### Motivation
Hopefully, this will help with #1382, and is a generally useful feature.

### References
* #1382

### DIFF
--- a/lib/Mojo/DOM.pm
+++ b/lib/Mojo/DOM.pm
@@ -178,9 +178,26 @@ sub type { shift->tree->[0] }
 
 sub val {
   my $self = shift;
+  # "form"
+  return {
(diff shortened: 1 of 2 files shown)


kraih:
Unfortunately this is wrong, you need to simulate a form submit to extract values correctly.

kiwiroy:
Indeed, but that seems out of scope for a Minimalistic DOM parser.

kiwiroy:
closing... see #1387

PR #1387, opened 2019-07-22
feature: form submission

kiwiroy:
### Summary
Adding the form submission codes. This includes `Mojo::DOM->val` updates, adding `Mojo::DOM->target` and `Mojo::Transaction::HTTP->submit` similar to discussion here and IRC.

```perl
my $ua = Mojo::UserAgent->new;
my $tx = $ua->get('/');
my $submit_tx = $tx->submit('#submit-form', username => 'fry', ...);
$ua->start($submit_tx);
```

This is a draft at the moment.
### Motivation
This seems like a generally useful thing to have

### References
* #1382 
* #1383

### DIFF
--- a/lib/Mojo/DOM.pm
+++ b/lib/Mojo/DOM.pm
@@ -166,6 +166,18 @@ sub tag {
   return $self;
 }
 
+sub target {
+  my ($self, $submit) = @_;
+  return () if ($self->tag // '') ne 'form';
+  return () unless defined($submit = $self->at($submit || ''));
+  return () if $submit->matches('[disabled]');
+  my $method = uc($submit->attr('formmethod') || $self->attr('method') || 'GET');
+  my $action = $submit->attr('formaction') || $self->attr('action') || '#';
+  my $enctyp = $submit->attr('formenctype') || $self->attr('enctype') ||
+    'url-encoded';
+  return $method, $action, $enctyp;
+}
+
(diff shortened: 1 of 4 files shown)


lib/Mojo/DOM.pm
```diff
@@ -166,6 +166,18 @@ sub tag {
   return $self;
 }
 
+sub target {
+  my ($self, $submit) = @_;
+  return () if ($self->tag // '') ne 'form';
+  return () unless defined($submit = $self->at($submit || ''));
+  return () if $submit->matches('[disabled]');
+  my $method = uc($submit->attr('formmethod') || $self->attr('method') || 'GET');
```
kberov: Looking at [the form tag attributes](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form#Attributes), I don't see such like `formmethod` or `formaction`. What is the rationale behind this check?

kiwiroy: See the docs for [button](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button)

kraih:
Am i misremembering, or did this patch end up as a role on CPAN?

kiwiroy:
You remember [correctly](https://metacpan.org/pod/Mojo::Transaction::HTTP::Role::Mechanize).

PR #1389, opened 2019-07-26
Brotli compression

ankitpati:
### Summary
Add support for [`IO::Compress::Brotli`](https://metacpan.org/pod/IO::Compress::Brotli "MetaCPAN") to [`Mojolicious::Renderer`](https://mojolicious.org/perldoc/Mojolicious/Renderer "Mojolicious.org").

### Motivation
[`Mojolicious::Renderer`](https://mojolicious.org/perldoc/Mojolicious/Renderer "Mojolicious.org") already supports [`gzip`](https://mojolicious.org/perldoc/Mojolicious/Renderer#compress "Mojolicious.org"). [Brotli](https://en.wikipedia.org/wiki/Brotli "Wikipedia") is a newer, but mature compression algorithm, [developed by Google](https://github.com/google/brotli "GitHub"), and supported by [all major browsers](https://caniuse.com/#feat=brotli "Can I Use?"), and [command-line tools](https://daniel.haxx.se/blog/2017/11/29/curl-7-57-0-happiness "Curl").

Where possible, Brotli should be preferred over GZip, as it gives [latency reduction of around 37%](https://medium.com/oyotech/how-brotli-compression-gave-us-37-latency-improvement-14d41e50fee4 "OYO") in real-world usage. It does not harm clients who don’t speak Brotli because it is only a preferred format, with fallback to GZip and even uncompressed text always available.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,7 +30,7 @@ env:
 install:
   - cpanm -n Cpanel::JSON::XS EV IO::Socket::Socks IO::Socket::SSL
   - cpanm -n Net::DNS::Native Role::Tiny Test::Pod Test::Pod::Coverage
-  - cpanm -n IO::Compress::Brotli
+  - if test -n "$(perl -e 'print $] >= 5.014')"; then cpanm -n IO::Compress::Brotli; fi
   - cpanm -n --installdeps .
 sudo: false
 notifications:
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,6 +30,7 @@ env:
 install:
   - cpanm -n Cpanel::JSON::XS EV IO::Socket::Socks IO::Socket::SSL
   - cpanm -n Net::DNS::Native Role::Tiny Test::Pod Test::Pod::Coverage
+  - cpanm -n IO::Compress::Brotli
   - cpanm -n --installdeps .
 sudo: false
 notifications:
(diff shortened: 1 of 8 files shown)


kraih:
I don't think this is a good idea. Brotli is very slow, and we only compress dynamically generated content so far.

ankitpati:
`Mojolicious::Plugin::AssetPack` is only for static assets. This PR allows dynamic assets to be compressed through Brotli.

Also, Brotli compression being slow is largely a misconception at this time. Both Cloudflare and Google use it for dynamic content.

Please also read [this opinion](https://certsimple.com/blog/nginx-brotli "CertSimple"). Here is a more [balanced and detailed take](https://blog.cloudflare.com/results-experimenting-brotli/ "Cloudflare") from Cloudflare.

My opinion is that the decision should be left to the user.

jhthorsen:
👎 from me.
I think this should be released as a separate mojolicious plugin instead.

kraih:

marcusramberg:
```The way I see it, Brotli already has an advantage over zlib for large files (larger than 64KB) on slow connections. However, those constitute only 20% of our sampled dataset (and 80% of the total size).``` - Seems like this is an unlikely use case for most Mojo apps. 

Thus I'm 👎 on adding this to the core, but would love to see a plugin with Brotli support.

ankitpati:
Thank you for the consideration of this idea and suggestions for a plugin. I will get that plugin on CPAN soon. Please feel free to close this PR.

PR #1390, opened 2019-07-29
generate longer request_ids, to avoid duplication

karenetheridge:
I have been seeing duplication occur frequently when request_ids are truncated. Instead, keep the entire md5sum string.

### DIFF
--- a/lib/Mojo/Message/Request.pm
+++ b/lib/Mojo/Message/Request.pm
@@ -11,7 +11,7 @@ has env    => sub { {} };
 has method => 'GET';
 has [qw(proxy reverse_proxy)];
 has request_id => sub {
-  substr sha1_sum($SEED . ($COUNTER = ($COUNTER + 1) % 0xffffff)), 0, 8;
+  sha1_sum($SEED . ($COUNTER = ($COUNTER + 1) % 0xffffff));
 };
 has url       => sub { Mojo::URL->new };
 has via_proxy => 1;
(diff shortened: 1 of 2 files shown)


karenetheridge:
Also possible, something like: `substr b64_encode($SEED . ($COUNTER = ($COUNTER + 1) % 0xffffff)), 0, 12;`

I understand the goal of keeping the id short, but it cannot be so short that it loses realistic uniqueness.

christopherraa:
If I remember the initial discussions correctly the generation of the id was done mostly with performance in mind. Have you benchmarked your implementation and compared it with the original to see if there is any noticeable difference?

karenetheridge:
> Have you benchmarked your implementation

No, or I would have included such data here (may I suggest a more neutral phrasing: "could you benchmark your implementation..."), but I will do so.

kraih:
The short ids are all about being readable in debug logs during development.

jhthorsen:
I'm neutral to this change. When I use ID's I always copy paste them, so I'm not sure if I will care too much if they are long or short.

kraih:
Think i'd also like to see other ways to format long ids explored (like `ab07acbb....1e496801937adfa772424bf7` for example).

jhthorsen:
I'm giving this PR 👎 since I don't see why the ID can't be changed to whatever in the consuming application.

karenetheridge:
> I don't see why the ID can't be changed to whatever in the consuming application

That's exactly what I did.

PR #1393, opened 2019-07-31
Protect from undef values in Mojo::File objects

kraih:
It is currently very easy to delete the wrong directory by accident, like:
```perl
my $subdir = undef;
path('usr', 'share', $subdir)->remove_tree;
```
This would currently delete `usr/share` and only show a pretty insignificant warning about the `undef` value. I noticed the problem because we had a similar accident at work. 😉
```
$ perl -Mojo -E 'n { f("foo", "bar", "baz")->child("yada") } 1000000'
2.67053 wallclock secs ( 2.67 usr +  0.00 sys =  2.67 CPU) @ 374531.84/s (n=1000000)

$ perl -Ilib -Mojo -E 'n { f("foo", "bar", "baz")->child("yada") } 1000000'
2.80627 wallclock secs ( 2.81 usr +  0.00 sys =  2.81 CPU) @ 355871.89/s (n=1000000)
```
The performance cost is pretty minimal. I would have liked to include the empty string too, but unfortunately we depend on that currently for a [Mojo::Home hack](https://github.com/mojolicious/mojo/blob/master/lib/Mojo/Home.pm#L15) that i don't dare to change again.

### DIFF
--- a/lib/Mojo/File.pm
+++ b/lib/Mojo/File.pm
@@ -94,6 +94,7 @@ sub move_to {
 
 sub new {
   my $class = shift;
+  croak('Invalid path') if grep { !defined } @_;
   my $value = @_ == 1 ? $_[0] : @_ > 1 ? catfile @_ : canonpath getcwd;
   return bless \$value, ref $class || $class;
 }
--- a/t/mojo/file.t
+++ b/t/mojo/file.t
@@ -295,6 +295,12 @@ utime $future, $future, $file->to_string;
 is $file->stat->mtime, $future, 'right mtime';
 isnt $file->touch->stat->mtime, $future, 'different mtime';
 
+# Dangerous paths
+eval { path('foo', undef, 'bar') };
+like $@, qr/Invalid path/, 'right error';
+eval { path(undef) };
+like $@, qr/Invalid path/, 'right error';
+
 # I/O
 $dir  = tempdir;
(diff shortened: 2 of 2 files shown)


kraih:
I'm not 100% sure if it is worth the performance cost, but i'm giving it a :+1: personally and call for a vote anyway @mojolicious/core.

christopherraa:
If I had voting rights I'd be :+1: for this. I've fallen into this trap myself.

marcusramberg:
Im 👍 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

jberger:
:+1:

lib/Mojo/File.pm
```diff
@@ -24,7 +24,7 @@ our @EXPORT_OK = ('path', 'tempdir', 'tempfile');
 
 sub basename { File::Basename::basename ${shift()}, @_ }
 
-sub child { $_[0]->new(@_) }
+sub child { $_[0]->new(${shift()}, @_) }
```
jberger: I've already voted in favor but can you explain the significance of this line?

kraih: Totally insignificant. It's an optimization that's already in master actually. 😆

kraih: It makes `->child` calls *a lot* faster though since it avoids stringification later.

PR #1394, opened 2019-07-31
require List::Util 1.50 and delegate for head and tail

Grinnz:
### Summary
Delegate to the head and tail functions from List::Util 1.50.

Note that this raises the minimum Perl version that fatpacked Mojolicious will work on from 5.22 to 5.28, as List::Util cannot be fatpacked.

### Motivation
Code cleanliness and efficiency.

### References
#1360


Grinnz:
An alternative is to delegate only if List::Util 1.50 is available, which retains the performance benefit if any on newer perls, but adds more code.

Grinnz:
I am closing this as I cannot find a set of data where delegating to List::Util outperforms the resulting sub call overhead.

PR #1397, opened 2019-08-07
require Cpanel::JSON::XS 4.09+ for JSON speedup and set allow_dupkeys

Grinnz:
### Summary
Recent versions of Cpanel::JSON::XS throw an exception when decoding duplicate keys. allow_dupkeys was added in 4.09 to disable this behavior which is incompatible with Mojo::JSON.

### Motivation
Behavior compatibility with Cpanel::JSON::XS passthrough.

### DIFF
--- a/t/mojo/json_xs.t
+++ b/t/mojo/json_xs.t
@@ -4,7 +4,7 @@ use Test::More;
 use Mojo::JSON qw(decode_json encode_json false from_json j to_json true);
 
 BEGIN {
-  plan skip_all => 'Cpanel::JSON::XS 4.04+ required for this test!'
+  plan skip_all => 'Cpanel::JSON::XS 4.09+ required for this test!'
     unless Mojo::JSON->JSON_XS;
 }
 
@@ -71,5 +71,9 @@ like encode_json({test => -sin(9**9**9)}), qr/^{"test":".*"}$/,
 
 # "escape_slash"
 is_deeply encode_json('/test/123'), '"\/test\/123"', 'escaped slash';
+
+# "allow_dupkeys"
+is_deeply decode_json('{"test":1,"test":2}'), {test => 2},
+  'no duplicate keys error';
 
 done_testing();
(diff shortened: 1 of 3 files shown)


Grinnz:
@mojolicious/core Calling for a vote

t/mojo/json_xs.t
```diff
@@ -72,4 +72,9 @@ like encode_json({test => -sin(9**9**9)}), qr/^{"test":".*"}$/,
 # "escape_slash"
 is_deeply encode_json('/test/123'), '"\/test\/123"', 'escaped slash';
 
+# "allow_dupkeys"
+my $decoded;
+ok eval { $decoded = decode_json('{"test":1,"test":2}'); 1 }, 'no duplicate keys error';
```
kraih: Why `eval`?

Grinnz: The test failure condition (allow_dupkeys not being set) throws an exception.

kraih: Then why not let it throw? Think this is inconsistent with other tests.

kraih:
Normally i'd vote +1, but with the eval test it's only a neutral from me.

t/mojo/json_xs.t
Grinnz: Removed the eval

jberger:
What is the issue with an eval test? I don't see anything that jumps out to me as problematic but I'll withhold judgement until I hear from @kraih

marcusramberg:
👍 on this too.

christopherraa:
I'm :+1: on this. Seems like a sensible change.

Grinnz:
The eval test had been removed already. Merging.

kraih:
I'm afraid this vote has been invalidated by a rule change and needs to be redone. https://github.com/mojolicious/mojo/compare/41d1d1c6fa95...b4152fdcccf5

jberger:
👍

PR #1402, opened 2019-09-04
IOLoop::Stream warnings: Use of uninitialized value $self->{buffer}

jiyouyou125:
Use of uninitialized value in length at /xxxxxx/lib/perl5/lib/perl5/Mojo/IOLoop/Stream.pm line 43

### Summary

### Motivation
EXPLAIN WHY YOU BELIEVE THESE CHANGES ARE NECESSARY HERE

### References
LIST RELEVANT ISSUES, PULL REQUESTS AND IRC/MAILING-LIST DISCUSSIONS HERE

### DIFF
--- a/lib/Mojo/IOLoop/Stream.pm
+++ b/lib/Mojo/IOLoop/Stream.pm
@@ -40,7 +40,7 @@ sub is_readable {
 sub is_writing {
   my $self = shift;
   return undef unless $self->{handle};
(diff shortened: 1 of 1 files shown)


Grinnz:
Unfortunately Mojolicious does not optimize for 5.10, which is the only version of Perl these warnings occur on.

PR #1403, opened 2019-09-05
do not strengthen weak promise callback arguments

karenetheridge:
### Summary
Arguments to promise callback subs are weakened as they are copied, if the originals were weak.

### Motivation
Otherwise, memory leaks occur when defining callback subs that close over things like `$app`, `$ua` etc.

### References
Briefly discussed on \#mojo today.

### DIFF
--- a/lib/Mojo/Promise.pm
+++ b/lib/Mojo/Promise.pm
@@ -3,7 +3,7 @@ use Mojo::Base -base;
 
 use Mojo::IOLoop;
 use Mojo::Util 'deprecated';
-use Scalar::Util 'blessed';
+use Scalar::Util qw(blessed isweak weaken);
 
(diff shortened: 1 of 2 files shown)


karenetheridge:
My first attempt at a fix did not work -- there is more that needs to be done lower down. I have followed the code down into `$self->ioloop->next_tick(..)` and then more needs to be done in there as well.

jhthorsen:
Do you have a test that actually shows that the objects leak? There's many places in mojo where they appear to leak, but everything is cleaned up once the promise/delay is done doing what it's supposed to do.

kraih:
This pull request appears stale and broken, so i'm closing it instead of calling for a vote. Please feel free to resubmit a more complete patch.

karenetheridge:
This wasn't a patch proposal, but a demonstration of a problem, which @jberger asked me to submit.

Issue #1400, opened 2019-08-31
Logging with request id


kraih:
This has been a recurring topic for quite some time, and we've talked about this yesterday on IRC again. There are two logging scopes in Mojolicious, the first is operations that can be mapped to a request/response, and the second is everything outside the request/response cycle (such as a timer defined in the `startup` method).

For this feature we only care about the first case, any operation that can be mapped to the request/response cycle.
```perl
get '/' => sub ($c) {
  $c->log->debug('Some log message with request id!');
  $c->render(text => 'Hello Mojo!');
};
```
A log message created inside an action is the simplest case. In this context we want to get a log message that contains the request id (`$c->req->request_id`).
```
[2018-11-08 14:20:13.77168] [info] [1af234b] Some log message with request id!
```
But we also want to allow models to generate log messages with request id. One possible solution for that would be to make `$c->log` a secondary log object that contains the request id, and that can be passed to model objects/methods.
```perl
get '/' => sub ($c) {
  my $log = $c->log;
  my $foo = $c->param('foo');
  $c->some_model->create({foo => $foo}, $log);
  $c->render(text => 'Hello Mojo!');
};

...
package MyApp::Model::SomeModel;

sub create ($self, $values, $log) {
  $self->do_stuff($values);
  $log->debug('Another log message with request id!');
}
...
```
Of course this also means the secondary log object would work for non-blocking operations started inside the request/response cycle.
```perl
get '/' => sub ($c) {
  my $log = $c->log;
  Mojo::IOLoop->timer(1 => sub {
    $log->debug('Yet another log message with request id!');
  });
  $c->render(text => 'Hello Mojo!');
};
```
Logging in models is a problem that most of us are facing in web applications at some point. And i think the framework should provide at least one reasonable solution out of the box.

Thoughts?

christopherraa:
Very well summarized! This _exact_ usecase (logging in models etc) is a recurring need. Currently all my implementations of logging with request id's involve some form of unattractive hack to get a request id tied to the model-operations. I do not have any implementations to propose, but I will be following this issue closely.

s1037989:
This would be such a huge improvement!  Like @christopherraa, my implementations of logging are unattractive hacks.  It's the single most difficult part in regards to building a good model.

kraih:
I thought @jberger was working on this, but i guess he might have given up.

kraih:
[...]

PR #1404, opened 2019-09-09
Contextual logging

kraih:
This is one possible solution for #1400. The main goals were performance and simplicity. And there would be no backwards compatibility breakage.

### DIFF
--- a/lib/Mojo/Log.pm
+++ b/lib/Mojo/Log.pm
@@ -38,6 +38,13 @@ sub append {
 }
 
 sub debug { 1 >= $LEVEL{$_[0]->level} ? _log(@_, 'debug') : $_[0] }
+
+sub context {
+  my ($self, $context) = (shift, shift);
+  return $self->new(parent => $self, context => $context,
+    level => $self->level);
+}
+
 sub error { 4 >= $LEVEL{$_[0]->level} ? _log(@_, 'error') : $_[0] }
 sub fatal { 5 >= $LEVEL{$_[0]->level} ? _log(@_, 'fatal') : $_[0] }
 sub info  { 2 >= $LEVEL{$_[0]->level} ? _log(@_, 'info')  : $_[0] }
@@ -60,7 +67,12 @@ sub _default {
   return "[$time] [$$] [$level] " . join "\n", @_, '';
 }
 
-sub _log { shift->emit('message', pop, ref $_[0] eq 'CODE' ? $_[0]() : @_) }
+sub _log {
+  my ($self, $level) = (shift, pop);
+  my @msgs = ref $_[0] eq 'CODE' ? $_[0]() : @_;
+  $msgs[0] = "$self->{context} $msgs[0]" if $self->{context};
+  ($self->{parent} || $self)->emit('message', $level, @msgs);
+}
 
 sub _message {
   my ($self, $level) = (shift, shift);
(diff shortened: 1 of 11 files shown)


kraih:
Not putting this up for a vote yet (will wait a few days), because some people on IRC have been working on alternative proposals.

kraih:
I've also updated Mojolicious default log messages to include the request id when possible. Performance cost seems pretty minimal, less than 2% difference for `examples/hello-template.pl`.

kraih:
Since it now looks like nobody else is working on an alternative proposal, i'll just call for a vote @mojolicious/core.

kraih:
@jberger But if we are already planning to revisit structured logging, shouldn't the `context` method be designed around that? Like `$log->context('foo', 'bar')` instead of the current `$log->context('[foo] [bar]')`?

jberger:
For now, I think assuming a string in the context is fine. The way I see it, later any scalar could be used, including a data structure, and we'd have some additional formatter that would handle structured data when given. That's just my thought though

kraih:
That is true, `$log->context(['foo', 'bar'])` would work just as well later on.

lib/Mojolicious/Plugin/DefaultHelpers.pm
```diff
@@ -42,6 +42,14 @@ sub register {
   $app->helper(dumper  => sub { shift; dumper @_ });
   $app->helper(include => sub { shift->render_to_string(@_) });
 
+  $app->helper(
+    log => sub {
+      my $c = shift;
+      return $c->stash->{'mojo.log'}
+        ||= $c->app->log->context('[' . $c->req->request_id . ']');
+    }
+  );
```
jhthorsen: Does this mean that if I want my log messages to have a different context formatting, then I need to redefine the whole helper? (As in, if I want to remove the brackets or something) Could we change the log helper to take a $log object to set the value of `mojo.log`, or will that make the helper too slow?
(Not a big issue to me btw, since I think the default makes sense)

lib/Mojo/Log.pm
```diff
@@ -60,7 +67,12 @@ sub _default {
   return "[$time] [$$] [$level] " . join "\n", @_, '';
 }
 
-sub _log { shift->emit('message', pop, ref $_[0] eq 'CODE' ? $_[0]() : @_) }
+sub _log {
+  my ($self, $level) = (shift, pop);
+  my @msgs = ref $_[0] eq 'CODE' ? $_[0]() : @_;
+  $msgs[0] = "$self->{context} $msgs[0]" if $self->{context};
```
jhthorsen: Why isn't this done inside _default() and _short()? I was thinking if context() could be a getter as well, then custom log format() could format it any way they wanted.

lib/Mojolicious/Plugin/DefaultHelpers.pm
kraih: Yes, just redefined the helper, that's what helpers are for after all. For most people the default will be just fine anyway.

kraih: Or even add more log helpers with different contexts.

lib/Mojo/Log.pm
kraih: This has been discussed in great detail on IRC, it's unfortunate you did not participate. There is no clean way to do that without breaking backwards compatibility. We might do it in the next major release.

jhthorsen: Since there's log public logs from IRC, then I thought it might be worth mentioning it here as well. Anyways, I'll resolve this comment as well.

lib/Mojo/Log.pm
```diff
@@ -199,6 +211,20 @@ following new ones.
...
+  my $new = $log->context('[extra] [information]');
+
+Construct a new child L<Mojo::Log> object that will include context information
+with every log message. Note that this method is B<EXPERIMENTAL> and might
+change without warning!
+
+  # Log with context
+  my $log = Mojo::Log->new;
+  my $context = $log->context('[17a60115]');
+  $context->debug('This is a log message with context informtation');
```
marcusramberg: informtation => information

kraih:
This is just a test notification, please ignore @mojolicious/core.

PR #1407, opened 2019-09-16
set userinfo using Mojo::URL in Mojo::UserAgent Basic auth example

Grinnz:
### Summary
Basic auth commonly uses username and password values from external sources, so they should be set using the Mojo::URL userinfo method rather than directly in the URL where they need to be escaped.

### Motivation
More correct example for a common use case

### References
IRC discussion

### DIFF
--- a/lib/Mojo/UserAgent.pm
+++ b/lib/Mojo/UserAgent.pm
@@ -385,7 +385,8 @@ Mojo::UserAgent - Non-blocking I/O HTTP and WebSocket user agent
   my $tx = $ua->put('[::1]:3000' => {'Content-Type' => 'text/plain'} => 'Hi!');
 
   # Quick JSON API request with Basic authentication
-  my $value = $ua->get('https://sri:t3st@example.com/test.json')->result->json;
+  my $url = Mojo::URL->new('https://example.com/test.json')->userinfo('sri:☃');
+  my $value = $ua->get($url)->result->json;
 
   # JSON POST (application/json) with TLS certificate authentication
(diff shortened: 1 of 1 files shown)


Grinnz:
@mojolicious/core Calling for a vote

jberger:
I don't know that I agree with removing the first style, I could see adding the other style too. Might also want to show url escaped example

Grinnz:
I considered that but it seems too busy for what is a tiny segment of an already pretty full synopsis. I think it's best to only show the best way that covers all situations.

lib/Mojo/UserAgent.pm
```diff
@@ -385,7 +385,9 @@ Mojo::UserAgent - Non-blocking I/O HTTP and WebSocket user agent
   my $tx = $ua->put('[::1]:3000' => {'Content-Type' => 'text/plain'} => 'Hi!');
 
   # Quick JSON API request with Basic authentication
-  my $value = $ua->get('https://sri:t3st@example.com/test.json')->result->json;
+  use Mojo::URL;
```
kraih: I don't like that this example contains a `use` line. That doesn't feel right for a synopsis.

Grinnz: Would you prefer it just be omitted? (The example for the prepare event does this already)

Grinnz: Removed the use line

Grinnz:
Restarting the vote @mojolicious/core

PR #1408, opened 2019-09-16
Add curfile constructor for Mojo::File and recommend its usage

Grinnz:
### Summary
The curfile constructor for Mojo::File allows one to succinctly retrieve the cleaned-up absolute file path to the current source file.

### Motivation
The mechanism of curfile has a number of advantages over what has previously been used:
* It is shorter and easier to remember than writing `path(__FILE__)->realpath`.
* Calling realpath ensures that it is both an absolute path and resolves any symlinks. Many use cases, especially when adding to `@INC`, should always use the absolute path. Resolving symlinks may be important if the file is loaded via a symlink but references files next to its real path. This shortcut makes it less likely that this step will be forgotten.
* It is less prone to error than FindBin, as it does not depend on what script was executed initially or proper usage of FindBin->again, and is not affected by the presence of executables in PATH or bugs in FindBin shipped with old Perls.

Additionally, the documented recommendation for adding the project-local lib directory to `@INC` has been changed to use `lib`, since this is a clearer equivalent to `unshift @INC`, plus includes any arch or version specific directories if present. Explicit stringification is needed here as blessed references in `@INC` have a special meaning.

### References
Discussion in IRC

### DIFF
--- a/t/mojo/file.t
+++ b/t/mojo/file.t
@@ -54,13 +54,13 @@ is path('file.t')->basename('.t'), basename('file.t', '.t'), 'same path';
 is path('file.t')->to_abs->dirname, scalar dirname(rel2abs 'file.t'),
   'same path';
 
-# Checks
-ok path(__FILE__)->to_abs->is_abs, 'path is absolute';
-ok !path('file.t')->is_abs, 'path is not absolute';
-
 # Current file
 ok curfile->is_abs, 'path is absolute';
 is curfile, realpath(__FILE__), 'same path';
+
+# Checks
+ok path(__FILE__)->to_abs->is_abs, 'path is absolute';
+ok !path('file.t')->is_abs, 'path is not absolute';
 
 # Temporary directory
 my $dir  = tempdir;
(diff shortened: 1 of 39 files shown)


Grinnz:
"curfile" is mostly to be both short and consistent with "tempfile"/"tempdir", but please suggest alternate names. Calling for a vote @mojolicious/core

t/mojo/file.t
```diff
@@ -54,6 +54,10 @@ is path('file.t')->basename('.t'), basename('file.t', '.t'), 'same path';
 is path('file.t')->to_abs->dirname, scalar dirname(rel2abs 'file.t'),
   'same path';
 
+# Current file
+ok curfile->is_abs, 'path is absolute';
+is curfile, realpath(__FILE__), 'same path';
```
kraih: I have a feeling this test might cause portability issues at some point.

Grinnz: I can't think of any - realpath/abs_path only has portability issues for file paths that don't exist, and __FILE__ will always exist in this test; and curfile uses realpath itself so it will always match (the test is to ensure that it does, in fact).

jberger:
I too like some of the cleanups, but I also dislike the name. Classic bike shedding I fear. Are there really no other names?

Grinnz:
I haven't thought of any yet. The only precedent are things like `this` and `self` which have more object connotations and too ambiguous.

s1037989:
`thisfile`?

t/mojo/file.t
kraih: Windows Perl. 😬

jhthorsen:
I must say that `curfile` is growing on me... The only alternative I could come up with is `module_filename()`, which is sort of taken from `Module.filename` in JavaScript.

Grinnz:
This can be used in a script as easily as a module so module_filename doesn't seem sufficient.

t/mojo/file.t
Grinnz: Windows is what I was referring to regarding nonexistent paths unless there's another issue I'm not aware of?

PR #1411, opened 2019-09-25
Only create one Mojo::File object for curfile

Grinnz:
### Summary
Call Cwd::realpath directly, avoiding instantiating an intermediary Mojo::File object.

### Motivation
Optimization


Grinnz:

kraih:
Overruled. 😉

Issue #1405, opened 2019-09-10
:scope selector for Mojo::DOM


kraih:
We have an [almost finished branch](https://github.com/mojolicious/mojo/compare/scope_only) for [scope selectors](https://www.w3.org/TR/selectors-4/#the-scope-pseudo) already. There is just one problem, we have not yet found a way to make sibling combinators work with `:scope`.
```perl
# TODO: Make :scope work with sibling combinators
#is $dom->find('foo')->[1]->at(':scope ~ foo')->text, 'B', 'right result';
```
It's a bit of a brain teaser.

Eckankar:
I'll just relay my thoughts from IRC on here for posterity.

https://metacpan.org/release/Mojolicious/source/lib/Mojo/DOM/CSS.pm#L231 trims out the current subtree, but the answer to `':scope ~ foo'` doesn't reside inside of that.

It feels like the problem could be solved by deciding on the queue to search through on a group-by-group basis. If the group mentions `:scope` anywhere in the selector, search from the root. Otherwise, search the subtree as is currently done.

kraih:
@Eckankar Yes, that would seem reasonable.

Eckankar:
I'll take a look at if I can get something working to do this.

kraih:
The feature has been added.

PR #1414, opened 2019-10-03
Handle :scope cases where we need more than current subtree

Eckankar:
### Summary
This is an attempt at addressing #1405.

Currently, `$e->at` searches only inside the subtree belonging to `$e`. However, in cases like `':scope + p'`, we cannot do the search strictly in the this subtree, as the desired `p` resides outside of it.

To circumvent this problem, "absolutize" the selector (as the RFC calls it) if the selector is relative, and search from the root.

I do not absolutize the selector generated in `$e->matches`, as I'm not sure that makes sense.

### Motivation
Addresses #1405

### References
Addresses #1405


Eckankar:
I have rewritten the code to avoid doing a full tree search except when scope is explicitly invoked in the selector. This means we avoid incurring the performance penalty except in scoped cases.

I have also brought it up to the current master instead of the css_scope branch.

lib/Mojo/DOM/CSS.pm
```diff
@@ -173,18 +215,28 @@ sub _namespace {
   return !length $ns;
 }
 
+sub _parent {
+  my ($tree) = @_;
+  return $tree->[3] if $tree->[0] eq 'tag';
+  return $tree->[2] if $tree->[0] eq 'text';
+  return undef      if $tree->[0] eq 'root';
```
kraih: This conditional makes no sense.

kraih:
Test coverage still seems to be lacking.

kraih:
This PR seems to have gone stale. But i really like the approach, and it doesn't appear too hard to finish. So i'll probably just leave it open for now.

Eckankar:
Apologies for me being slow at getting back to the feedback - I only work on this while at work, and have been a bit sidetracked by other things.

Wrt. test coverage, I have tried to cover a few different reasonable looking common cases, as well as cases I in implementation felt might be likely to go wrong. Looking over the output of `Devel::Cover`, nothing in particular jumps out at me. (Which of course doesn't mean I haven't missed something.)

`_parent` isn't fully covered, but I'm not sure it can be without calling it manually. I decided to make it work regardless of the type of node it receives, even though it probably isn't called with all 3 kinds.

I can ask a colleague to look over my test-cases, if you'd like a second pair of eyes on them.

lib/Mojo/DOM/CSS.pm
Eckankar: I am unsure what you mean by this.

`_parent` is intended as a general utility function, that given a node traverses one step up in the syntax tree, and explicitly handles the 3 possible cases as I understand them.

kraih:
What specifically jumps out at me is the fact that you add `eq 'root'` checks, yet there is no `:root` selector in your tests. And i don't understand why `_parent` can't be fully covered with normal tests. If a user can't trigger a case with the public API then that is dead code.

lib/Mojo/DOM/CSS.pm
kraih: Take a closer look, what happens when `if $tree->[0] eq 'root'` doesn't match? 😉

Eckankar:
The function doesn't relate to the `:root` selector. Its main use is in implementing `_root` (which locates the root node by traversing up the tree), which is used to find the tree to search in when dealing with scoped selectors.

Technically, the `root` case is dead code, and I can remove the line, but then one has to be more careful when applying the function. I opted to implement the case, so the function can be called on any type of node, and produce a sensible result.

kraih:
We do not accept dead code.

Eckankar:
I have removed the case, and rebased to current master.

lib/Mojo/DOM/CSS.pm
```diff
@@ -173,18 +215,27 @@ sub _namespace {
   return !length $ns;
 }
 
+sub _parent {
+  my ($tree) = @_;
+  return $tree->[3] if $tree->[0] eq 'tag';
+  return $tree->[2] if $tree->[0] eq 'text';
```
kraih: You seem to have misunderstood me. Now this conditional doesn't make sense. There is no else case.

kraih: When the only option is `text` here, why do you check at all?

lib/Mojo/DOM/CSS.pm
```diff
@@ -19,19 +19,52 @@ my $ATTR_RE   = qr/
...
+  # Traverse through selector to see if it's absolute.
+  # https://drafts.csswg.org/selectors-4/#absolutizing
+  for my $part (@$selector) {
+    next unless ref($part);
+
+    for my $pc (@$part) {
+      next unless ref($pc) && $pc->[0] eq 'pc';
+
+      # If ":root" or ":scope" are mentioned, it's absolute.
+      return 1 if $pc->[1] eq 'root' || $pc->[1] eq 'scope';
```
kraih: I suspect this `:root` case is untested. At least there are no new tests for it.

lib/Mojo/DOM/CSS.pm
Eckankar: The else case won't be hit unless the function is called with the `root` node, which isn't currently the case. With or without the conditional, the line is dead in its current usage.

Do you prefer a plain `return undef;` without the conditional?

lib/Mojo/DOM/CSS.pm
Eckankar: Any existing test involving `:root` would be affected by this. Since `:root` is already covered by tests, I thought those tests would suffice to check that functionality.

lib/Mojo/DOM/CSS.pm
kraih: I don't know what the correct code would be, i'm just saying that the conditional that's currently there on the last line of the function is not useful.

Eckankar: Personally, I find it useful for the cases to be explicitly listed, so I can verify at a glance that the function covers the cases I want it to cover.

To move one step up in the tree, there are 3 cases: If you're on a tag, the parent is entry 3; if you're on a text node, the parent is entry 2; if you're on the root node, there is no parent. 

Easy to read and verify that it does what I'd like it to do.

If you would prefer it written in another way, I can do that, but it'd probably be best you tell me how you'd like it written in that case, rather than me trying variations until something sticks. :)

kraih:
This PR needs a lot of cleaning up, but i still think it has potential.

Eckankar:
Unfortunately, it looks like not much has happened on this in the past half year.

I'd like to inquire if there's still interest in getting this merged in?

If there's anything I can do to help facilitate getting this merged in, let me know. It is a feature I would like to use myself. :)

You have mentioned it requiring "a lot of cleaning up". Unfortunately, I'm not sure what kind of clean-up you are looking for. If you can be more specific about exactly what needs to be cleaned up, perhaps I can be of help. :)

kraih:
I'd still like to see the feature added, but it looks like nobody from the team currently has time to help with the cleanup unfortunately.

kraih:
Had some time and decided to finish the patch, thanks.

kraih:
There was a surprising amount of code that seemed to do nothing, if that's just missing tests it would be great if you could take a look. There is still a pretty high risk of the feature getting removed again if it ends up causing too many problems.

Eckankar:
I'm not sure what code you're referring to as doing nothing - comparing 66de0815 and bd2602d side by side, they look pretty similar, aside from minor rewrites, primarily in the name of shortening the code. (As far as I can tell.)

The only thing I noticed removed was my explicit [`return undef` when calling `_attr` on a non-tag](https://github.com/mojolicious/mojo/commit/66de0815d7ac524cbd319fc52d557f5e21778545#diff-a497a33049109dc96208d13a14c1612aR74) - but if it works without it, that was probably just something I needed at some point, but not the final product. Knowing exactly what type of value each method can be called can be rather cryptic to figure out, so I probably just played it safe.

If you run into any issues, feel free to ping me.

PR #1416, opened 2019-10-05
Substitution replaced with transliteration

k-mx:
### Summary
Transliteration is 10x faster and often more expressive

### Motivation
Of course changes are not necessary, but having faster operator inside nested loops is not a bad idea? :)

Good luck!

### DIFF
--- a/lib/Mojo/DOM/CSS.pm
+++ b/lib/Mojo/DOM/CSS.pm
@@ -303,7 +303,7 @@ sub _unescape {
   $value =~ s/\\([0-9a-fA-F]{1,6})\s?/pack 'U', hex $1/ge;
 
   # Remove backslash
-  $value =~ s/\\//g;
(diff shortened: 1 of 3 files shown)


k-mx:
1. No performance gain for Mojo::DOM::CSS, so I will undo this change.
2. Changes in Mojo::IOLoop::Server is about readability. I think with transliteration it looks nicer
3. tablify works about 5-7% faster with transliteration, but of course, it is not slow in existing variation. I can provide benchmark for tablify today, when I get home.

I just thought, that when we want to replace group of characters, transliteration is literally instrument for that. That's the main reason, not performance.

jhthorsen:
I'm indifferent to this change, but I also don't see the harm. 😐

kraih:

marcusramberg:
:+1:

kraih:
This PR passed the vote.

PR #1419, opened 2019-10-07
With empty roles

DabeDotCom:
### Summary
Allow `My::Class->with_roles()->new()`  # empty list

### Motivation
A) It doesn't hurt anything, and

B) I defined a base class that accepted an array of roles to [optionally] include...  When that list was empty, though, `with_roles()` caused the whole program to die.  :-(

### DIFF
--- a/t/mojo/roles.t
+++ b/t/mojo/roles.t
@@ -58,6 +58,16 @@ my $obj = Mojo::RoleTest->new(name => 'Ted');
 is $obj->name,  'Ted',       'attribute';
 is $obj->hello, 'hello Ted', 'method';
 
+# Empty roles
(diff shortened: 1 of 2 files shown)


t/mojo/roles.t
```diff
@@ -58,6 +58,15 @@ my $obj = Mojo::RoleTest->new(name => 'Ted');
 is $obj->name,  'Ted',       'attribute';
 is $obj->hello, 'hello Ted', 'method';
 
+# Empty roles
+my $fred = Mojo::RoleTest->with_roles()->new(name => 'Fred');
+is $fred->name,  'Fred',       'attribute';
+is $fred->hello, 'hello Fred', 'method';
+
```
kraih: Extra newline or missing comment.

Issue #1409, opened 2019-09-19
Move DATA section from Mojo::Util into a resources file


kraih:
Very simple task, we have other modules that do this already... like for the TLS certificates. So there is lots of examples.

s1037989:
In wanting to use Mojo::File::path like Mojo::IOLoop::TLS, I'm having troubles with `use Mojo::File 'path';`.  Mojo::File -- through Collection/ByteSteam -- depends on Mojo::Util, so it seems that Mojo::File can't find what it needs from Mojo::Util yet because Mojo::Util isn't loaded yet.
Am I correct in perceiving that this would be a problem for this task?  Suggestions?  Should I not use Mojo::File and just find and open the `resources` file using core Perl functions?

Grinnz:
You can use `require` to delay loading of another module so as not to create a circular import.

Grinnz:
Example: https://metacpan.org/release/Mojolicious/source/lib/Mojo/Base.pm#L13-14

s1037989:
Thanks, @Grinnz !

I'm on the latest commit, 4652927b
```
$ git log -1
commit 4652927b5b4fb3607dce667d013cb6611c6e49f9
Author: Sebastian Riedel <sri@cpan.org>
Date:   Thu Sep 12 00:33:19 2019 +0200

    Bump version
```

All tests pass:
```
$ prove -Ilib t/mojo/util.t 
t/mojo/util.t .. ok     
All tests successful.
Files=1, Tests=233,  1 wallclock secs ( 0.06 usr  0.01 sys +  0.39 cusr  0.01 csys =  0.47 CPU)
Result: PASS
```

Then I add Mojo::File:
```
$ git diff
diff --git a/lib/Mojo/Util.pm b/lib/Mojo/Util.pm
index 7171e36..b8f2958 100644
--- a/lib/Mojo/Util.pm
+++ b/lib/Mojo/Util.pm
@@ -36,6 +36,7 @@ use constant {
 
 # To generate a new HTML entity table run this command
 # perl examples/entities.pl
+require Mojo::File;
 my %ENTITIES;
 for my $line (split "\n", join('', <DATA>)) {
   next unless $line =~ /^(\S+)\s+U\+(\S+)(?:\s+U\+(\S+))?/;
```

Re-run the tests, and things break!
```
$ prove -Ilib t/mojo/util.t 
t/mojo/util.t .. "deprecated" is not exported by the Mojo::Util module
[...]

PR #1424, opened 2019-10-14
Moved HTML entities from DATA section to separate resources file.

rage311:
### Summary
Move HTML entities from Mojo::Util DATA section to separate resources file (#1409)

### Motivation
These changes were requested.

### References
#1409 

The use of `require` helps to circumvent some circular dependencies that occur otherwise.  I'm not sure if this is exactly what was envisioned, and I'm certainly open to input.


kraih:
I really don't like the inline requires. Perhaps the code would be cleaner if we just didn't use `Mojo::File` in this one case?

jberger:
> I really don't like the inline requires. Perhaps the code would be cleaner if we just didn't use `Mojo::File` in this one case?

I can agree with that. I'd suggest putting a comment in like "don't use Mojo::File here because of circular dependencies" or somethinf

Grinnz:
sibling makes it a little ugly to avoid it, can be cleaned up a bit but: `File::Spec->catfile(File::Basename::dirname(Cwd::realpath(__FILE__)), 'resources', 'html_entities.txt')`

rage311:
I reverted all the `require`s and used @Grinnz solution for a `sibling` equivalent.

lib/Mojo/Util.pm
```diff
@@ -37,11 +38,23 @@ use constant {
 # To generate a new HTML entity table run this command
 # perl examples/entities.pl
 my %ENTITIES;
-for my $line (split "\n", join('', <DATA>)) {
-  next unless $line =~ /^(\S+)\s+U\+(\S+)(?:\s+U\+(\S+))?/;
-  $ENTITIES{$1} = defined $3 ? (chr(hex $2) . chr(hex $3)) : chr(hex $2);
+{
+  # Don't use Mojo::File here due to circular dependencies
+  my $html_path
+    = File::Spec->catfile(File::Basename::dirname(Cwd::realpath(__FILE__)),
```
kraih: Why `Cwd::realpath` here?

lib/Mojo/Util.pm
```diff
@@ -37,11 +38,23 @@ use constant {
...
+    'resources', 'html_entities.txt');
+
+  local $/ = undef;
```
kraih: Why do this on an extra line instead of using the `do` idiom?

lib/Mojo/Util.pm
Grinnz: It handles the edge case where Mojo::Util is loaded via a symlink to still reference the correct location

lib/Mojo/Util.pm
rage311: Like this?
`my $file_content = do { local $/; <$fh> };`

I wasn't familiar with that pattern, I had to Google it.

lib/Mojo/Util.pm
kraih: How could that ever happen? Until your recent addition of `curfile` we did not use `Cwd::realpath` for any of this and there were never any issues.

lib/Mojo/Util.pm
kraih: Yes, that's the standard way of using `local $/` actually.

lib/Mojo/Util.pm
rage311: It does seem to work the same if I remove `Cwd::realpath`.

lib/Mojo/Util.pm
```diff
@@ -7,6 +7,7 @@ use Digest::MD5 qw(md5 md5_hex);
 use Digest::SHA qw(hmac_sha1_hex sha1 sha1_hex);
 use Encode 'find_encoding';
 use Exporter 'import';
+use File::Basename;
```
kraih: I'm almost certain there are default exports polluting the namespace now.

karenetheridge: You can use Test::CleanNamespaces in an xt/author test to know for sure.

Grinnz: Yes File::Basename has default exports, can be suppressed with `use File::Basename ()`

rage311:
Imported `File::Basename 'dirname'` only.
Changed to `do { ... }` pattern for setting the record separator and slurping the file.
Removed `Cwd::realpath`.

lib/Mojo/Util.pm
```diff
@@ -37,11 +38,21 @@ use constant {
...
-for my $line (split "\n", join('', <DATA>)) {
-  next unless $line =~ /^(\S+)\s+U\+(\S+)(?:\s+U\+(\S+))?/;
-  $ENTITIES{$1} = defined $3 ? (chr(hex $2) . chr(hex $3)) : chr(hex $2);
+{
+  # Don't use Mojo::File here due to circular dependencies
+  my $html_path = File::Spec->catfile(dirname(__FILE__),
+    'resources', 'html_entities.txt');
+
+  open my $html_fh, '<', $html_path
+    or croak 'Unable to open html entities file';
```
kraih: I don't like hiding the actual error, especially when it would be such an unexpected exception.

lib/Mojo/Util.pm
```diff
@@ -37,11 +38,21 @@ use constant {
...
+  my $entity_lines = do { local $/; <$html_fh> };
+  close $html_fh;
```
kraih: Do we gain anything by not just letting it go out of scope here?

rage311: A bit of explicitness, I suppose, but I don't feel strongly about it either way.

kraih:
Thanks, merged.

PR #1427, opened 2019-10-27
Auto-load plugins defined in MOJO_PLUGINS

s1037989:
### Summary
Auto-load plugins defined in MOJO_PLUGINS environment variable.

### Motivation
Use case:
I have a plugin based on [Doug Bell's post about Reverse Proxy]( https://mojolicious.io/blog/2019/03/18/reverse-proxy-with-path/) and I want to be able to download / install an app and not have to modify it in order to use this plugin.
This small change will auto-load any plugins defined in the MOJO_PLUGINS environment variable.

### References
None.
If this is an acceptable feature, I'll add test and documentation and update the PR.  If not, close it.

### DIFF
--- a/lib/Mojolicious.pm
+++ b/lib/Mojolicious.pm
@@ -172,7 +172,8 @@ sub new {
   $r->hide(qw(stash tx url_for write write_chunk));
 
   $self->plugin($_)
-    for qw(HeaderCondition DefaultHelpers TagHelpers EPLRenderer EPRenderer);
+    for qw(HeaderCondition DefaultHelpers TagHelpers EPLRenderer EPRenderer),
+    split ';', $ENV{MOJO_PLUGINS} || '';
 
   # Exception handling should be first in chain
   $self->hook(around_dispatch => \&_exception);


jhthorsen:
I've doing something similar in [Convos](https://github.com/Nordaaker/convos/blob/12cb600494d2dfe9d5b6dfeec9c3ba275910fa3f/lib/Convos.pm#L196), but I'm not sure if I think it's a good idea to do it as a generic feature for Mojolicious.

If we do, then we should also have the possibility to pass in arguments, and then things are getting more difficult.

So I'm giving this PR 👎 as it stands now.

CandyAngel:
Regarding passing in arguments, an "AutoPlugin" plugin which loads plugins from a configuration file could work. Specifying "MOJO_PLUGINS" with arguments could be the list of plugins to load from the configuration file (in addition to any already set to enabled).

```
{
  autoplugin => {
    PluginName => {
      enabled => 0,
      args => {
        route_prefix => 'Foo',
      },
    },
  },
}
```

Kind of how the `CHI` plugin will load its own configuration from configuration files, rather than having to pass them in. Not sure if that behaviour is approved of though! :P

s1037989:
@jhthorsen Unfortunately, making it a generic feature for Mojolicious is the only way to allow a user to override a capability in an app without forking the app. An alternative would be to mount the given application into a custom application, but there are some features of Mojolicious that cannot work as a mounted app and therefore some apps may simply not be able to be mounted.

@CandyAngel That sounds like a great way to handle passing in arguments.

Grinnz:
You could maybe make a role that applies to the application to create a new application.

s1037989:
Hi, @Grinnz !  Thant sounds like an intriguing suggestion!  Could you say more?  I'm having a hard time picturing this...

kraih:
Is this PR abandoned?

s1037989:
Is there interest in including this in core? If so, are there any decisions that need to be made?

christopherraa:
I feel like @jhthorsen here, that this would have to handle plugin arguments _somehow_ or we run into the issue of `MOJO_PLUGINS` only being usable in certain situations. At `$work` the standard is to set up new Mojolicious applications as containerized workloads, and as such we try to avoid writing configuration files pr deployment. Usually we have a file in the project looking something like this:

```perl
{
    some_setting    => $ENV{SOME_SETTING}    // 'some-sane-default',
    another_setting => $ENV{ANOTHER_SETTING} // 'another-sane-default',
}
```
That file holds sane defaults and let us override all variables from our CI/CD or container orchestration tools. It has proven extremely convenient not having to deal with additional files _pr-deployment_. That configuration file is checked in to the repository and never changed as part of deployment.

The reason for me being this verbose is to say that I do not personally see the need to arbitrarily load plugins from an environment variable (we usually know which we want our app to use, and try to limit them), but if that usecase is to be supported it needs to also handle arguments. If the solution to handling arguments is introducing another file in the deployment then it suddenly looses attractiveness for me.

As a side-note: if I felt an application needed to auto-load arbitrary plugins I would most likely build that into each application that needed it. I have yet to run into this need.

@s1037989 Thank you for your effort though. Don't let my comment be a discouragement. If you feel this would be useful to you and the general community then feel free to work at a general solution. If it "plays well" with patterns used for containerized workloads then I'd be more than happy to review.

As it stands I am currently :-1: .

kraih:
I don't see a scenario where this turns into a core Mojo feature. Try making it a plugin on CPAN first.

s1037989:
Thanks, everyone, for your comments!  Closing this PR.

This PR was just a simple and general solution to a single, very specific problem: how can one launch a downloaded Mojolicious app on a Reverse Proxy With Path (see the link in the original PR), without modifying the app's code to install a before_dispatch hook?

I can see how this approach is definitely the wrong solution to the problem.

PR #1436, opened 2019-11-19
fix mojo get index command to not rely on removed Mojo::Collection->slice

Grinnz:
### Summary
The mojo get command still had an instance of the recently removed slice method. This replaces it with creating a new collection of that index since that's all it was doing.


Grinnz:
Tests are failing due to the recent EV 4.28 release, which is failing to build on travis's trusty machines.

PR #1437, opened 2019-11-20
support a behaviour like 'with' when using roles

reneeb:
### Summary
Currently Mojo::Base's with_roles creates a new class, but we often need to apply the roles
to the package. So this is basically the behavior of Moose's `with` and/or `Role::Tiny::With`

### Motivation
We do not want to apply the roles when we create a new object. We want the role behavior included in the classes.

### DIFF
--- a/lib/Mojo/Base.pm
+++ b/lib/Mojo/Base.pm
@@ -127,6 +127,18 @@ sub import {
       experimental->import('signatures');
     }
 
+    # apply role to package
+    elsif ($flag eq '-with') {
+      Carp::croak 'Role::Tiny 2.000001+ is required for roles' unless ROLES;
+
+      my $role_list = shift @flags;
+      my @roles = map {
+        /^\+(.+)$/ ? "${caller}::Role::$1" : $_
+      } (ref $role_list ? @{$role_list} : $role_list);
+
+      Role::Tiny->apply_roles_to_package($caller, @roles);
+    }
+
     # Module
     elsif ($flag !~ /^-/) {
       no strict 'refs';
@@ -153,17 +165,9 @@ sub tap {
 sub with_roles {
   Carp::croak 'Role::Tiny 2.000001+ is required for roles' unless ROLES;
   my ($self, @roles) = @_;
-
-  my $opts = pop @roles if ref $roles[-1];
-
   return $self unless @roles;
 
-  my $method = 'create_class_with_roles';
-  if ( $opts && $opts->{apply_to_package} ) {
-    $method = 'apply_roles_to_package';
-  }
-
-  return Role::Tiny->$method($self,
+  return Role::Tiny->create_class_with_roles($self,
     map { /^\+(.+)$/ ? "${self}::Role::$1" : $_ } @roles)
     unless my $class = Scalar::Util::blessed $self;
 
@@ -267,6 +271,11 @@ enable support for L<subroutine signatures|perlsub/"Signatures">.
 
 This will also disable experimental warnings on versions of Perl where this
 feature was still experimental.
+
+If you use Roles, you can apply them to your class with C<-with>:
+
+  use Mojo::Base -with => ['Some::Role','+Logger'];
+  use Mojo::Base -with => 'Some::Role';
 
 =head1 FLUENT INTERFACES
 
@@ -392,28 +401,6 @@ returns the new class, or if called on an object reblesses the object into the
 new class. For roles following the naming scheme C<MyClass::Role::RoleName> you
 can use the shorthand C<+RoleName>. Note that role support depends on
 L<Role::Tiny> (2.000001+).
-
-  # Create a new class with the role "SubClass::Role::Foo" and instantiate it
-  my $new_class = SubClass->with_roles('+Foo');
-  my $object    = $new_class->new;
-
-When the last parameter passed to C<with_roles> is a hash reference, you can
-pass options. Then you can use roles and apply them to the package where
-C<with_roles> is used.
-
-  # a simple class that applies a role
-  package AppliedRoles;
-  use Mojo::Base -base;
-
-  # Logger provides a method 'log'
-  __PACKAGE__->with_roles( '+Logger', { apply_to_package => 1 } );
-  1;
-
-  # in other code
-  use AppliedRoles;
-
-  my $obj = AppliedRoles->new;
-  $obj->log( 'a message' );
 
 =head1 SEE ALSO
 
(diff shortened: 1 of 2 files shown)


jhthorsen:
I will not be in favor of this PR, since it's way too much typing to use it. I think I would rather have something like:

```
use Mojo::Base 'Some::Class', -with => ['Some::Role'];
use Mojo::Base -with => 'Some::Role'; # "-base" implied?
```

Grinnz:
This is already provided by Role::Tiny::With.

reneeb:
I like @jhthorsen  suggestion.

@Grinnz: It feels weird to use Role::Tiny::With to use one Role::Tiny feature while other Role::Tiny features can be used with a Mojo::Base method.

Grinnz:
Not sure what you mean. All of the Mojo::Base role features are just wrappers to use Role::Tiny properly. There is no wrapper required here.

reneeb:
But you suggested the use of Role::Tiny::With to have the 'with' like in Moose...

Grinnz:
Yes, this is the appropriate way to compose a role statically into a class.

reneeb:
To be more precise, to use Role::Tiny's `apply_roles_to_object` and `create_class_with_roles` can be used with `with_roles`, but not `apply_roles_to_package`.

jberger:
I tend to agree with @Grinnz and @jhthorsen here. You can already accomplish this with the Role::Tiny standard module Role::Tiny::With. If we wanted something in core, I'd vote for adding a -with flag to be consistent with the existing Mojo::Base mechanisms

reneeb:
I've updated the PR to use the "-with" flag that @jhthorsen proposed...

lib/Mojo/Base.pm
```diff
@@ -385,10 +402,6 @@ new class. For roles following the naming scheme C<MyClass::Role::RoleName> you
 can use the shorthand C<+RoleName>. Note that role support depends on
 L<Role::Tiny> (2.000001+).
 
-  # Create a new class with the role "SubClass::Role::Foo" and instantiate it
-  my $new_class = SubClass->with_roles('+Foo');
-  my $object    = $new_class->new;
```
jhthorsen: Why are these lines removed?

reneeb: My mistake...

reneeb:
@jhthorsen I'll have a look at it tomorrow...

lib/Mojo/Base.pm
```diff
@@ -127,6 +127,18 @@ sub import {
...
     }
 
+    # apply role to package
+    elsif ($flag eq '-with') {
+      Carp::croak 'Role::Tiny 2.000001+ is required for roles' unless ROLES;
+
+      my $role_list = shift @flags;
+      my @roles = map {
+        /^\+(.+)$/ ? "${caller}::Role::$1" : $_
+      } (ref $role_list ? @{$role_list} : $role_list);
```
kraih: I bet this block could be much shorter.

t/mojo/roles.t
```diff
@@ -46,6 +46,25 @@ use Mojo::Base -role;
 
 sub hello {'hello mojo!'}
 
+package Mojo::RoleListApplied::Role::LOUD;
```
kraih: We include the word `Test` in test class names.

t/mojo/roles.t
```diff
@@ -124,5 +143,16 @@ is $dom->hello, 'hello mojo!', 'right result';
 my $file = Mojo::File->with_roles('Mojo::RoleTest::Hello')->new;
 is $file->hello, 'hello mojo!', 'right result';
 
+# role applied to package, do not create a new class
```
kraih: Inconsistently formatted comment.

t/mojo/roles.t
```diff
@@ -124,5 +143,16 @@ is $dom->hello, 'hello mojo!', 'right result';
...
+my $list_applied = Mojo::RoleListApplied->new;
+is $list_applied->hello, 'hello mojo!';
+is $list_applied->yell, 'HEY!';
+
```
kraih: Extra newline.

t/mojo/roles.t
```diff
@@ -124,5 +143,16 @@ is $dom->hello, 'hello mojo!', 'right result';
...
+my $role_applied = Mojo::RoleApplied->new;
+is $role_applied->yell, 'HEY!';
+
```
kraih: Extra newline.

lib/Mojo/Base.pm
```diff
@@ -260,6 +272,11 @@ enable support for L<subroutine signatures|perlsub/"Signatures">.
 This will also disable experimental warnings on versions of Perl where this
 feature was still experimental.
 
+If you use Roles, you can apply them to your class with C<-with>:
+
+  use Mojo::Base -with => ['Some::Role','+Logger'];
+  use Mojo::Base -with => 'Some::Role';
```
kraih: Shouldn't there be another flag?

lib/Mojo/Base.pm
```diff
@@ -260,6 +272,11 @@ enable support for L<subroutine signatures|perlsub/"Signatures">.
 This will also disable experimental warnings on versions of Perl where this
 feature was still experimental.
 
+If you use Roles, you can apply them to your class with C<-with>:
```
kraih: Inconsistent formatting.

kraih: And description of the feature.

lib/Mojo/Base.pm
jhthorsen: Here's my idea:

```perl
my $roles  = shift @flags;
my $prefix = "${caller}::Role::";
Role::Tiny->apply_roles_to_package($caller,
  map { /^\+(.+)/ ? "$prefix$1" : $_ } ref $roles ? @$roles : $roles);
```

lib/Mojo/Base.pm
```diff
@@ -260,6 +272,11 @@ enable support for L<subroutine signatures|perlsub/"Signatures">.
...
+
+  use Mojo::Base -with => ['Some::Role','+Logger'];
```
jhthorsen: Missing whitespace inside the array-ref.

kraih:
As we've recently learned from the `-async` flag, you can't use commonly used function names for flags. So using `-with` here is impossible.

kraih:
I'm rejecting this PR without a vote for the reasons mentioned above. Too dangerous.

shadowcat-mst:
package MyApp::Base;
    
    use Mojo::Base -strict;
    use Role::Tiny::With ();
    use Import::Into;
    
    sub import {
        shift;
        Mojo::Base->import::into(1, @_);
        Role::Tiny::With->import::into(1);
    }
    
    1;

PR #1439, opened 2019-11-22
Fail fast if asked to listen on invalid unix+http URL

nrdvana:
### Summary

Currently, if Mojo::Server::Daemon is asked to listen on a http+unix URL that lacks a Host portion, it silently falls back to listening on TCP:*, but still prints that it is listening on the requested URL.  This change causes the daemon to abort with a helpful error so the user can correct the URL.

### Motivation

The current behavior could possibly cause security problems if a sensitive service listens on a public IP instead of a unix socket.

But more likely, I just want to help the next person who runs into examples of [malformed URLs][1] so they don't have to go digging through the code to find out why it isn't working.

### References

[1] https://github.com/mojolicious/mojo/wiki/WHM-Easy-Apache-4-Hypnotoad-with-Unix-Sockets

Listens on TCP *:
```
app->start('daemon', -l => 'http+unix:test.sock');
```

Listens on `./test.sock`:
```
app->start('daemon', -l => 'http+unix://test.sock');
```

### DIFF
--- a/lib/Mojo/Server/Daemon.pm
+++ b/lib/Mojo/Server/Daemon.pm
@@ -171,7 +171,11 @@ sub _listen {
   my $query   = $url->query;
   my $options = {backlog => $self->backlog};
   $options->{$_} = $query->param($_) for qw(fd single_accept reuse);
-  if ($proto eq 'http+unix') { $options->{path} = $url->host }
+  if ($proto eq 'http+unix') {
+    $options->{path} = $url->host;
+    croak qq{Socket path must be specified in host portion of URL}
+      unless defined $options->{path} && length $options->{path};
+  }
   else {
     if ((my $host = $url->host) ne '*') { $options->{address} = $host }
     if (my $port = $url->port) { $options->{port} = $port }


jhthorsen:
The problem here seems to be that there's no scheme portion?

marcusramberg:
I am neutral to this change.

nrdvana:
@jhthorsen Well that's an interesting point, though in the TCP cases there is other code that performs the sanity checks:

```
app.pl daemon --listen http://ldskfjldkfj:80
Can't create listen socket: Name or service not known at /home/user/perl5/perlbrew/perls/perl-5.26.3/lib/site_perl/5.26.3/Mojo/IOLoop.pm line 124.
```

```
app.pl daemon --listen http://*:xyz
Can't create listen socket: Servname not supported for ai_socktype at /home/user/perl5/perlbrew/perls/perl-5.26.3/lib/site_perl/5.26.3/Mojo/IOLoop.pm line 124.
```

Though interestingly, nothing ever checks for a valid port range:
```
app.pl daemon --listen http://*:65537
[2020-01-12 12:24:37.23116] [22663] [info] Listening at "http://*:65537"
Server available at http://127.0.0.1:65537
```
(is actually listening on port 1)

and lacking a port in the URI, it listens on a random port:
```
daemon --listen http:///foo
[2020-01-12 12:33:36.50697] [22721] [info] Listening at "http:///foo"
Server available at http:///foo
```

but none of those cases are quite as extreme as listening on a tcp port when asked to listen on a unix socket, and without any warning.

@kralh I can add tests if that is the only reason.  The valid listen URLs already have test cases, and I didn't figure you'd want to slow down the test suite just to check on handling of invalid URLs.

jhthorsen:
Seems like there's not enough interest in this issue to take it one way or the other, so I'm closing it.

PR #1440, opened 2019-11-23
Missing $ sign to loop variable in the docs

marcoarthur:
### Summary
Small documentation fix in Mojo::IOLoop::Delay


kraih:
Looks like this has already been fixed in the meantime. Please don't open overly simplistic pull requests. If you want to check Mojolicious for typos, it should be a check of the whole project.

Issue #1441, opened 2019-11-24
Default "text/plain;charset=UTF-8" returned by Mojolicious::Types::content_type may be confusing


aitap:
* Mojolicious version: 8.26
* Perl version: 5.24.1
* Operating system: Debian GNU/Linux 9.11 (stretch)

### Steps to reproduce the behavior

Use the `reply->file` helper to serve a static binary file not known by `Mojolicious::Types`, such as a `.tar.xz` archive or a `.gpg` encrypted file and try to get the served content using a browser.

### Expected behavior

Browser offers to download the binary file.

### Actual behavior

Browser displays a screenful of [mojibake](https://en.wikipedia.org/wiki/Mojibake).
Browser assumes that the reply can be displayed as text because Mojolicious sends the `Content-Type: text/plain;charset=UTF-8` header:

```
$ perl -Mojo -E'a("/" => sub { shift->reply->file("test.tar.xz") })->start' get -M HEAD /
HEAD / HTTP/1.1
Host: 127.0.0.1:38719
User-Agent: Mojolicious (Perl)
Content-Length: 0
Accept-Encoding: gzip

HTTP/1.1 200 OK
Last-Modified: Sun, 24 Nov 2019 17:39:39 GMT
ETag: "a31a2393cc3244a02333b5515918bf9d"
Accept-Ranges: bytes
Content-Length: 6701144
Content-Type: text/plain;charset=UTF-8
Date: Sun, 24 Nov 2019 17:43:06 GMT
Server: Mojolicious (Perl)
```

The `reply->file` helper calls `Mojolicious::Static::serve_asset`, which uses `Mojolicious::Types::content_type` (currently marked as experimental) to determine the `Content-Type` to send back to the user. Before the introduction of `content_type`, `serve_asset` didn't send the `Content-Type` header at all.

`Mojolicious::Types::content_type` [defaults](/mojolicious/mojo/blob/243f705fc67c4bd4c1b2f4ad4f405cc49b4dabb2/lib/Mojolicious/Types.pm#L43) [to `text/plain;charset=UTF-8`](/mojolicious/mojo/blob/243f705fc67c4bd4c1b2f4ad4f405cc49b4dabb2/lib/Mojolicious/Types.pm#L27) for types it cannot classify otherwise. A default of [`application/octet-stream`](https://tools.ietf.org/html/rfc2046#section-4.5.1) might be a safer option. Alternatively, the default `Content-Type` could be made into a setting.

[...]

PR #1442, opened 2019-11-24
default to Content-Type: application/octet-stream

aitap:
### Summary

Mojolicious::Types::content_type used to return `text/plain;charset=UTF-8`
for unknown types. Use `application/octet-stream` as default instead.

### Motivation

`Content-Type: text/plain;charset=UTF-8` may confuse browsers if returned
with binary data.

### References

Fixes: mojolicious/mojo#1441


kraih:

jhthorsen:
👍 from me as well. Same reasoning as @Grinnz.

PR #1450, opened 2019-12-18
Add ability to test HTML attribute values with Test::Mojo

pau4o:
### Summary
Allow to test values of HTML attributes with test methods

### Motivation
In one of projects I need to test value in HTML attribute which is generated.
I don't like idea of concatenating strings in tests, so decided that having test methods for HTML attributes will be good addition to Test::Mojo

### DIFF
--- a/lib/Test/Mojo.pm
+++ b/lib/Test/Mojo.pm
@@ -30,6 +30,30 @@ sub app {
   return $self;
 }
 
+sub attr_is {
+  my ($self, $selector, $attribute, $value, $desc) = @_;
+  return $self->_test('is', $self->_attr($selector, $attribute),
+    $value, _desc($desc, qq{exact match for attribute "$attribute" at selector "$selector"}));
+}
+
+sub attr_isnt {
+  my ($self, $selector, $attribute, $value, $desc) = @_;
+  return $self->_test('isnt', $self->_attr($selector, $attribute),
+    $value, _desc($desc, qq{no match for attribute "$attribute" at selector "$selector"}));
+}
+
+sub attr_like {
+  my ($self, $selector, $attribute, $regex, $desc) = @_;
+  return $self->_test('like', $self->_attr($selector, $attribute),
+    $regex, _desc($desc, qq{similar match for attribute "$attribute" at selector "$selector"}));
+}
+
+sub attr_unlike {
+  my ($self, $selector, $attribute, $regex, $desc) = @_;
+  return $self->_test('unlike', $self->_attr($selector, $attribute),
+    $regex, _desc($desc, qq{no similar match for attribute "$attribute" at selector "$selector"}));
+}
+
 sub content_is {
   my ($self, $value, $desc) = @_;
   return $self->_test('is', $self->tx->res->text,
@@ -337,6 +361,13 @@ sub websocket_ok {
   return $self->_request_ok($self->ua->build_websocket_tx(@_), $_[0]);
 }
 
+sub _attr {
+  my ($self, $selector, $attribute) = @_;
+  return '' unless my $e = $self->tx->res->dom->at($selector);
+  return '' unless my $attr_value = $e->attr($attribute);
+  return $attr_value;
+}
+
 sub _build_ok {
   my ($self, $method, $url) = (shift, shift, shift);
   local $Test::Builder::Level = $Test::Builder::Level + 1;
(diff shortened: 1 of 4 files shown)


lib/Test/Mojo.pm
```diff
@@ -337,6 +361,13 @@ sub websocket_ok {
   return $self->_request_ok($self->ua->build_websocket_tx(@_), $_[0]);
 }
 
+sub _attr {
+  my ($self, $selector, $attribute) = @_;
+  return '' unless my $e = $self->tx->res->dom->at($selector);
+  return '' unless my $a = $e->attr($attribute);
+  return $a;
```
kraih: `$a` is a reserved variable in Perl.

pau4o: Yeah, I know ;-)
Added commit to fix this.

kraih:
Squash followup commits.

pau4o:
@jhthorsen Is there an easier and more fun way to test values of HTML tag attributes?

I think that documentation clearly state that tests are for given HTML attribute on element selected by given selector. Please note that methods are named similarly to ["attr" in Mojo::Dom](https://metacpan.org/pod/Mojo::DOM#attr)
Can you propose better naming?

kraih:
```perl
$t->element_exists('p.joy[style*=color]');
```

pau4o:
@kraih using anchored attribute selectors:
- `[attr*=value]` - elements with attribute which contains `value`
- `[attr^=value]` - elements with attribute which starts with `value`
- `[attr$=value]` - elements with attribute which ends with `value`
or more advanced attribute selectors:
- `[attr~=value]` - one of words generated with split by white-space is `value`
- `[attr|=value]` - attribute is `value` or starts with `value-` 

give us some possibilities, but they are very limited compared to proposed `attr_like` which can use full power of Perl regular expressions.

kraih:
The code, tests and docs look fine to me. Now we just have to find out if we actually want the feature in core Mojolicious (it can always be a role on CPAN). So i'm calling for a vote @mojolicious/core.

marcusramberg:
Seems like a good for for the Core to me. :+1:

lib/Test/Mojo.pm
```diff
@@ -590,6 +621,36 @@ Access application with L<Mojo::UserAgent::Server/"app">.
   $t->get_ok('/hello')->status_is(200);
   is $stash->{foo}, 'bar', 'right value';
 
+=head2 attr_is
+
+  $t = $t->attr_is('img.cat' => 'alt', 'Grumpy cat');
+  $t = $t->attr_is('img.cat' => 'alt', 'Grumpy cat', 'right alt text');
```
kraih: Considering there are 3 required args, the fat comma is probably not the right choice here.

jhthorsen:
I’m 👎 to this, since I’ve tried to do this before, but it always falls short in most cases. I think we should optimize for the common cases, which is already covered imo.

s1037989:
@jhthorsen in what way has it fallen short for you in the past?

jberger:
While I have always used the element_exists test with complex attribute selectors, this does make a simple and relatively common use-case much easier. :+1:

lib/Test/Mojo.pm
pau4o: Replaced fat comma

pau4o:
After amending commit to remove usage of "fat coma" in POD as suggested [above](https://github.com/mojolicious/mojo/pull/1450#discussion_r360647364) there is failure in AppVeyour when trying to install Strawberry Perl https://ci.appveyor.com/project/kraih73737/mojo/builds/29705033/job/bckbh3jmmbs3il7x#L8636

kraih:
Passes now.

kraih:
I'll let this vote run a little longer since some people might have been on vacation.

kraih:
This is a reminder for everyone in @mojolicious/core who was on vacation that this vote is still in progress.

yusrideb:
I've tested it on [App-Mojo-Test](https://github.com/yusrideb-mojolicious-test/App-Mojo-Test) it seems very useful 

And the results of travis also succeeded :
[Build Status](https://travis-ci.com/yusrideb-mojolicious-test/App-Mojo-Test)

marcusramberg:
This one has passed the vote as well. :tada:

kraih:
Unfortunately we all missed that [perltidy had not been used](https://github.com/mojolicious/mojo/compare/58025aa9a022...e0bcc47ae6e9) for this PR.

PR #1452, opened 2019-12-27
Support async/await in Mojo::EventEmitter

kraih:
Currently using `Mojo::EventEmitter` with `async`/`await` is very awkward, since the promises returned by async functions need to be referenced manually. This PR is a possible solution, but there is a performance cost. We use `Mojo::EventEmitter` very extensively in Mojolicious (and countless 3rd party modules), so it is hard to say if this could become a real problem.

Here are [before](https://gist.github.com/kraih/d3741337e881f35035549d077b16044f) and [after](https://gist.github.com/kraih/ba9fad6b5676d440a14ccbf0f4e20658) example apps.

### DIFF
--- a/lib/Mojo/EventEmitter.pm
+++ b/lib/Mojo/EventEmitter.pm
@@ -12,11 +12,15 @@ sub emit {
 
   if (my $s = $self->{events}{$name}) {
     warn "-- Emit $name in @{[blessed $self]} (@{[scalar @$s]})\n" if DEBUG;
-    for my $cb (@$s) { $self->$cb(@_) }
+    for my $cb (@$s) {
+      my $val = $self->$cb(@_);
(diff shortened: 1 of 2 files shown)


kraih:
I'm very unsure what the correct way for handling this is. The `finally` hack could also be applied at the `Mojo::Promise` async protocol level, but that essentially disables a feature of `Future::AsyncAwait` to detect unhandled promises.

kraih:
I'm bringing this up so early because i'm fairly certain many of our users will want to use `async`/`await` with WebSockets.

kraih:
After thinking some more about this, the impact on the `Mojo::IOLoop::Stream` layer really worries me. Some events are very relevant to overall performance.

kraih:
I think we can rule out this solution, it's pretty terrible.

PR #1453, opened 2019-12-27
Allow for async functions to be used almost everywhere

kraih:
This is an alternative to #1452. It has the advantage that async functions work pretty much everywhere automatically. It does have the downside that it deactivates the `Future::AsyncAwait` feature for detecting unhandled promises however. So it becomes much easier to miss exceptions in async functions.


kraih:
If we decide to go with this solution, we should probably switch to a much more efficient internal alternative to `finally` (which requires the creation of yet another promise).

kraih:
This solution is almost there, we just need a way to detect and throw unhandled rejections during `DESTROY`.

Issue #1456, opened 2020-01-10
Test::Mojo->json_is(): misleading output on some test failures


DrHyde:
* Mojolicious version: 8.29
* Perl version: 5.26.1
* Operating system: High Sierra

When testing a JSON value like `{ "fruit": "lemon" }` with `...->json_is('/animal', 'bat')` we correctly get a failure, but the diagnostic message complains that `undef` doesn't equal `'bat'` which isn't quite right. The value isn't even undef, it *doesn't exist*. Same applies to `json_like`.

jhthorsen:
How do you suggest this to be fixed @DrHyde?

DrHyde:
My plan was to factor out the important bit of `json_has` and reuse that to decide whether to spit out an alternative message.

DrHyde:
FYI ... [work in progress](https://github.com/DrHyde/mojo/commits/improve-error-output-from-json_is). Still need to muck about with `json_like`.

jhthorsen:
The problem with https://github.com/DrHyde/mojo/commit/3417e72b8def8e3da5a0ee80fff83e751d323951#diff-a3cb567ecd4d756d754d1baee62ee931R215 is that it ignores `$desc`.

DrHyde:
Hurrah for code review :-)

DrHyde:
This is activity. The issue shouldn't be closed.

kraih:
I do not appreciate attempts to cheat the system. So i'm closing this issue right away. If you're still working on a patch you should have said so.

PR #1459, opened 2020-01-16
Improve error output from json_is and json_like

DrHyde:
### Summary
Tweaked `Test::Mojo::json_is` and `Test::Mojo::json_like` to detect when a pointer is invalid, and to then just `fail()` instead of comparing data using `is_deeply()` or `like()`. The detection is done using what used to be the core of `json_has` and `json_hasnt`, which has been factored out.

Also added tests for `json_is` and `json_like`.

### Motivation
`...->json($p)` returns `undef` when $p points to lala-land, which leads to `is_deeply()` and `like()` trying to compare the desired value to undef, failing (usually) as expected, *but then erroneously talking about a comparison with undef in their explanations* as if the pointer was to an undef value instead of pointing to nothing at all. This may also lead to confusion when there are legitimate JSON `null`s in the data although I've not tested that.

### References
[issue 1456](https://github.com/mojolicious/mojo/issues/1456)

### DIFF
--- a/lib/Test/Mojo.pm
+++ b/lib/Test/Mojo.pm
@@ -191,23 +191,29 @@ sub header_unlike {
     $regex, $desc);
 }
 
+sub _does_json_contain {
+  my ($self, $p) = @_;
+  Mojo::JSON::Pointer->new($self->tx->res->json)->contains($p);
+}
+
 sub json_has {
   my ($self, $p, $desc) = @_;
   $desc = _desc($desc, qq{has value for JSON Pointer "$p"});
-  return $self->_test('ok',
-    !!Mojo::JSON::Pointer->new($self->tx->res->json)->contains($p), $desc);
+  return $self->_test('ok', !!_does_json_contain($self, $p), $desc);
 }
 
 sub json_hasnt {
   my ($self, $p, $desc) = @_;
   $desc = _desc($desc, qq{has no value for JSON Pointer "$p"});
-  return $self->_test('ok',
-    !Mojo::JSON::Pointer->new($self->tx->res->json)->contains($p), $desc);
+  return $self->_test('ok', !_does_json_contain($self, $p), $desc);
 }
 
 sub json_is {
   my $self = shift;
   my ($p, $data) = @_ > 1 ? (shift, shift) : ('', shift);
+  if($p && !_does_json_contain($self, $p)) {
+      return $self->_test('fail', "no data for pointer $p");
+  }
   my $desc = _desc(shift, qq{exact match for JSON Pointer "$p"});
   return $self->_test('is_deeply', $self->tx->res->json($p), $data, $desc);
 }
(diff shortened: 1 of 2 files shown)


lib/Test/Mojo.pm
```diff
@@ -191,31 +191,40 @@ sub header_unlike {
     $regex, $desc);
 }
 
+sub _does_json_contain {
+  my ($self, $p) = @_;
+  Mojo::JSON::Pointer->new($self->tx->res->json)->contains($p);
```
jhthorsen: This method is not in the right place in the file. All methods are ordered alphabetically. I also think I would rename it so it takes less time to parse what the method does:

```perl
sub _json_contains {
  return Mojo::JSON::Pointer->new(shift->tx->res->json)->contains(shift);
}
```

lib/Test/Mojo.pm
```diff
@@ -191,31 +191,40 @@ sub header_unlike {
...
+}
+
 sub json_has {
   my ($self, $p, $desc) = @_;
   $desc = _desc($desc, qq{has value for JSON Pointer "$p"});
-  return $self->_test('ok',
-    !!Mojo::JSON::Pointer->new($self->tx->res->json)->contains($p), $desc);
+  return $self->_test('ok', !!_does_json_contain($self, $p), $desc);
```
jhthorsen: Why don't you call a method, instead of a function?

lib/Test/Mojo.pm
```diff
@@ -191,31 +191,40 @@ sub header_unlike {
...
-  return $self->_test('ok',
-    !Mojo::JSON::Pointer->new($self->tx->res->json)->contains($p), $desc);
+  return $self->_test('ok', !_does_json_contain($self, $p), $desc);
 }
 
 sub json_is {
   my $self = shift;
   my ($p, $data) = @_ > 1 ? (shift, shift) : ('', shift);
   my $desc = _desc(shift, qq{exact match for JSON Pointer "$p"});
+  if($p && !_does_json_contain($self, $p)) {
```
jhthorsen: This PR must be run through perltidy. Pretty sure this if can be condensed into fewer lines of code as well.

t/test/mojo.t
```diff
@@ -0,0 +1,118 @@
+use Mojo::Base -strict;
```
jhthorsen: I'm not sure if this file is required. What does it really test? I can't see that we test Test::Mojo directly anywhere.

kraih:
Afraid i have to disagree with @jhthorsen here, i don't think the little gain is worth maintaining those messy tests.

kraih:
That said, better test coverage for our `Test::Mojo` methods is not a bad idea at all. But if we go there it needs to be consistent.

kraih:
Having thought about this a bit more, i think we need real tests for all `Test::Mojo` methods with their current behaviour before we can think about handling more edge cases. So this PR is blocked until that happened.

jhthorsen:
Seems like this PR has gone stale.

I also agree that figuring out #1460 first is a good idea.

DrHyde:
On the whole I agree that having comprehensive tests before applying this would be a good thing, I just (currently) lack the time to create them. It's on my rather long to-do list.

Issue #1431, opened 2019-11-02
TLS tests fail on newer versions of macOS


kraih:
* Mojolicious version: 8.26
* Perl version: v5.30.0
* Operating system: macOS 10.15.1

### Steps to reproduce the behavior
Run tests with `TEST_TLS=1` and `OpenSSL 1.1.1d` from Homebrew.

### Expected behavior
```
t/mojo/tls.t .......................... ok
```

### Actual behavior
```
#   Failed test 'AES256-SHA has been negotiatied'
#   at t/mojo/tls.t line 55.
#          got: 'TLS_AES_256_GCM_SHA384'
#     expected: 'AES256-SHA'

#   Failed test 'AES256-SHA has been negotiatied'
#   at t/mojo/tls.t line 57.
#          got: 'TLS_AES_256_GCM_SHA384'
#     expected: 'AES256-SHA'
# Looks like you failed 2 tests of 6.
t/mojo/tls.t ............................... Dubious, test returned 2 (wstat 512, 0x200)
```

rage311:
I don't know whether this is helpful or not, but the `tls.t` tests pass on OpenBSD 6.6 (-current) with LibreSSL 3.0.2.

```
% uname -rsvp
OpenBSD 6.6 GENERIC.MP#424 amd64

% openssl version
LibreSSL 3.0.2

~/dev/mojo % TEST_TLS=1 prove -l t/mojo/tls.t
t/mojo/tls.t .. ok   
All tests successful.
Files=1, Tests=6,  0 wallclock secs ( 0.04 usr  0.01 sys +  0.56 cusr  0.06 csys =  0.67 CPU)
Result: PASS
```

tianon:
It seems odd that it's affecting OpenSSL from Homebrew, but could it be similar to what's described in https://github.com/FiloSottile/mkcert/issues/174#issuecomment-508955273 ? (Changes to maximum certificate validity in newer macOS)

kraih:
I just installed LibreSSL 2.9.2 with Homebrew, and all tests still pass with it. Looks like it's only OpenSSL.

marked:
- Ubuntu 19.10
- OpenSSL 1.1.1c  28 May 2019
- Perl        (v5.28.1, linux)
- Mojolicious (8.32, Supervillain)

```
$ export TEST_TLS=1
$ prove tls.t
tls.t .. 1/?
#   Failed test 'AES256-SHA has been negotiatied'
#   at tls.t line 55.
[...]

PR #1468, opened 2020-02-11
Fix tls.t test fail with OpenSSL 1.1.1 by checking TLS version before checking cipher

marked:
### Summary
OpenSSL has moved on to TLSv1.3 which has a different cipher suite than TLSv1.2.  The test was expecting/checking a particular cipher to have been negotiated which is not one of the choices.

### Motivation

This is the smallest code change to fix this test to pass with OpenSSL 1.1.1.

The alternative(s) are to set a default version in Mojo::IOLoop::TLS instead or perhaps a large code change to instead have the negotiation fail, depends on the expectation.

### References
#1431

### DIFF
--- a/t/mojo/tls.t
+++ b/t/mojo/tls.t
@@ -43,7 +43,7 @@ $delay  = $loop->delay;
 $server = Mojo::IOLoop::TLS->new($server_sock2)->reactor($loop->reactor);
 $server->once(upgrade => $delay->begin);
 $server->once(error   => sub { warn pop });
-$server->negotiate(server => 1, tls_ciphers => 'AES256-SHA:ALL', tls_version => 'TLSv1_2');
+$server->negotiate(server => 1, tls_ciphers => 'AES256-SHA:ALL');
 $client = Mojo::IOLoop::TLS->new($client_sock2)->reactor($loop->reactor);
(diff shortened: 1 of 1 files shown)


marked:
commit 5cd7ab4d8830d0bfd72ec5244cc881de9a469cf6
$ TEST_TLS=1 prove tls.t
tls.t .. ok
All tests successful.
Files=1, Tests=6,  0 wallclock secs ( 0.03 usr  0.00 sys +  0.25 cusr  0.02 csys =  0.30 CPU)
Result: PASS

marked:
> I think we should rather "upgrade" the test/code to be compatible with TLSv1.3.

My preference is 5cd7ab4 be checked in as the fix for bug #1431.

Then a new issue for how to get comprehensive coverage to simultaneously test TLS 1.2, 1.3, eventually 1.4 should be opened, and go through a design review.

Grinnz:
There will probably be many people without an openssl that supports TLS1.3, but since this test is gated by TEST_TLS it's probably fine. Travis's linux image does not appear to support it which is more of a problem.

kraih:
This PR seems to ignore the other TLS tests, which will fail.

kraih:
Thanks. Since this is a solution for #1431, i'll assume it is resolved now.

Issue #1473, opened 2020-02-19
Automated tests are passing when test build env is broken


marked:
* Mojolicious version: 8.34
* Perl version: 5.28, 5.12, 5.10
* Operating system: Travis-CI bionic dist

### Steps to reproduce the behavior
Build logs are already viewable here:
https://travis-ci.com/marked/mojo/builds/149593882

Or trigger new test build using a yml file set to bionic:
https://github.com/marked/mojo/blob/aabea1637a0f8d7ba9618110af6905e14730e5ea/.travis.yml

### Expected behavior
Automated tests should fail when test build environment is broken.
That is the environments for 5.28, 5.12, 5.10 are currently broken when using bionic.

### Actual behavior
Tests passed 3 times by picking up system perl with wrong version number, and skipping tests with missing CPAN modules.

marked:
Adding these lines to .travis.yml fixes it
```
before_install:
  - which perl | grep perlbrew
```

3x Builds now flagged as errored rather than passing:
https://travis-ci.com/marked/mojo/builds/150001306

Opening PR

kraih:
This report seems obsolete.

PR #1474, opened 2020-02-21
Add check to travis.yml to detect perlbrew failure

marked:
### Summary
Just verifies version of perl is from perlbrew when in Travis-CI environment

### Motivation
Travis-CI's perl environments has occasional bugs which bypasses perlbrew and this will cause tests to pass that should instead fail.

### References
#1473

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,10 +2,27 @@ language: perl
 matrix:
   include:
   - perl: "5.30"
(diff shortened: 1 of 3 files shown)


kraih:
Thanks. Since this is a solution for #1431, i'll assume it is resolved now.

PR #1478, opened 2020-03-03
Add lines to Mojo::File

melak:
### Summary
Add a `lines` method to Mojo::File, returning a Mojo::Collection of the lines in a file

### Motivation
Modeled after [Path::Tiny](https://metacpan.org/pod/Path::Tiny#edit_lines,-edit_lines_utf8,-edit_lines_raw)'s `lines` method (albeit with no knobs), this is a nice addition helping with fluid code when dealing with files on a line-by-line basis.

### References


kraih:
You can't implement a `lines` method without handling encodings.

PR #1480, opened 2020-03-09
Add plugin section to Tutorial guide

Grinnz:
### Summary
A tutorial section for using plugins in applications.

### Motivation
Using plugins is not currently covered anywhere in the guides in a general manner.

### DIFF
--- a/lib/Mojolicious/Guides/Tutorial.pod
+++ b/lib/Mojolicious/Guides/Tutorial.pod
@@ -342,6 +342,32 @@ actions to templates.
 A list of all built-in ones can be found in
 L<Mojolicious::Plugin::DefaultHelpers> and L<Mojolicious::Plugin::TagHelpers>.
 
+=head2 Plugins
+
+Plugins are application extensions that help with code sharing and
+organization. You can load a plugin with the keyword
+L<Mojolicious::Lite/"plugin">, which can omit the C<Mojolicious::Plugin::> part
+of the name, and optionally provide configuration for the plugin.
+
+  use Mojolicious::Lite;
+
+  plugin Config => {file => '/etc/myapp.conf', default => {foo => 'bar'}};
+
+  # Return configured foo value, or default if no configuration file
+  get '/foo' => sub {
+    my $c   = shift;
+    my $foo = $c->app->config('foo');
+    $c->render(json => {foo => $foo});
+  };
+
+  app->start;
+
+L<Mojolicious::Plugin::Config> is a built-in plugin which can populate
+L<Mojolicious/"config"> using a config file. Plugins can also set up routes,
+hooks, handlers, or even load other plugins. A list of built-in plugins can be
+found at L<Mojolicious::Plugins/"PLUGINS">, and many more are available from
+L<CPAN|https://metacpan.org/search?q=Mojolicious+Plugin>.
+
 =head2 Placeholders
 
 Route placeholders allow capturing parts of a request path until a C</> or C<.>


lib/Mojolicious/Guides/Tutorial.pod
```diff
@@ -342,6 +342,37 @@ actions to templates.
...
+  # Return configured foo value, or default if no configuration file
+  get '/foo' => sub {
+    my $c   = shift;
+    my $foo = $c->app->config('foo');
+    $c->render(json => {foo => $foo});
+  };
+
+  app->start;
+
+Some built-in plugins include L<Mojolicious::Plugin::DefaultHelpers>, which
```
kraih: Mentioning an automatically loaded plugin here is a bit misleading. I guarantee it will be misunderstood.

lib/Mojolicious/Guides/Tutorial.pod
```diff
@@ -342,6 +342,37 @@ actions to templates.
...
+adds many helpers and is loaded automatically, and
+L<Mojolicious::Plugin::Config>, which can populate L<Mojolicious/"config">
+using a config file. Plugins can also set up routes, hooks, handlers, or even
+load other plugins.
+
+A list of built-in plugins can be found at L<Mojolicious::Plugins/"PLUGINS">.
+Many more plugins are available from CPAN, and you can even write plugins
+specifically to help organize your application as described at
+L<Mojolicious::Guides::Cookbook/"Adding a plugin to your appplication">.
```
kraih: This is too verbose for a tutorial section. Just replace all of this with a link to [MetaCPAN](https://metacpan.org/requires/distribution/Mojolicious), like we have on the frontpage.

Grinnz:
Updated to be more concise

Grinnz:

lib/Mojolicious/Guides/Tutorial.pod
```diff
@@ -342,6 +342,32 @@ actions to templates.
 A list of all built-in ones can be found in
 L<Mojolicious::Plugin::DefaultHelpers> and L<Mojolicious::Plugin::TagHelpers>.
 
+=head2 Plugins
+
+Plugins are application extensions that help with code sharing and
+organization. You can load a plugin with the keyword
+L<Mojolicious::Lite/"plugin">, which can omit the C<Mojolicious::Plugin::> part
+of the name, and optionally provide configuration for the plugin.
```
jhthorsen: I think "which can omit the C<Mojolicious::Plugin::> part of the name" is out of place. Doesn't read very well to me. I think instead of mentioning which arguments `plugin()` can take, I would say *why* you want to write plugins. Something like "code sharing", "sharing logic between controllers", ...

Grinnz: It is necessary to explain why `plugin Config` loads `Mojolicious::Plugin::Config`.

Grinnz: If not documented here, I'd want to document it in either the Mojolicious::Lite/plugin or Mojolicious/plugin docs, neither of which currently contain this explanation of the arguments (just examples).

kraih: Every plugin worth using documents how it prefers to be loaded. I don't think it adds much value for a tutorial context.

Grinnz: That does not help with the other direction: seeing "plugin Config" in code and wanting to find out what that means.

kraih:
Guess i'm voting neutral.

lib/Mojolicious/Guides/Tutorial.pod
jhthorsen: > It is necessary to explain why `plugin Config` loads `Mojolicious::Plugin::Config`.

I think it's up to Mojolicious::Plugin::Config to document how it is loaded. I do agree that it's useful to have the rules mention in a general section as well, but I don't think this is the right place.

PR #1481, opened 2020-03-13
Clarify that WebSocket finish `reason` is a character string.

FGasper:
### Summary
A small clarification to the WebSocket documentation.

### Motivation
Because unlike with the other parts of the WebSocket docs, there’s no indication that the finish “reason” is expected to be a character string.

### DIFF
--- a/lib/Mojo/Transaction/WebSocket.pm
+++ b/lib/Mojo/Transaction/WebSocket.pm
@@ -237,7 +237,7 @@ Emitted once all data has been sent.
 =head2 finish
 
   $ws->on(finish => sub {
-    my ($ws, $code, $reason) = @_;
+    my ($ws, $code, $reason_chars) = @_;
     ...
   });
 
@@ -402,7 +402,7 @@ Connection identifier.
 
   $ws = $ws->finish;
   $ws = $ws->finish(1000);
-  $ws = $ws->finish(1003 => 'Cannot accept data!');
+  $ws = $ws->finish(1003 => do { use utf8; 'Can’t accept data!' });
 
 Close WebSocket connection gracefully.
 


kraih:
That's inconsistent and i find unicode punctuation rather confusing. 👎

FGasper:
@kraih Perhaps:
```
    my $chars = "Cannot accept data. \x{2639}";
    $ws = $ws->finish(1003 => $chars);
```
?

Is `$reason_chars` OK? (That’s the same approach used elsewhere in the WS docs.)

jhthorsen:
Sorry, but this PR does not clarify anything to me 👎

FGasper:
@jhthorsen Are you arguing that nothing needs clarification, or that what I submitted doesn’t suit?

kraih:
I'll have to reject this proposal, the use of the `utf8` pragma there is misleading. All Mojolicious code has it activated already.

FGasper:
@kraih The `utf8` pragma is a red herring. Sorry for my lack of clarity.

Nothing tells me, as a Mojo user, that `reason` is a UTF-8 decoded string. This is an important piece of information because, without that knowledge, I may think I have to UTF-8 decode the string in order to manipulate it—or, that I don’t have to encode it before outputting it, leading to wide character warnings.

The fact that Mojo’s code uses `utf8` internally doesn’t matter, and this is an area of ambiguity because Mojo’s WebSocket interface can accept/return either decoded or undecoded strings.

Shall I resubmit without the `utf8` in the example?

Grinnz:
We use the variable name `$chars` in examples to indicate character strings, as mentioned at https://metacpan.org/pod/Mojolicious::Guides#CONVENTIONS.

FGasper:
@Grinnz The documentation right now uses `$reason`, so I thought `$reason_chars` made sense.

PR #1489, opened 2020-04-03
fix typo in example test method

karenetheridge:
### Summary
A recent commit changed a variable name but one usage was missed.


kraih:
Thanks, this seems to have already been resolved.

Issue #1492, opened 2020-04-16
Split up inactivity_timeout into two timeouts for more granularity


kraih:
Everyone's web apps are different. Some of us need our servers to shut down more quickly, others need extra safety from connection close race conditions. Right now `Mojo::Server::Daemon` only has one `inactivity_timeout` setting that applies to connections with active transactions, as well as keep-alive connections without a transaction.

The default `inactivity_timeout` is `15` seconds, which works reasonably well. But many users end up setting the value much higher because their HTTP requests and WebSocket connections can take much longer. Probably using the global setting, affecting all connections. But even connection specific settings (like for WebSockets) are rarely reset, making server restarts very slow.

Now if we were to split up the global timeout settings into `inactivity_timeout` (for connections with active transactions) and `keep_alive_timeout` (for connections without active transactions), and then have `Mojo::Server::Daemon` switch automatically between the two, it could be a lot more user friendly.

I believe giving everyone two settings to tweak, might in many cases even eliminate the need for `$c->inactivity_timeout(300)`.

kraih:
It might be necessary to optimise the lower level bits of `Mojo::IOLoop::Stream::timeout` a bit for more frequent timeout changes (to avoid performance regressions).

kraih:
Testing this feature should be very simple though, just check the stream `inactivity_timeout` at various points during the request life cycle.

kraih:
I've made a branch with a possible `Mojo::IOLoop::Stream::timeout` optimisation. https://github.com/mojolicious/mojo/commit/ffedf7e3ac28

With patch applied:
```
$ perl -Ilib -Mojo -E 'a(sub ($c) { $c->inactivity_timeout(30); $c->render(text => "Hello Mojo!") })->start' daemon -m production
Server available at http://127.0.0.1:3000

$ wrk -c 20 -d 10 http://127.0.0.1:3000
Running 10s test @ http://127.0.0.1:3000
  2 threads and 20 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     9.66ms    0.96ms  23.77ms   93.89%
    Req/Sec     1.04k    43.88     1.11k    64.50%
  20761 requests in 10.05s, 3.05MB read
Requests/sec:   2066.38
Transfer/sec:    311.13KB
```

Without patch applied:
```
$ perl -Mojo -E 'a(sub ($c) { $c->inactivity_timeout(30); $c->render(text => "Hello Mojo!") })->start' daemon -m production
[...]

PR #1493, opened 2020-04-18
Separate keep-alive and inactivity timeouts for all built-in web servers

kraih:
This patch is preparation for #1492. It adds a low level way to reset timers, which is much more efficient than completely replacing timers every time you want to change a setting like the `inactivity_timeout`.

Baseline, this is pretty much equal with and without patch (EV):
```
$ perl -Ilib -Mojo -E 'a(sub ($c) { $c->render(text => "Hello Mojo!") })->start' daemon -m production
Server available at http://127.0.0.1:3000

$ wrk -c 20 -d 10 http://127.0.0.1:3000
Running 10s test @ http://127.0.0.1:3000
  2 threads and 20 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     9.46ms  735.87us  18.77ms   92.09%
    Req/Sec     1.06k    44.73     1.15k    69.50%
  21183 requests in 10.03s, 3.11MB read
Requests/sec:   2111.02
Transfer/sec:    317.85KB
````

With patch applied (EV):
```
$ perl -Ilib -Mojo -E 'a(sub ($c) { $c->inactivity_timeout(30); $c->render(text => "Hello Mojo!") })->start' daemon -m production
Server available at http://127.0.0.1:3000

$ wrk -c 20 -d 10 http://127.0.0.1:3000
Running 10s test @ http://127.0.0.1:3000
  2 threads and 20 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     9.66ms    0.96ms  23.77ms   93.89%
    Req/Sec     1.04k    43.88     1.11k    64.50%
  20761 requests in 10.05s, 3.05MB read
Requests/sec:   2066.38
Transfer/sec:    311.13KB
```

Without patch applied (EV):
```
$ perl -Mojo -E 'a(sub ($c) { $c->inactivity_timeout(30); $c->render(text => "Hello Mojo!") })->start' daemon -m production
Server available at http://127.0.0.1:3000

$ wrk -c 20 -d 10 http://127.0.0.1:3000
Running 10s test @ http://127.0.0.1:3000
  2 threads and 20 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency    10.16ms    0.91ms  20.89ms   93.22%
    Req/Sec     0.99k    41.46     1.10k    73.00%
  19736 requests in 10.04s, 2.90MB read
Requests/sec:   1965.20
Transfer/sec:    295.88KB
```

With patch applied (Poll):
```
$ MOJO_REACTOR=Mojo::Reactor::Poll perl -Ilib -Mojo -E 'a(sub ($c) { $c->inactivity_timeout(30); $c->render(text => "Hello Mojo!") })->start' daemon -m production
Server available at http://127.0.0.1:3000

$ wrk -c 20 -d 10 http://127.0.0.1:3000
Running 10s test @ http://127.0.0.1:3000
  2 threads and 20 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     9.92ms    0.92ms  25.57ms   93.01%
    Req/Sec     1.01k    47.22     1.11k    74.50%
  20207 requests in 10.04s, 2.97MB read
Requests/sec:   2012.91
Transfer/sec:    303.09KB
```

Without patch applied (Poll):
```
$ MOJO_REACTOR=Mojo::Reactor::Poll perl -Mojo -E 'a(sub ($c) { $c->inactivity_timeout(30); $c->render(text => "Hello Mojo!") })->start' daemon -m production
Server available at http://127.0.0.1:3000

$ wrk -c 20 -d 10 http://127.0.0.1:3000
Running 10s test @ http://127.0.0.1:3000
  2 threads and 20 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency    10.15ms    0.87ms  20.10ms   92.19%
    Req/Sec     0.99k    38.45     1.10k    78.00%
  19723 requests in 10.03s, 2.90MB read
Requests/sec:   1967.33
Transfer/sec:    296.20KB
```

### DIFF
--- a/lib/Mojo/IOLoop/Stream.pm
+++ b/lib/Mojo/IOLoop/Stream.pm
@@ -72,16 +72,24 @@ sub stop {
 }
 
 sub timeout {
-  my $self = shift;
+  my ($self, $timeout) = @_;
 
-  return $self->{timeout} unless @_;
+  return $self->{timeout} unless defined $timeout;
+  $self->{timeout} = $timeout;
 
   my $reactor = $self->reactor;
-  $reactor->remove(delete $self->{timer}) if $self->{timer};
-  return $self unless my $timeout = $self->{timeout} = shift;
-  weaken $self;
-  $self->{timer} = $reactor->timer(
-    $timeout => sub { delete $self->{timer}; $self->emit('timeout')->close });
+  if ($self->{timer}) {
+    if (!$self->{timeout}) { $reactor->remove(delete $self->{timer}) }
+    else                   { $reactor->again($self->{timer}, $self->{timeout}) }
+  }
+  else {
+    weaken $self;
+    $self->{timer} = $reactor->timer(
+      $timeout => sub {
+        $self and delete($self->{timer}) and $self->emit('timeout')->close;
+      }
+    );
+  }
 
   return $self;
 }
--- a/lib/Mojo/Reactor.pm
+++ b/lib/Mojo/Reactor.pm
@@ -87,9 +87,10 @@ the following new ones.
 =head2 again
 
   $reactor->again($id);
+  $reactor->again($id, 0.5);
 
-Restart timer. Meant to be overloaded in a subclass. Note that this method
-requires an active timer.
+Restart timer and optionally change the invocation time. Meant to be overloaded
+in a subclass. Note that this method requires an active timer.
 
 =head2 detect
 
--- a/lib/Mojo/Reactor/EV.pm
+++ b/lib/Mojo/Reactor/EV.pm
@@ -9,8 +9,10 @@ my $EV;
 sub DESTROY { undef $EV }
 
 sub again {
-  croak 'Timer not active' unless my $timer = shift->{timers}{shift()};
-  $timer->{watcher}->again;
+  my ($self, $id, $after) = @_;
+  croak 'Timer not active' unless my $timer = $self->{timers}{$id};
+  my $w = $timer->{watcher};
+  defined $after ? $w->set($after, $w->repeat ? $after : 0) : $w->again;
 }
 
 # We have to fall back to Mojo::Reactor::Poll, since EV is unique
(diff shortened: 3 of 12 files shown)


kraih:

kraih:
Instead of adding a new API, i've just reused `$reactor->again($id)` for `$reactor->again($id, $seconds)`. It does raise the bar a little bit for 3rd party reactors, but being able to split `inactivity_timeout` into `inactivity_timeout` and `keep_alive_timeout` is probably worth it.

Grinnz:
I don't mind updating my reactor backends to support this, but what's the practical impact if someone updates Mojolicious without updating the reactor backend they're using?

kraih:
Actually, i've amended the PR with the `keep_alive_timeout` feature since it was really easy now. Thanks to the optimisation is costs pretty much nothing and i can't measure a difference with the micro benchmark above.

kraih:
@Grinnz Since i've never used an alternative reactor backend i have no idea.

marcusramberg:
I'm voting :+1:

kraih:
To be clear, this is actually an important feature that i strongly believe should be added. It addresses fairly common problems with the current inactivity timeout.

To give one example i ran into at work. There we had a micro service running in prefork mode with 12 workers. Because the service could run fairly long SQL queries we had to limit concurrency (one request per worker) and increase the inactivity timeout a bit. So the service could run 12 requests at a time. Now unexpectedly we suddenly had more than 12 clients, and clients were configured to use keep-alive. And boom, everything broke because idle keep-alive connections blocked all the workers.

There are many more such scenarios. With this feature i could simply increase the inactivity timeout to say 300 (for very very slow SQL queries) and set the keep-alive timeout to 5. Letting idle connections get closed very quickly if there are not multiple requests in quick succession.

jberger:
I missed the vote with a busy weekend but I'll still hop in here with a 👍

PR #1494, opened 2020-04-18
Allow curfile to work in Perl Config files

haukex:
### Summary
Allows `__FILE__`, `__LINE__`, and `Mojo::File::curfile` to work correctly in Perl config files.

### Motivation
Allows one to create filenames relative to the current configuration file in the config. For example:

```
use Mojo::File 'curfile';
{
    migrations_file => curfile->sibling('foo.sql'),
}
```

### DIFF
--- a/lib/Mojolicious/Plugin/Config.pm
+++ b/lib/Mojolicious/Plugin/Config.pm
@@ -10,14 +10,24 @@ sub parse {
   my ($self, $content, $file, $conf, $app) = @_;
 
   # add the "#line" directive *only* if it's safe to do so
-  my $line_directive = $file=~/[\r\n]/ ? '' # can't break "#line" across lines
-    : ( $file=~/\s/ ? ( $file=~/"/ ? ''     # can't escape double-quotes
-      : qq{\n#line 1 "$file"\n} ) : qq{\n#line 1 $file\n} );
+  my $line_directive = $file =~ /[\r\n]/
+    ? ''    # can't break "#line" across lines
+    : (
+    $file =~ /\s/
+    ? (
+      $file =~ /"/
+      ? ''    # can't escape double-quotes
+      : qq{\n#line 1 "$file"\n}
+      )
+    : qq{\n#line 1 $file\n}
+    );
 
   # Run Perl code in sandbox
-  my $config = eval 'package Mojolicious::Plugin::Config::Sandbox; no warnings;'
+  my $config
+    = eval 'package Mojolicious::Plugin::Config::Sandbox; no warnings;'
     . "sub app; local *app = sub { \$app }; use Mojo::Base -strict;"
-    . $line_directive . $content;
+    . $line_directive
+    . $content;
   die qq{Can't load configuration from file "$file": $@} if $@;
   die qq{Configuration file "$file" did not return a hash reference.\n}
     unless ref $config eq 'HASH';
--- a/lib/Mojolicious/Plugin/Config.pm
+++ b/lib/Mojolicious/Plugin/Config.pm
@@ -9,25 +9,11 @@ sub load { $_[0]->parse(decode('UTF-8', path($_[1])->slurp), @_[1, 2, 3]) }
 sub parse {
   my ($self, $content, $file, $conf, $app) = @_;
 
-  # add the "#line" directive *only* if it's safe to do so
-  my $line_directive = $file =~ /[\r\n]/
-    ? ''    # can't break "#line" across lines
-    : (
-    $file =~ /\s/
-    ? (
-      $file =~ /"/
-      ? ''    # can't escape double-quotes
-      : qq{\n#line 1 "$file"\n}
-      )
-    : qq{\n#line 1 $file\n}
-    );
-
   # Run Perl code in sandbox
-  my $config
-    = eval 'package Mojolicious::Plugin::Config::Sandbox; no warnings;'
-    . "sub app; local *app = sub { \$app }; use Mojo::Base -strict;"
-    . $line_directive
-    . $content;
+  my $config = eval 'package Mojolicious::Plugin::Config::Sandbox; no warnings;'
+    . "sub app; local *app = sub { \$app }; use Mojo::Base -strict;\n#line 1"
+    # add the file to "#line" directive *only* if it's safe to do so
+    . ($file =~ /[\n"]/ ? '' : qq{ "$file"}) . "\n" . $content;
   die qq{Can't load configuration from file "$file": $@} if $@;
   die qq{Configuration file "$file" did not return a hash reference.\n}
     unless ref $config eq 'HASH';
@@ -125,6 +111,10 @@ will be detected automatically.
 
 If the configuration value C<config_override> has been set in
 L<Mojolicious/"config"> when this plugin is loaded, it will not do anything.
+
+The C<__LINE__> and C<__FILE__> directives, and related things such
+as L<Mojo::File/"curfile">, I<won't> work if the filename contains
+double quotes (C<">) or newlines (C<\n>).
 
 The code of this plugin is a good example for learning to build new plugins,
 you're welcome to fork it.
--- a/lib/Mojolicious/Plugin/Config.pm
+++ b/lib/Mojolicious/Plugin/Config.pm
@@ -9,11 +9,15 @@ sub load { $_[0]->parse(decode('UTF-8', path($_[1])->slurp), @_[1, 2, 3]) }
 sub parse {
   my ($self, $content, $file, $conf, $app) = @_;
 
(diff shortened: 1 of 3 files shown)


lib/Mojolicious/Plugin/Config.pm
```diff
@@ -11,7 +11,7 @@ sub parse {
 
   # Run Perl code in sandbox
   my $config = eval 'package Mojolicious::Plugin::Config::Sandbox; no warnings;'
-    . "sub app; local *app = sub { \$app }; use Mojo::Base -strict; $content";
+    . "sub app; local *app = sub { \$app }; use Mojo::Base -strict;\n#line 1 $file\n$content";
```
kraih, line 14: Is that line longer than 80 chars?

Grinnz:
I'm not totally against this, but the #line hack can be very finicky and I don't think this is really necessary. It's just as easy to store paths relative to your app home dir in the config file, and have the reader append it to the homedir.

t/mojolicious/perl_config_lite_app.conf
```diff
@@ -0,0 +1,8 @@
+use Mojo::File 'curfile';
+{
+	# Just a comment
+	foo => "bar",
+	utf => "утф",
+	file => curfile->to_string,
+	line => __LINE__, # 7
+}
```
kraih, line 8: This file doesn't follow our formatting rules.

t/mojolicious/perl_config_lite_app.t
```diff
@@ -0,0 +1,21 @@
+use Mojo::Base -strict;
+
+BEGIN {
+  $ENV{MOJO_MODE}    = 'development';
+  $ENV{MOJO_REACTOR} = 'Mojo::Reactor::Poll';
+}
```
kraih, line 6: Why is that necessary for the test?

t/mojolicious/perl_config_lite_app.t
```diff
@@ -0,0 +1,21 @@
...
+use Mojo::File qw(curfile);
+use Mojolicious::Lite;
+
+plugin 'Config';
+is_deeply app->config, {
+		foo => "bar",
+		utf => "утф",
+		file => curfile->sibling('perl_config_lite_app.conf')->to_string,
+		line => 7,
+	}, 'right value';
```
kraih, line 19: That also does not look like it follows our formatting rules.

lib/Mojolicious/Plugin/Config.pm
haukex: Fixed

t/mojolicious/perl_config_lite_app.conf
haukex: If you mean indents as two space chars, I fixed that. Anything else?

t/mojolicious/perl_config_lite_app.t
haukex: Cargo-culted from `json_config_lite_app.t`; I removed it.

t/mojolicious/perl_config_lite_app.t
haukex: Fixed as above

haukex:
@Grinnz
> the #line hack can be very finicky

Hm, if the filename contains a newline, that might be a problem, so I "fixed" that in the updated commit (not pretty...). Is there anything else?

haukex:
I was surprised by the failure on Windows:
```
#          $got->{file} = 'C:/projects/mojo/t/mojolicious/perl_config_lite_app.conf'
#     $expected->{file} = 'C:\projects\mojo\t\mojolicious\perl_config_lite_app.conf'
```
I hope my attempted fix of adding `->to_abs` works.

Grinnz:
Spaces in the filename also break the #line directive, but this will just make `__FILE__` not work in that case and not break anything else.

Grinnz:
Spaces in a filename work if it's double quoted, but then double quotes in the filename don't work. It's really just not well designed for generated code.

haukex:
@Grinnz 
> Spaces in the filename also break the #line directive, but this will just make `__FILE__` not work in that case and not break anything else.

> Spaces in a filename work if it's double quoted, but then double quotes in the filename don't work. It's really just not well designed for generated code.

Hrm, that is indeed not great. On the one hand, I could just add quotes around the filename and hope nothing breaks - but on the other hand, I guess the "defensive" way to code it would be to not include the `#line` directive if it's going to break, so that's what I've done in the latest commit. Still not super pretty, unfortunately...

> It's just as easy to store paths relative to your app home dir in the config file, and have the reader append it to the homedir.

Background for why I suggested this in the first place: I'd like to use the same configuration in a non-Mojo script by `eval`ing it, and it may be in a different path. (And of course adding the `#line` makes error messages nicer.)

t/mojolicious/perl_config_lite_app.conf
jhthorsen: @haukex: You should run `perltidy` on all the files instead of guessing what to change. See #1370 for details, unless you want to install something like https://metacpan.org/pod/App::githook::perltidy

haukex: Thanks, I had missed the `.perltidyrc`.

lib/Mojolicious/Plugin/Config.pm
```diff
@@ -9,9 +9,25 @@ sub load { $_[0]->parse(decode('UTF-8', path($_[1])->slurp), @_[1, 2, 3]) }
...
+    ? ''    # can't break "#line" across lines
+    : (
+    $file =~ /\s/
+    ? (
+      $file =~ /"/
+      ? ''    # can't escape double-quotes
+      : qq{\n#line 1 "$file"\n}
+      )
+    : qq{\n#line 1 $file\n}
+    );
```
kraih, line 23: This is really ugly code.

haukex: It's the same code as before, but run through perltidy, as per the Rules in Mojolicious::Guides::Contributing. I could rewrite it as `if` statements for example - what is your preference?

kraih: I don't mean formatting alone. The code also creates lots of untested special cases. Unless the whole feature can be reduced to say one line, i doubt i would vote for it.

haukex:
I looked at Perl's [code](https://github.com/Perl/perl5/blob/8100f3cd3c45f1190105982b7fed1028074526ae/toke.c#L1775) and indeed there is no way to escape double quotes in `#line`, but it looks like any other characters (except `\n`) should work.

So, I've **simplified** the code changes accordingly, added **more tests** to cover the special cases, and added a line in the **documentation**.

I feel like I've done everything I can for now - I hope this passes muster.

lib/Mojolicious/Plugin/Config.pm
haukex: Ah, I see, and yes, [I agree](https://github.com/mojolicious/mojo/pull/1494#issuecomment-615942450). I've made some updates to attempt to address it, [see my comment below](https://github.com/mojolicious/mojo/pull/1494#issuecomment-623005604).

kraih:
While i might vote in favour of the feature itself if done in the most minimalistic way possible, there is a lot of special cases in this PR i do not agree with. So 👎 from me on this specific implementation of the feature.

kraih:
Or to be more precise, there is already `#line` use that should be emulated in `Mojo::Template`.

haukex:
@kraih 
Yes, originally it did contain more special cases than necessary (the commits should be squash merged, or I could open a new PR if desired), but now the only thing that's left is `\n` and `"`, just like in `Mojo::Template` - or do you mean other special cases?
I'm not sure I understand your comment in regards to `#line` in `Mojo::Template`, are you suggesting code re-use between the two?

jberger:
I personally don't love that the file is evaluated at all and I've been tending to use JSON and YAML config wherever possible. Meanwhile if your case for it is

> Background for why I suggested this in the first place: I'd like to use the same configuration in a non-Mojo script by evaling it, and it may be in a different path.

then I really disagree. My understanding of the reasoning for evaluating was to give access to the app itself. In a non-mojo context, the app having no meaning, you're using a config file as a generic runnable file. If you need that I'd suggest simply doing so, `do $file` the file on your own and load it into `$app->config` at startup. The config loader module isn't magic (apart from running the file), it essentially does the same thing. 👎  from me unfortunately

marcusramberg:
I'm also against adding this functionality to the config. 👎

haukex:
Thanks for the feedback everyone.

kraih:
In case someone stumbles over this in the future, the current solution is:
```
{
    migrations_file => app->home->child('foo.sql'),
}
```
Without the need for imports.

PR #1500, opened 2020-04-24
Mojo::DOM - doc patch, link to all_text, from text.

MartinMcGrath:
### Summary
This is a short PR to add a link to the `all_text` method, from `text`, explaining that `all_text` should be used to extract text content from all descendant nodes. 

### Motivation
Simply to make it a little clearer to the reader, perhaps as yet unfamiliar with the framework who may have missed the already explicit documentation, and providing a helpful link to the `all_text` method. 

### References

https://perlmonks.org/?node_id=11115910


jberger:
Personally I like cross-linking in documentation, but I don't think we've done much of it for the Mojo class-level docs.

I'm inclined to vote +1 by utility and -1 by precedent. Not sure where that leaves me. Neutral I guess.

kraih:
@jberger For precedent there is existing cases like [param](https://mojolicious.org/perldoc/Mojolicious/Controller#param) and  [every_param](https://mojolicious.org/perldoc/Mojolicious/Controller#every_param).

PR #1501, opened 2020-04-29
Always use dash instead of underscore for commands

kraih:
Simple proposal. All commands now look like `mojo generate lite-app` instead of `mojo generate lite_app`. And the command list will also only show the version with dash. Both variants still work though.
```
$ mojo generate
Usage: APPLICATION generate GENERATOR [OPTIONS]

  mojo generate app
  mojo generate lite-app

Generators:
 app       Generate Mojolicious application directory structure
 lite-app  Generate Mojolicious::Lite application
 makefile  Generate "Makefile.PL"
 plugin    Generate Mojolicious plugin directory structure

See 'APPLICATION generate help GENERATOR' for more information on a specific
generator.
```


kraih:

jberger:
I'm glad that both versions still work. Dashes look nice. 👍

CandyAngel:
Just for the record, I vote against.

jhthorsen:
> Just for the record, I vote against.

Why do you vote against @CandyAngel?

I'm neutral. It looks better, but I'm not sure if it's worth the change. (Not opposed though)

CandyAngel:
My opposition is that this introduces complexity, change and risk for no/little gain.

Existing documentation (tutorials, blog posts, user documentation) now look different from what Mojolicious outputs (confusing for newbies). You can use dashes.. except for the first character (more edge cases to consider)[1]. If a problem with this is discovered and it gets reverted, that will then be a breaking change.

I would say I would prefer it if dashes worked as well (almost like a hidden feature for power users?), but that still means that what gets written in user documentation could depend on the author preference, rather than the "actual way" Mojolicious uses commands (which is neither complicated from a developer point of view, because it's just a module name, or from an end user point of view, because it's shown to you).

Normally these things are balanced out by the improvement but it doesn't seem worth it to me for a minor cosmetic change.

[1] Case in point, the PR title is wrong, because it doesn't "always use a dash instead of underscore"

macsnowball:
Hello.  I've been lurking on the sidelines of the project for a while, as I struggle to muster the familiarity, knowledge, ability and courage to effectively contribute.  So I realise I'm not qualified to weigh in on this, except to put forward the perspective of the newcomer, whose plight seems to form the bulk of the argument against the change.

I believe the use of hyphens appears more modern / standardised, and in my experience suits general taste more widely. It's adoption might not win any new followers, but its absence might deter them...

Personally, I was disappointed when I first encountered underscores, considering it an indicator of style for the project as a whole, and foreseeing the frustration of frequent typos in my future.

During the early stages of researching and thinning down alternative tools, such trivialities can count for more to the newcomer than they do for someone already convinced of the tool's merit (as with website snazz).  As the number of alternatives in the marketplace increase, the filtering criteria in the early stages become increasingly superficial.  For this reason, I would count this amongst those barriers to adoption that alone might not deter a new user but that in combination with others very well might.

I hope it isn't out of place for me to volunteer the above.

PR #1502, opened 2020-05-01
Exclude <script> and <style> from Mojo::DOM text extraction

kraih:
This came up on IRC recently, and everyone seemed to agree that excluding `<script>` and `<style>` would be the correct behaviour for `Mojo::DOM::all_text`.

### DIFF
--- a/lib/Mojo/DOM.pm
+++ b/lib/Mojo/DOM.pm
@@ -337,7 +337,10 @@ sub _text {
     }
 
     # Nested tag
-    elsif ($type eq 'tag' && $all) { unshift @$nodes, @{_nodes($node)} }
+    elsif ($type eq 'tag' && $all) {
+      unshift @$nodes, @{_nodes($node)}
+        unless $node->[1] eq 'script' || $node->[1] eq 'style';
+    }
   }
 
   return $text;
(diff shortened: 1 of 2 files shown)


kraih:

shadowcat-mst:
< kraih> everyone who agreed that it's the right behaviour better vote

Well, OK, since you insist.

@shadowcat-mst votes Monster Raving Loony Party

(also +1 for what little it's worth as a non-core contrib but I can't quite bring myself to spoil my ballot entirely)

marcusramberg:
I'm neutral to this change.

CandyAngel:
I think this would be better if the list of attributes to omit from `all_text` processing was an attribute. `script` and `style` seem like good defaults and it would allow for easy extension by the user (like if they had custom tags in their DOM) or clearing if they want the existing behaviour.

kraih:
Quick disclaimer, i'm not going to modify this proposal. It is up for vote as is, and anyone is welcome to reuse the code for alternative proposals.

Grinnz:
@CandyAngel This sort of more extensible method could pretty easily be provided by a role or so, just possibly less efficient.

kraih:
This change also does not prevent another PR with support for `$dom->all_text({exclude => ['script', 'style']})` in the future.

CandyAngel:
This seems like a useful change so I'm +1 on it, I just want it to be user controllable rather than hard-coded, which can be added in the future.

jberger:
While in concept I agree with the spirit of this change, in xml mode should tags like script and style handled differently like this?

Grinnz:
I agree, this exclusion should not occur in XML mode.

kraih:
Since @Grinnz voted +1, yet raised concerns, i consider this vote failed. I will probably not be working on another attempt. So feel free to reuse whatever you want from this PR.

PR #1506, opened 2020-05-04
Proposal: Add "walk" method to Mojo::DOM

haukex:
### Summary
This new `walk` method is similar in spirit to `Mojo::Collection`'s `map`, except that this method walks the tree of DOM nodes depth-first, invoking a callback or a method for each node, and (optionally) building a nested set of `Mojo::Collection`s of the return values of the callback.

### Motivation
Walking a tree of nodes is, in my experience, a pretty common task - I show several possible uses in the POD. The `walk` method makes implementing this task much easier than if the user had to implement the recursion themselves. Since it can be implemented in roughly 10 lines of code (`_walk_void` is just a version of `_walk` optimized for void context, it is optional), I hope it's not too expensive.

### Example
```perl
use Mojo::Base -strict;
use Test::More tests => 1;
use Mojo::Collection 'c';
use Mojo::DOM;

my $dom = Mojo::DOM->new('<html><head><title>Testing</title></head>'
  .'<body><h1>Example <i>One</i></h1></body></html>');

my $texts = $dom->walk(sub {
  $_->type =~ /^(text|cdata|raw)$/ ? $_->content : $_->tag//$_->type });

is_deeply $texts,
  c( 'root',
    c( 'html',
      c( 'head',
        c( 'title', c('Testing') ),
      ),
      c( 'body',
        c( 'h1', c('Example '), c( 'i', c('One') ) ),
      )
    )
  );
```

### DIFF
--- a/lib/Mojo/DOM.pm
+++ b/lib/Mojo/DOM.pm
@@ -198,7 +198,6 @@ sub val {
 
 sub walk {
   my $self = shift;
-  if (not defined wantarray) { $self->_walk_void(0, @_); return }
   $self->_walk(0, @_) || Mojo::Collection->new;
 }
 
@@ -356,12 +355,6 @@ sub _walk {
     map { $_->_walk($depth+1, $cb, @_) } @{$self->child_nodes}
   );
   @$c ? $c : ();    # prune empty subtrees
-}
-
-sub _walk_void {    # roughly 10% faster than _walk
-  my ($self, $depth, $cb) = (shift, shift, shift);
-  $_->$cb(ref $cb ? ($depth, @_) : @_) for $self;
-  $_->_walk_void($depth+1, $cb, @_) for @{$self->child_nodes};
 }
 
 sub _wrap {
--- a/lib/Mojo/DOM.pm
+++ b/lib/Mojo/DOM.pm
@@ -196,6 +196,12 @@ sub val {
   return exists $self->{multiple} ? $v->size ? $v->to_array : undef : $v->last;
 }
 
+sub walk {
+  my $self = shift;
+  if (not defined wantarray) { $self->_walk_void(0, @_); return }
+  $self->_walk(0, @_) || Mojo::Collection->new;
+}
+
 sub with_roles { shift->Mojo::Base::with_roles(@_) }
 
 sub wrap         { shift->_wrap(0, @_) }
@@ -341,6 +347,21 @@ sub _text {
   }
 
   return $text;
+}
+
+sub _walk {
+  my ($self, $depth, $cb) = (shift, shift, shift);
+  my $c = Mojo::Collection->new(
+    (map { $_->$cb(ref $cb ? ($depth, @_) : @_) } $self),
+    map { $_->_walk($depth+1, $cb, @_) } @{$self->child_nodes}
+  );
+  @$c ? $c : ();    # prune empty subtrees
+}
+
+sub _walk_void {    # roughly 10% faster than _walk
+  my ($self, $depth, $cb) = (shift, shift, shift);
+  $_->$cb(ref $cb ? ($depth, @_) : @_) for $self;
+  $_->_walk_void($depth+1, $cb, @_) for @{$self->child_nodes};
 }
 
 sub _wrap {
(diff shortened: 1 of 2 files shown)


Grinnz:
Semantically it seems a little confusing to me to have a method called "walk" return something.

haukex:
@Grinnz 
Thanks for the feedback! The user-visible behavior doesn't change at all based on context, it's really just an internal optimization to save on needlessly building a bunch of collections. As for the semantics, sure, maybe there's a better name... (I initially considered whether it should *just* run the callback and return `$self`, but IMHO also being able to build data structures is really useful thing.)

haukex:
That's fine, as I said it's just an optional optimization (~10% speedup), I've removed it.

t/mojo/dom_walk.t
```diff
@@ -0,0 +1,168 @@
...
+  ], 'text nodes, nested, no ws' or diag explain $struct_texts;
+
+if (0) {
+  diag "Running benchmark...";
+  require Benchmark;
+  Benchmark::cmpthese(-3, {
+         void => sub {         $dom->walk(sub { $_->type }); 1 },
+      nonvoid => sub { my $x = $dom->walk(sub { $_->type })    },
+    });
+}
```
kraih, line 166: You forgot to remove test code.

t/mojo/dom_walk.t
```diff
@@ -0,0 +1,168 @@
...
+      [                                   # body
+        [['Example']],                    # h1
+        [
+          ['Foo'],                        # p
+          [['Bar!']],                     # button
+        ],
+        [['Quz']],                        # div
+      ],
+    ],
+  ], 'text nodes, nested, no ws' or diag explain $struct_texts;
```
kraih, line 157: Why is this not a bunch of subtests in `t/mojo/dom.t`?

lib/Mojo/DOM.pm
```diff
@@ -343,6 +348,15 @@ sub _text {
   return $text;
 }
 
+sub _walk {
+  my ($self, $depth, $cb) = (shift, shift, shift);
+  my $c = Mojo::Collection->new(
+    (map { $_->$cb(ref $cb ? ($depth, @_) : @_) } $self),
+    map { $_->_walk($depth+1, $cb, @_) } @{$self->child_nodes}
+  );
+  @$c ? $c : ();    # prune empty subtrees
```
kraih, line 357: We don't use implicit `return` for anything else than one-liners.

lib/Mojo/DOM.pm
```diff
@@ -196,6 +196,11 @@ sub val {
   return exists $self->{multiple} ? $v->size ? $v->to_array : undef : $v->last;
 }
 
+sub walk {
+  my $self = shift;
+  $self->_walk(0, @_) || Mojo::Collection->new;
+}
```
kraih, line 202:
```perl
sub walk { shift->_walk(0, @_)) || Mojo::Collection->new }
```

lib/Mojo/DOM.pm
```diff
@@ -1046,6 +1060,42 @@ if none could be found.
...
+    if ( $_->type=~/^(text|cdata|raw)$/ && $_->content=~/\S/ ) {
+      return $_->content =~ s/^\s+|\s+$//gr
+    } else { return }
+  });
+
+  # prints the contents of text nodes in the body, indented by their depth
+  $dom->at('body')->walk(sub {
+    my ($node, $depth) = @_;
+    say "  " x $depth, $node->content if $node->type eq 'text';
+  });
```
kraih, line 1080: The documentation format is inconsistent with other methods in the same class.

kraih:
Personally i don't see much use in this feature, so i tend towards 👎.

haukex:
@kraih
Thanks, I'd be happy to make all the changes you suggested if the feature has a chance of being merged. Over on PerlMonks, we've been answering a lot of "how do I parse XML/HTML" questions with Mojo::DOM, it's a really nice module on its own. So I thought it would be good to have such a "general-purpose" tree walking feature in the module, it would have been useful in helping at least one or two people (in fact, the code derives directly from [this thread](https://www.perlmonks.org/?node_id=11116409)); since it's only 8 lines of code it felt a little wasteful to put it in a role, hence my suggestion here.

jberger:
I think I tend to agree with @kraih, I don't think I've ever felt the need for this method and further if I did, I can imagine I'd need something particular to the case. Is passing in the depth generically useful to most cases? Would some other argument?

Certainly it seems useful for some cases and as it doesn't need access to core details I think it is a good candidate for a Role on cpan, even if it is initially a small one. I say "initially" because I think if it does get traction you'll see requests for slight variations. This kind of shaking out on CPAN can always lead to core inclusion later (see Subprocess as an example).

kraih:
Actually thinking some more about the shown use cases, they don't make sense to me. The first might just as well use `child_nodes`, and the second indents text based on the nesting level of the tag (can't think of any reason to want that).

haukex:
I personally have had the need to walk DOM trees when parsing HTML and XML, but if you disagree, that's fine. Thanks for the consideration.

Issue #1454, opened 2019-12-28
Make unhandled rejected promises warn


kraih:
Opening this issue so we don't forget. Similar to JavaScript promises we want rejected `Mojo::Promise` objects to warn if they were not handled before the object is destroyed. One small complication is that `await $promise` also counts as handling the promise, not just rejection handlers.

kraih:
For example:
```perl
# Handled
Mojo::Promise->reject('test')->catch(sub {...});

# Handled
my $promise = Mojo::Promise->reject('test');
$promise->catch(sub {...});

# Handled
my $promise = Mojo::Promise->reject('test');
await $promise;

# Unhandled (needs to result in a warning)
Mojo::Promise->reject('test');

# Unhandled (needs to result in a warning)
my $promise = Mojo::Promise->reject('test');
undef $promise;
```

kraih:
This [branch](https://github.com/mojolicious/mojo/compare/unhandled_promises) is almost enough to make it work, but there are still a few tests left that generate a warning and shouldn't.
```
$ prove -l t/mojo/promise.t 
t/mojo/promise.t .. 1/? Unhandled rejected promise: second
Unhandled rejected promise: second
Unhandled rejected promise: second
Unhandled rejected promise: ARRAY(0x7fef7ee9ee78) ARRAY(0x7fef7eeb9178) ARRAY(0x7fef7eeb9808)
Unhandled rejected promise: Timeout2
Unhandled rejected promise: third
Unhandled rejected promise: 1
Unhandled rejected promise: 1
Unhandled rejected promise: 1
Unhandled rejected promise: 1
t/mojo/promise.t .. ok    
All tests successful.
Files=1, Tests=65,  1 wallclock secs ( 0.02 usr  0.01 sys +  0.17 cusr  0.01 csys =  0.21 CPU)
Result: PASS
```

kraih:
After https://github.com/mojolicious/mojo/compare/0265b632bf60...19b00feb621c there is just two more warnings left to fix.
```
$ prove -l t/mojo/promise.t 
t/mojo/promise.t .. 1/? Unhandled rejected promise: 1
t/mojo/promise.t .. ok    
All tests successful.
Files=1, Tests=65,  1 wallclock secs ( 0.02 usr  0.00 sys +  0.16 cusr  0.01 csys =  0.19 CPU)
Result: PASS
```
```
$ TEST_ASYNC_AWAIT=1 prove -l t/mojo/promise_async_await.t 
[...]

PR #1510, opened 2020-05-17
Pr warnings in promise tests

akarelas:
My attempt to fix the warnings of promise.t (described here: https://github.com/mojolicious/mojo/issues/1454#issuecomment-629848828)

### References
#1454

### DIFF
--- a/lib/Mojo/Promise.pm
+++ b/lib/Mojo/Promise.pm
@@ -23,6 +23,7 @@ sub AWAIT_IS_CANCELLED {undef}
 
 sub AWAIT_IS_READY {
   my $self = shift;
+  $self->{handled} = 1;
   return !!$self->{result} && !@{$self->{resolve}} && !@{$self->{reject}};
 }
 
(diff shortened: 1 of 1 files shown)


lib/Mojo/Promise.pm
```diff
@@ -163,10 +165,12 @@ sub _all {
         sub {
           $results->[$i] = {status => 'fulfilled', value => [@_]};
           $all->resolve(@$results) if --$remaining <= 0;
+          ();
```
akarelas, line 168: No warnings were emitted before adding this line, but still thought it could be a good idea to include.

lib/Mojo/Promise.pm
```diff
@@ -163,10 +165,12 @@ sub _all {
...
         },
         sub {
           $results->[$i] = {status => 'rejected', reason => [@_]};
           $all->resolve(@$results) if --$remaining <= 0;
+          ();
```
akarelas, line 173: No warnings were emitted before adding this line, either.

PR #1513, opened 2020-05-19
Restore promise map

jberger:
### Summary
Restore Mojo::Promise::map and improve where there were concerns

### Motivation
Because it was mostly ok, just needed some tweaks, and it is very useful

I have improved the test case to fail when the $proto promise wasn't cloned, however, I have then removed the $proto promise all together as an attempt to prevent warnings as we attempt to warn on unhandled rejected promises.

### DIFF
--- a/Changes
+++ b/Changes
@@ -1,7 +1,5 @@
 
 8.43  2020-05-18
-  - Removed experimental map method from Mojo::Promise, since its test coverage
-    was lacking, getting in the way of new feature development.
   - Removed deprecated argument handling from Mojo::Promise::new.
   - Fixed a bug where the resolve class method in Mojo::Promise would
     unnecessarily create new promises.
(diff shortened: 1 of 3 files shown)


jberger:
Note that it might be easier to review the comparison from before map was removed (and is restored in this PR): https://github.com/mojolicious/mojo/compare/3ca458038..restore_promise_map

jberger:
I'll also note that the weaken on the $loop is only needed because we close over $loop to return a rejection [here](https://github.com/mojolicious/mojo/compare/3ca458038..restore_promise_map#diff-27bd5e0ace07b04f2ba7f5e7ec0115c1R80). If "die" would work to "rethrow" the error then that weaken could be removed.

kraih:
It does not look good to me, `weaken $loop` is  a pretty ugly hack.

jberger:
Closing as superseded by a1f6d8a66d068b9c269efcf0bfb6e25bfd5a1d9f

PR #1522, opened 2020-05-23
Run GitHub Actions (on Linux) as non-root user

tianon:
### Summary

Runs the Linux-based GitHub Actions tests as a non-root user.

### Motivation

Most users won't be running Mojolicious itself as root (hopefully??), so IMO it makes sense to run the tests here that way too.

### References

This is something that's been discussed on IRC off and on since the GitHub Actions were first introduced, so I did some testing and worked out what I think is the cleanest way to get it working with minimal changes (I was hoping we could just do `options: --user 1000:1000` on the `container:` section, but that hilariously makes `actions/checkout` fail with permission issues on the working directory).

I also offered on IRC to maintain a set of Docker images for the Mojolicious project to use that are up-to-date and actively rebuilt periodically if that's something that would be of interest. :+1:

(Also worth noting, I'm happy to rebase, amend, adjust, etc as desired, and no hard feelings if this either isn't something the team wants or is considered too ugly to be worthwhile! :+1:)


tianon:
Another potential solution would be to not use GitHub's `container:` and instead invoke `docker run` and/or `docker build` directly in one of the `run:` steps, which I'd be happy to work on if that approach is preferred, but I think at some point if these older versions are important to the project it will require someone maintaining images (as I've noted on IRC, the only tags/versions in https://hub.docker.com/_/perl that are currently actively supported/rebuilt, esp for security updates, are 2.26+ at the time of this writing).

A different option would be running the tests against a matrix of specific distribution versions (Debian Stretch, Ubuntu Bionic, etc) instead, installing the distribution packages of Perl, etc, but IMO that's not testing quite the same thing.

tianon:
I managed to get my original `--user` approach working in https://github.com/mojolicious/mojo/commit/4b28ace3c829631b17bee214a96dc4f4cb970fe1 (see https://github.com/tianon/mojo/actions/runs/113004747), but it relies on an (as far as I can tell) undocumented GitHub user 1001 I found via https://github.com/actions/checkout/issues/47, so I'm not sure it's really suitable.

kraih:
I have no idea if we want this, lets collect come feedback.

tianon:
I think it's probably safe to say the core team doesn't think this is important -- is that a fair assessment?  Should we close this? :smile: :heart:

jberger:
@tianon I also don't really think it is important. I can see the utility, for example a test that accidentally tried to open a privileged port should fail. That seems like an odd case though. As it lies, I'm generally for it in concept but I think I'd rather wait until actions supported it directly rather than hacking their platform to make it so.

tianon:
Sounds good ❤️

Feel free to give me a poke (:fire: :sweat_smile:) if you find a reason for it later!
(or just steal my changes 👍)

PR #1525, opened 2020-05-26
Pre-upgrade "App::cpanminus" and "ExtUtils::MakeMaker"

### LINKED ISSUE #1508: Fix GitHub Actions to work with Perl 5.16 and 5.18
For [some reason](https://github.com/mojolicious/mojo/runs/650564585) our GitHub Actions do not install `Sub::Util` on Perl 5.16 and 5.18. So we had to [disable](https://github.com/mojolicious/mojo/commit/face9a1f2bde963cfb7ede31002c3d10c8988b1e) those versions for now. But they are rather important and we'd like to get them back. So this needs to be investigated!

tianon:
### Summary

On Perl 5.18 and older, `cpanm --installdeps` was not picking up `Sub::Util` and `IO::Socket::IP` properly, and updating `ExtUtils::MakeMaker` fixes it.  I included an update of `App::cpanminus` just for good measure.

### Motivation

Fixes #1508 :smile: :birthday:

### References

See #1508 for more details, especially https://github.com/mojolicious/mojo/issues/1508#issuecomment-634247300. :+1:

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -25,8 +25,8 @@ jobs:
         run: perl -V
       - name: Install Dependencies
         run: |
+          cpanm -n App::cpanminus ExtUtils::MakeMaker
           cpanm -n --installdeps .
-          cpanm -n Sub::Util IO::Socket::IP
           cpanm -n Cpanel::JSON::XS EV Role::Tiny
           cpanm -n Test::Pod Test::Pod::Coverage
       - name: Run Tests


karenetheridge:
What is missing in older EUMM that these prereqs are missing? Doesn't that suggest that the CONFIGURE_REQUIRES entry for EUMM in Makefile.PL should be increased?

tianon:
@karenetheridge a very good point, although definitely out of my depth -- hopefully this gives someone else enough to look into that :innocent: :heart:

tianon:
Hmm, even just `cpanm -n ExtUtils::MakeMaker@6.68` seems to be enough to fix it.

Looking through https://metacpan.org/diff/file?target=BINGOS/ExtUtils-MakeMaker-6.68/&source=BINGOS%2FExtUtils-MakeMaker-6.66#Changes, I think this might be relevant:

```diff
+6.67_04 Mon Jun 10 20:18:25 BST 2013
+    Bug fixes
+    * Address RT#85406, where specifying 'meta-spec' in
+      META_[ADD|MERGE] would remove all prereqs (bingos)
```

(The Mojolicious `Makefile.PL` includes `'meta-spec'    => {version => 2},` in `META_MERGE`.)

karenetheridge:
okay then, I would recommend adding `CONFIGURE_REQUIRES => '6.68'` to Makefile.PL, and then you should be able to remove all workarounds from docker and workflow configs.

thanks for tracking this down @tianon!

Issue #1520, opened 2020-05-20
Convert tests to using subtests


kraih:
Not a particularly hard task, but converting all tests is a lot of work. We want to go from:
```perl
# Promisify
is ref Mojo::Promise->resolve('foo'), 'Mojo::Promise', 'right class';
$promise = Mojo::Promise->reject('foo');
is ref $promise, 'Mojo::Promise', 'right class';
@errors = ();
$promise->catch(sub { push @errors, @_ })->wait;
is_deeply \@errors, ['foo'], 'promise rejected';
$promise = Mojo::Promise->resolve('foo');
is refaddr(Mojo::Promise->resolve($promise)), refaddr($promise), 'same object';
$promise = Mojo::Promise->resolve('foo');
isnt refaddr(Mojo::Promise->new->resolve($promise)), refaddr($promise),
  'different object';
$promise = Mojo::Promise->reject('foo');
is refaddr(Mojo::Promise->resolve($promise)), refaddr($promise), 'same object';
@errors = ();
$promise->catch(sub { push @errors, @_ })->wait;
is_deeply \@errors, ['foo'], 'promise rejected';
$promise  = Mojo::Promise->reject('foo');
$promise2 = Mojo::Promise->reject($promise);
isnt refaddr($promise2), refaddr($promise), 'different object';
@errors = ();
$promise2->catch(sub { push @errors, @_ })->wait;
is_deeply \@errors, ['foo'], 'promise rejected';
```
to:
```perl
subtest 'Promisify' => sub {
  is ref Mojo::Promise->resolve('foo'), 'Mojo::Promise', 'right class';

  $promise = Mojo::Promise->reject('foo');
  is ref $promise, 'Mojo::Promise', 'right class';
  @errors = ();
  $promise->catch(sub { push @errors, @_ })->wait;
  is_deeply \@errors, ['foo'], 'promise rejected';

  $promise = Mojo::Promise->resolve('foo');
  is refaddr(Mojo::Promise->resolve($promise)), refaddr($promise),
    'same object';

  $promise = Mojo::Promise->resolve('foo');
  isnt refaddr(Mojo::Promise->new->resolve($promise)), refaddr($promise),
    'different object';

  $promise = Mojo::Promise->reject('foo');
  is refaddr(Mojo::Promise->resolve($promise)), refaddr($promise),
    'same object';
  @errors = ();
  $promise->catch(sub { push @errors, @_ })->wait;
[...]

PR #1529, opened 2020-06-02
Convert t/mojo/date.t to use subtests

kiwiroy:
### Summary
Tests for `Mojo::Date` in `t/mojo/date.t` converted to subtests.

### Motivation
Another file converted.

### References
#1520.


kraih:
Thanks, applied.

PR #1530, opened 2020-06-17
Check app name in "mojo generate app"

Greenfactory:
### Summary
Some of the app names lead to the generation of broken code. So that this does not happen, it is worth checking the app name.

### Motivation
```
$ mojo generate app app-api
```
```
$ ./app-api/script/app-api
Can't find application class "app-api" in @INC.  (/home/gr...
```
```
$ perl -wc app-api/lib/app-api.pm 
Invalid version format (negative version number) at app-api/lib/app-api.pm line 1, near "package app"
syntax error at app-api/lib/app-api.pm line 1, near "package app-"
BEGIN not safe after errors--compilation aborted at app-api/lib/app-api.pm line 2.
```

In app-api.pm bad package name
```
$ head -1 app-api/lib/app-api.pm
package app-api;
```

### DIFF
--- a/lib/Mojolicious/Command/Author/generate/app.pm
+++ b/lib/Mojolicious/Command/Author/generate/app.pm
@@ -9,6 +9,10 @@ has usage       => sub { shift->extract_usage };
 sub run {
   my ($self, $class) = (shift, shift || 'MyApp');
 
+  unless($class =~ /^[a-zA-Z][a-zA-Z0-9_]*$/) {
+      die "Bad app name[$class]\n";
+  }
+
   # Script
   my $name = class_to_file $class;
   $self->render_to_rel_file('mojo', "$name/script/$name", {class => $class});
--- a/lib/Mojolicious/Command/Author/generate/app.pm
+++ b/lib/Mojolicious/Command/Author/generate/app.pm
@@ -9,9 +9,7 @@ has usage       => sub { shift->extract_usage };
 sub run {
   my ($self, $class) = (shift, shift || 'MyApp');
 
-  unless($class =~ /^[a-zA-Z][a-zA-Z0-9_]*$/) {
-      die "Bad app name[$class]\n";
-  }
+  $class =~ /^[a-zA-Z][a-zA-Z0-9_]*\z/ or die "Bad app name[$class]\n";
 
   # Script
   my $name = class_to_file $class;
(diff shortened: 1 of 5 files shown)


lib/Mojolicious/Command/Author/generate/app.pm
```diff
@@ -9,6 +9,10 @@ has usage       => sub { shift->extract_usage };
 sub run {
   my ($self, $class) = (shift, shift || 'MyApp');
 
+  unless($class =~ /^[a-zA-Z][a-zA-Z0-9_]*$/) {
```
Grinnz, line 12: This can be one line with postfix unless. Please also use `\z` instead of `$` so that trailing newlines are not accepted.

Grinnz: Double colons should also be allowed. Here is an example of a complete module name validation regex: https://metacpan.org/release/Module-Runtime/source/lib/Module/Runtime.pm#L176

Greenfactory: Done

kraih:
There's already a class validation regex [in the code](https://github.com/mojolicious/mojo/blob/2f0de50397a80e9b4f7f85880a616c230952c483/lib/Mojo/Loader.pm#L39).

Greenfactory:
> There's already a class validation regex [in the code](https://github.com/mojolicious/mojo/blob/2f0de50397a80e9b4f7f85880a616c230952c483/lib/Mojo/Loader.pm#L39).

Unfortunately, this check is not completely correct.
Package name can't begin from digit
```
perl -E'"1::2" =~ /^\w/ && say "OK"'
OK
```
```
$ perl -wcE'package a::2'
-e syntax OK
```

```
$ perl -wcE'package 1::2'
Bareword found where operator expected at -e line 1, near "1::2"
```
P.S. I think this option is more correct
```
^[a-zA-Z_]\w*(?:::\w+)*\z
```

kraih:
My point is not that it's correct, but that code sharing would be appropriate when the same regex is needed in two places. With lots and lots of tests.

Greenfactory:
Sorry for the long silence.
Was far from the internet.

kraih:
To be precise, i'd want to see `Mojo::Loader::is_package('Foo::Bar');`

kraih:
Looks like this PR might have been abandoned.

Greenfactory:
Cleared and fixed in https://github.com/mojolicious/mojo/pull/1559

PR #1531, opened 2020-06-22
Allow end and begin to occur on the same template line

jberger:
When `end` and `begin` occur on the same perl line or mix-line group, the presence of the `begin` token causes the `end` token to incorrectly be followed by a `;` which breaks expectations, is hard to work around, and is generally incorrect as it is still valid in theory.

### DIFF
--- a/t/mojo/template.t
+++ b/t/mojo/template.t
@@ -575,6 +575,19 @@ $output = $mt->render(<<'EOF');
 EOF
 is $output, " \n    2\n\n    3\n\n    4\n", 'block loop';
 
+subtest 'End and begin in the same perl line' => sub {
+  my $concat = 'no warnings "redefine"; sub concat { $_[0]->() . $_[1]->() }';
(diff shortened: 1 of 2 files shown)


t/mojo/template.t
```diff
@@ -575,6 +575,18 @@ $output = $mt->render(<<'EOF');
...
+my $concat = 'no warnings "redefine"; sub concat { $_[0]->() . $_[1]->() }';
+$mt = Mojo::Template->new(prepend => $concat);
+$output = $mt->render(<<'EOF');
+%= concat begin
+  1
+% end, begin
+  2
+% end
+EOF
+is $output, "\n  1\n  2\n", 'end, begin';
```
kraih, line 588: We use subtests for all new tests.

jberger:
I apologize if I didn't call the vote in the correct way, I don't think the change is very contentious and as it seems to have the necessary number of votes anyway, I'm going to merge. I'll do better next time when I'm not so stressed about writing a talk.

PR #1533, opened 2020-07-05
fix attr test for falsy values

tim-2:
### Summary
Make `attr_is` test work with falsy values.


kraih:
Actual `Test::Mojo` tests go [here](https://github.com/mojolicious/mojo/blob/master/t/test/mojo.t).

PR #1534, opened 2020-07-07
Allow render_later in before_dispatch

andrii-suse:
### Summary
Currently Controller->render_later has no effect if called from before_dispatch hook.
This blocks delayed rendering for tasks which don't have pre-defined routes, e.g. Mojolicious-Plugin-Directory

### Motivation
It looks that documentation doesn't forbid Controller->render_later in before_dispatch.
With proposed change delayed rendering from before_dispatch seems to work properly.

### DIFF
--- a/lib/Mojolicious.pm
+++ b/lib/Mojolicious.pm
@@ -123,7 +123,8 @@ sub dispatch {
 
   # Routes
   $plugins->emit_hook(before_routes => $c);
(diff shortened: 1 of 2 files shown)


jberger:
Running this through Perl::Tidy with the project's .perltidyrc I see that the line is too long. Please tidy.

marcusramberg:
image :+1:

PR #1538, opened 2020-07-11
Make proxy.start_p work on initial empty body

mohawk2:
### Summary
`proxy.start_p` misbehaves when it receives the headers only, with a zero-length body. It naively calls `$content->write(pop)` with that zero-length body, and "writ[ing] an empty chunk of data at any time [ends] the stream".

### Motivation
I believe this is not the intended behaviour. This PR has a test that captures this behaviour, and a patch to eliminate it.

### References
`#mojo` discussion over the last couple of weeks with @Grinnz and @CandyAngel (thanks!).

### DIFF
--- a/lib/Mojolicious/Plugin/DefaultHelpers.pm
+++ b/lib/Mojolicious/Plugin/DefaultHelpers.pm
@@ -189,7 +189,8 @@ sub _proxy_start_p {
       my $write = $source_content->is_chunked ? 'write_chunk' : 'write';
       $source_content->unsubscribe('read')->on(
         read => sub {
-          $content->$write(pop) and $tx->resume;
+          my $data = pop;
+          $content->$write(length $data ? $data : undef) and $tx->resume;
 
           # Throttle transparently when backpressure rises
           return if $stream->can_write;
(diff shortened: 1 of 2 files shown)


lib/Mojolicious/Plugin/DefaultHelpers.pm
```diff
@@ -189,7 +189,7 @@ sub _proxy_start_p {
       my $write = $source_content->is_chunked ? 'write_chunk' : 'write';
       $source_content->unsubscribe('read')->on(
         read => sub {
-          $content->$write(pop) and $tx->resume;
+          $content->$write(length $_[1] ? $_[1] : undef) and $tx->resume;
```
kraih, line 192: We only use `$_[x]` directly for one-liners and performance optimisations. Is this a performance optimisation?

mohawk2: No, I was attempting to imitate the "code-golfing" style. Would you prefer me to make it into `my $data = pop` above?

mohawk2: In anticipation that you do in fact want that, and hopefully save you an iteration, I've adjusted it and force-pushed. Please advise if in fact you'd like something different.

PR #1541, opened 2020-07-28
Convert more t/mojo/ tests to use subtests

kiwiroy:
### Summary

More mojo tests converted to use subtests.

### DIFF
--- a/t/mojo/bytestream.t
+++ b/t/mojo/bytestream.t
@@ -3,142 +3,177 @@ use Mojo::Base -strict;
 use Test::More;
 use Mojo::ByteStream qw(b);
 
-# Tap into method chain
-is b('test')->tap(sub { $$_ .= '1' })->camelize, 'Test1', 'right result';
+subtest 'Tap into method chain' => sub {
+  is b('test')->tap(sub { $$_ .= '1' })->camelize, 'Test1', 'right result';
+};
 
-# camelize
-is b('foo_bar_baz')->camelize, 'FooBarBaz', 'right camelized result';
+subtest 'camelize' => sub {
+  is b('foo_bar_baz')->camelize, 'FooBarBaz', 'right camelized result';
+};
 
-# decamelize
-is b('FooBarBaz')->decamelize, 'foo_bar_baz', 'right decamelized result';
+subtest 'decamelize' => sub {
+  is b('FooBarBaz')->decamelize, 'foo_bar_baz', 'right decamelized result';
+};
 
-# unindent
-is b(" test\n  123\n 456\n")->unindent, "test\n 123\n456\n", 'right unindented result';
+subtest 'unindent' => sub {
+  is b(" test\n  123\n 456\n")->unindent, "test\n 123\n456\n", 'right unindented result';
+};
 
-# b64_encode
-is b('foobar$%^&3217')->b64_encode, "Zm9vYmFyJCVeJjMyMTc=\n", 'right Base64 encoded result';
+subtest 'b64_encode' => sub {
+  is b('foobar$%^&3217')->b64_encode, "Zm9vYmFyJCVeJjMyMTc=\n", 'right Base64 encoded result';
+};
 
-# b64_decode
-is b("Zm9vYmFyJCVeJjMyMTc=\n")->b64_decode, 'foobar$%^&3217', 'right Base64 decoded result';
+subtest 'b64_decode' => sub {
+  is b("Zm9vYmFyJCVeJjMyMTc=\n")->b64_decode, 'foobar$%^&3217', 'right Base64 decoded result';
+};
 
-# url_escape
-is b('business;23')->url_escape, 'business%3B23', 'right URL escaped result';
+subtest 'url_escape' => sub {
+  is b('business;23')->url_escape, 'business%3B23', 'right URL escaped result';
+};
 
-# url_unescape
-is b('foo%C3%9F%C4%80bar%E2%98%BA')->url_unescape->decode, "foo\x{df}\x{0100}bar\x{263a}", 'right URL unescaped result';
-is b('foo%C3%9F%C4%80bar%E2%98%BA')->url_unescape->decode('UTF-8'), "foo\x{df}\x{0100}bar\x{263a}",
-  'right URL unescaped result';
+subtest 'url_unescape' => sub {
+  is b('foo%C3%9F%C4%80bar%E2%98%BA')->url_unescape->decode, "foo\x{df}\x{0100}bar\x{263a}",
+    'right URL unescaped result';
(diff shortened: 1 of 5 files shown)


t/mojo/bytestream.t
```diff
@@ -52,7 +52,8 @@ is b('"foo 23 \"bar"')->unquote, 'foo 23 "bar', 'right unquoted result';
 is b(' la la la ')->trim, 'la la la', 'right trimmed result';
 
 # md5_bytes
-is unpack('H*', b('foo bar baz ♥')->encode->md5_bytes), 'a740aeb6e066f158cbf19fd92e890d2d', 'right binary md5 checksum';
+is unpack('H*', b('foo bar baz ♥')->encode->md5_bytes), 'a740aeb6e066f158cbf19fd92e890d2d',
+  'right binary md5 checksum';
```
kraih, line 56: Why is this two lines now?

t/mojo/bytestream.t
```diff
@@ -77,49 +78,54 @@ ok !b('hell')->secure_compare('hello'), 'values are not equal';
...
+  is_deeply b('1,2,3,4,5,,,')->split(',')->to_array, [1, 2, 3, 4, 5], 'right elements';
+  is_deeply b('1,2,3,4,5,,,')->split(',', -1)->to_array, [1, 2, 3, 4, 5, '', '', ''], 'right elements';
+  is_deeply b('54321')->split('')->to_array, [5, 4, 3, 2, 1], 'right elements';
+  is_deeply b('')->split('')->to_array,    [], 'no elements';
+  is_deeply b('')->split(',')->to_array,   [], 'no elements';
+  is_deeply b('')->split(qr/,/)->to_array, [], 'no elements';
+  $stream = b('1/2/3');
+  is $stream->split('/')->map(sub { $_->quote })->join(', '), '"1", "2", "3"', 'right result';
+  is $stream->split('/')->map(sub { shift->quote })->join(', '), '"1", "2", "3"', 'right result';
+};
```
kraih, line 101: Why is this test only partially converted?

kiwiroy: All converted now.

t/mojo/bytestream.t
```diff
@@ -3,142 +3,177 @@ use Mojo::Base -strict;
...
-is $compressed->gunzip->to_string, $uncompressed->to_string, 'same bytestream';
-
-# humanize_bytes
-is b(8007188480)->humanize_bytes,  '7.5GiB',  'humanized';
-is b(-8007188480)->humanize_bytes, '-7.5GiB', 'humanized';
+subtest 'Tap into method chain' => sub {
+  is b('test')->tap(sub { $$_ .= '1' })->camelize, 'Test1', 'right result';
+};
+
+subtest camelize => sub {
```
kraih, line 10: Subtest descriptions are inconsistently quoted.

kiwiroy: All single quoted now.

kraih (approved):

t/mojo/collection.t
```diff
@@ -5,167 +5,186 @@ use Mojo::ByteStream qw(b);
...
-is_deeply $collection->tail(0)->to_array, [], 'right result';
-is_deeply $collection->tail(1)->to_array, [3], 'right result';
-is_deeply $collection->tail(2)->to_array, [4, 3], 'right result';
-is_deeply $collection->tail(-1)->to_array, [2, 5, 4, 3], 'right result';
-is_deeply $collection->tail(-3)->to_array, [4, 3], 'right result';
-is_deeply $collection->tail(5)->to_array, [1, 2, 5, 4, 3], 'right result';
-is_deeply $collection->tail(6)->to_array, [1, 2, 5, 4, 3], 'right result';
-is_deeply $collection->tail(-5)->to_array, [], 'right result';
-is_deeply $collection->tail(-6)->to_array, [], 'right result';
+subtest 'Array' => sub {
```
christopherraa, line 8: I see that there is a capital `A` here, whilst the other subtests (except the one right below this) use all-lowercase. Is this intentional or should this be converted to lowecase as well for consistency?

t/mojo/collection.t
```diff
@@ -5,167 +5,186 @@ use Mojo::ByteStream qw(b);
...
+  is c(1, 2, 3)->[1], 2, 'right result';
+  is_deeply [@{c(3, 2, 1)}], [3, 2, 1], 'right result';
+  my $collection = c(1, 2);
+  push @$collection, 3, 4, 5;
+  is_deeply [@$collection], [1, 2, 3, 4, 5], 'right result';
+};
+
+subtest 'Tap into method chain' => sub {
```
christopherraa, line 16: See previous comment about casing.

t/mojo/collection.t
```diff
@@ -5,167 +5,186 @@ use Mojo::ByteStream qw(b);
...
+  is_deeply $collection->uniq(sub { $_->[1] }), [[1, 2, 3], [3, 1, 2]], 'right result';
+  $collection = c(c(1, 2), c(1, 2), c(2, 1));
+  is_deeply $collection->uniq(join => ',')->flatten->to_array, [1, 2, 2, 1], 'right result';
+  $collection = c(undef, '', 3, 2, 1, 0);
+  is_deeply $collection->uniq->to_array, [undef, 3, 2, 1, 0], 'right result';
+  $collection = c(undef, '', 3, 2, 1, 0);
+  is_deeply $collection->uniq(sub {$_})->to_array, [undef, 3, 2, 1, 0], 'right result';
+};
+
+subtest 'TO_JSON' => sub {
```
christopherraa, line 160: Subtest called `TO_JSON` but the method called is `encode_json`. IMO this should be changed to match the style of the other tests.

christopherraa (changes requested):
Just a few minor nitpicks. When those are addressed I'll revisit this merge request.

Great job though!

t/mojo/collection.t
kraih: They use lowercase because they are method names. This is not a method name.

t/mojo/collection.t
kraih: `Mojo::Collection` contains a `TO_JSON` method that is actually tested here. All JSON encoders support that method.

t/mojo/collection.t
christopherraa: Ah, right. It was not clear to me that that was a convention that was followed.

t/mojo/collection.t
christopherraa: In my opinion it would be clearer exactly what is under test if that method was called directly, instead of the form that was chosen here. But I have no strong opinion that this needs changing, so I'll resolve this.

christopherraa (approved):
Seeing as my nitpicks were addressed / explained I'm adding my approval.

PR #1547, opened 2020-08-04
Add environment var to suppress "Unhandled rejected promise"

AdamWill:
This adds an environment variable, MOJO_NO_PROMISE_WARN, which
disables the emission of "Unhandled rejected promise" warnings.
This is to address an issue with openQA package builds in Fedora
(but may help in other similar situations): most openQA test
suites use Test::Warnings to fail on unexpected warnings, but
when run in Fedora's package build environment where networking
is intentionally nerfed, many dozens of these "Unhandled rejected
promise" warnings are generated by many pretty normal actions.
Test::Warnings does not have a convenient way to whitelist a
known error.

Signed-off-by: Adam Williamson <awilliam@redhat.com>

### DIFF
--- a/lib/Mojo/Promise.pm
+++ b/lib/Mojo/Promise.pm
@@ -39,7 +39,7 @@ sub AWAIT_ON_READY {
 sub DESTROY {
   my $self = shift;
   return                                                 if $self->{handled} || ($self->{status} // '') ne 'reject';
-  carp "Unhandled rejected promise: @{$self->{results}}" if $self->{results};
+  carp "Unhandled rejected promise: @{$self->{results}}" if $self->{results} &! $ENV{MOJO_NO_PROMISE_WARN};
 }
 
 sub all         { _all(2, @_) }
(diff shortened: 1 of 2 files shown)


kraih:
Things like this are why i'm opposed to the use of `Test::Warnings`. Warnings are just warnings and not exceptions for a reason.

shadowcat-mst:
Could you provide an actual example of something that goes wrong?

AdamWill:
@shadowcat-mst [here](https://kojipkgs.fedoraproject.org/work/tasks/4868/48634868/build.log) (it'll expire in a couple of weeks).

kraih:
I believe a feature for actually tracing and in turn helping with fixing unhandled rejected promises would be a better solution. I've made a branch with a possible proposal. https://github.com/mojolicious/mojo/compare/promise_debug

AdamWill:
building a newer openQA snapshot seems to avoid the problem happening, so I no longer need this. @kraih 's idea still seems like a good one, though.

Issue #1548, opened 2020-08-05
namespace method not returning expected value (Mojo::Dom)


sdondley:
* Mojolicious version: 8.57
* Perl version: 5.24.4
* Operating system: macOS 10.15.6

```
### Steps to reproduce the behavior
use Test::Most;
use Mojo::DOM;
use strict;
use warnings;

plan tests => 1;

my $html = '<html><head></head><body><booboo:p class="first">A paragraph.</p></body></html>';
my $ex = Mojo::DOM->new($html);
is $ex->at('booboo\:p')->namespace, 'booboo:p', 'fails'
```

### Expected behavior
test should pass

### Actual behavior
test fails with 

```
not ok 1 - fails

#   Failed test 'fails'
#   at t/02-namespace.t line 13.
#          got: undef
#     expected: 'booboo:p'
# Looks like you failed 1 test of 1.

```

MartinMcGrath:
Are you confusing a namespace with a tag?

```
# displays booboo:p
say $ex->at('booboo\:p')->tag;
```

See also [namespace](https://github.com/mojolicious/mojo/blob/master/lib/Mojo/DOM.pm#L80), [ namespace tests](https://github.com/mojolicious/mojo/blob/master/t/mojo/dom.t#L590).

sdondley:
I don't believe so, unless I'm badly misinterpreting/misreading the docs: https://docs.mojolicious.org/Mojo/DOM#namespace

kraih:
A namespace prefix is not the same as a namespace.

sdondley:
Might be helpful to show an example in the documentation of how to properly use it because I can't see to get the namespace method to return anything useful, even with this simple example:

```
my $html = '<html><head></head><body><svg:circle></svg:circle></svg></body></html>';
my $dom = Mojo::DOM->new($html);
my $namespace = $dom->at('svg\:circle')->namespace;
# $namespace is undef
```

sdondley:
OK, got this example working:

my $html = '<svg xmlns:svg="www.example.com/svg"><svg:circle>3.14</svg:circle></svg>';
my $dom = Mojo::DOM->new($html);
my $namespace = $dom->at('svg > svg\:circle')->namespace;

# returns  'www.example.com/svg'

Just let me know if you want me to submit a patch for the docs with this example, please. Thanks.

jberger:
[...]

PR #1549, opened 2020-08-06
provide example of a result for the namespace method

sdondley:
### Summary
Make documentation more friendly for developers not familiar with namespaces (like me)

### Motivation
I was confused on what method should return. A simple example should help eliminate confusion.

### References
Issue #1548

### DIFF
--- a/lib/Mojo/DOM.pm
+++ b/lib/Mojo/DOM.pm
@@ -686,6 +686,9 @@ Trailing key/value pairs can be used to declare xml namespace aliases.
 
 Find this element's namespace, or return C<undef> if none could be found.
 
+  # "http://www.w3.org/2000/svg"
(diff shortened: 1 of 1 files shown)


lib/Mojo/DOM.pm
```diff
@@ -686,6 +686,9 @@ Trailing key/value pairs can be used to declare xml namespace aliases.
 
 Find this element's namespace, or return C<undef> if none could be found.
 
+  # 'http://www.w3.org/2000/svg'
```
kraih, line 689: I believe other examples use double quotes.

jberger: Does that example even work? I had to do 

```
Mojo::DOM->new('<svg xmlns:svg="http://www.w3.org/2000/svg"><svg:circle>3.14</svg:circle></svg>')->at('circle')->namespace;
```

to get the result you see.

PR #1551, opened 2020-08-10
Add FAQ entry for prototype error when not using signatures

Grinnz:
### Summary
Add a faq entry for the error received when trying to follow examples with signatures without enabling the signatures feature.

### Motivation
Signatures from examples can get interpreted as prototypes and it's not immediately clear the signatures feature is required, especially for users not familiar with signatures or prototypes.

### DIFF
--- a/lib/Mojolicious/Guides/FAQ.pod
+++ b/lib/Mojolicious/Guides/FAQ.pod
@@ -212,6 +212,29 @@ weakened.
     $c->reply->exception($err);
   });
 
+=head2 What does "Illegal character in prototype" mean?
+
+Mojolicious assumes L<subroutine signatures|Mojolicious::Guides/"Signatures"> are enabled in documentation examples. If
+the signatures feature has not been enabled in that scope, they are interpreted as L<prototypes|perlsub/"Prototypes">,
+an unrelated parser feature. Mojolicious does not require signatures; if you don't want to or cannot use signatures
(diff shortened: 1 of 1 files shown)


jhthorsen (dismissed):
Looks good to me.

lib/Mojolicious/Guides/FAQ.pod
```diff
@@ -207,6 +207,30 @@ weakened.
     $c->reply->exception($err);
   });
 
+=head2 What does "Illegal character in prototype" mean?
+
+Mojolicious assumes L<subroutine signatures|Mojolicious::Guides/"Signatures"> are enabled in documentation examples. If
+the signatures feature has not been enabled in that scope, they are interpreted as L<prototypes|perlsub/"Prototypes">,
+an unrelated parser feature. The signatures feature is not required to use any Mojo tool; if you don't want to or
```
kraih, line 214: `Mojo tool` sounds weird.

kraih: Should probably say `Mojo feature`.

jberger: I'd just rephrase. `Mojolicious does not require signatures` possibly with a leading clause like `However, `

Grinnz:
Rephrased as suggested

kraih (approved):

kraih (approved):
That works.

jhthorsen (approved):
LGTM 👍

PR #1552, opened 2020-08-17
untaint command name before use

guillomovitch:
### Summary
Fix command usage in taint mode

### References
Issue #1553

### DIFF
--- a/lib/Mojolicious/Commands.pm
+++ b/lib/Mojolicious/Commands.pm
@@ -34,7 +34,7 @@ sub run {
   if (!$ENV{MOJO_NO_DETECT} && (my $env = $self->detect)) { $name = $env }
 
   # Run command
-  if ($name && $name =~ /^\w[\w-]+$/ && ($name ne 'help' || $args[0])) {
+  if ($name && (($name) = $name =~ /^(\w[\w-]+)$/) && ($name ne 'help' || $args[0])) {
     $name =~ s/-/_/g;
 
     # Help


guillomovitch:
I have no idea about the difficulty of establishing support for taint mode in Mojolicious. My impression was than most parameters were already passed through filtering regexps, and it missed just a few capture statements in order to make the interpreter aware. But I'm not sure either of the exact benefit of taint mode nowadays...

What I'm sure, however, is than an explicit statement about taint mode in documentation would benefit users: I spend a couple of hours trying to figure if the problem was coming from my own code, or in Mojolicious one.

genio:
https://www.nntp.perl.org/group/perl.perl5.porters/2019/09/msg256202.html

Taint mode is the single worst thing for security ever invented in Perl. It does not prevent any actual abuses, it breaks entire OSes, and only provides warm fuzzies until you realize that people only untaint with a `/(.*)/` because there's absolutely NO WAY to validate the responses from the majority of commands. It's pointless and should be removed all together from Perl.

Adding support for such in Mojo should not even be considered.

jberger:
> My impression was than most parameters were already passed through filtering regexps, and it missed just a few capture statements in order to make the interpreter aware. But I'm not sure either of the exact benefit of taint mode nowadays...

Consider the fact that if as you say, it almost already work accidentally, though it wasn't attempt to be so, how secure can its methodology be? That's not to say Mojolicious doesn't try to be secure, of course we do, but not via the definition that Taint mode defines.

kraih:
There's also an FAQ answer now. https://docs.mojolicious.org/Mojolicious/Guides/FAQ#Is-Perls-taint-mode-supported-by-Mojolicious

guillomovitch:
The FAQ entry is fine, I'm closing this PR.

PR #1559, opened 2020-08-30
Add Mojo::Loader::is_package

Greenfactory:
### Summary
Some of the app names lead to the generation of broken code. So that this does not happen, it is worth checking the app name.

### Motivation
```
$ mojo generate app app-api
```

```
$ ./app-api/script/app-api
Can't find application class "app-api" in @INC.  (/home/gr...
```

```
$ perl -wc app-api/lib/app-api.pm 
Invalid version format (negative version number) at app-api/lib/app-api.pm line 1, near "package app"
syntax error at app-api/lib/app-api.pm line 1, near "package app-"
BEGIN not safe after errors--compilation aborted at app-api/lib/app-api.pm line 2.
```

In app-api.pm bad package name

```
$ head -1 app-api/lib/app-api.pm
package app-api;
```

### DIFF
--- a/lib/Mojo/Loader.pm
+++ b/lib/Mojo/Loader.pm
@@ -6,7 +6,7 @@ use Mojo::Exception;
 use Mojo::File qw(path);
 use Mojo::Util qw(b64_decode class_to_path);
 
-our @EXPORT_OK = qw(data_section file_is_binary find_modules find_packages load_class);
+our @EXPORT_OK = qw(data_section file_is_binary find_modules find_packages load_class is_package);
 
 my (%BIN, %CACHE);
 
@@ -32,11 +32,31 @@ sub find_packages {
   return sort map { /^(.+)::$/ ? "${ns}::$1" : () } keys %{"${ns}::"};
 }
 
+sub is_package {
+  my $package = shift;
+  if($package && $package =~
+    /
+      ^
+      [a-zA-Z_]         # Can't begin from digit
+      [a-zA-Z0-9_]*
+      (?:
+          (?: :: | ' )  # Package::name eq Package'name
+          [a-zA-Z0-9_]+
+      )*
+      \z
+    /x) {
+      return 1;
+  }
+  return 0;
+}
+
+
+
 sub load_class {
   my $class = shift;
 
   # Invalid class name
-  return 1 if ($class || '') !~ /^\w(?:[\w:']*\w)?$/;
+  return 1 unless is_package($class);
 
   # Load if not already loaded
   return undef if $class->can('new') || eval "require $class; 1";
@@ -149,6 +169,14 @@ Check if embedded file from the C<DATA> section of a class was Base64 encoded.
   my @pkgs = find_packages 'MyApp::Namespace';
 
 Search for packages in a namespace non-recursively.
+
+=head2 is_package
+
+  if( is_package('a1::aaa') ){
+    say "package name valid";
+  }
+
+Check package name.
 
 =head2 find_modules
 
--- a/lib/Mojo/Loader.pm
+++ b/lib/Mojo/Loader.pm
@@ -6,7 +6,7 @@ use Mojo::Exception;
 use Mojo::File qw(path);
 use Mojo::Util qw(b64_decode class_to_path);
 
-our @EXPORT_OK = qw(data_section file_is_binary find_modules find_packages load_class is_package);
+our @EXPORT_OK = qw(data_section file_is_binary find_modules find_packages is_package load_class);
 
 my (%BIN, %CACHE);
 
--- a/lib/Mojo/Loader.pm
+++ b/lib/Mojo/Loader.pm
@@ -172,11 +172,17 @@ Search for packages in a namespace non-recursively.
 
 =head2 is_package
 
-  if( is_package('a1::aaa') ){
-    say "package name valid";
-  }
+  my $bool = is_package 'Mojo::Base';
 
-Check package name.
+Check if value is a package name.
+
+  # False
+  is_package 'MyApp.pm';
+  is_package 'My/App.pm';
+
+  # True
+  is_package 'MyApp';
+  is_package 'My::App';
 
 =head2 find_modules
 
--- a/lib/Mojolicious/Command/Author/generate/app.pm
+++ b/lib/Mojolicious/Command/Author/generate/app.pm
@@ -2,12 +2,15 @@ package Mojolicious::Command::Author::generate::app;
 use Mojo::Base 'Mojolicious::Command';
 
 use Mojo::Util qw(class_to_file class_to_path decamelize);
+use Mojo::Loader qw(is_package);
 
 has description => 'Generate Mojolicious application directory structure';
 has usage       => sub { shift->extract_usage };
 
 sub run {
   my ($self, $class) = (shift, shift || 'MyApp');
+
+  is_package($class) or die "Bad app name[$class]\n";
 
   # Script
   my $name = class_to_file $class;
--- a/lib/Mojolicious/Command/Author/generate/app.pm
+++ b/lib/Mojolicious/Command/Author/generate/app.pm
@@ -10,7 +10,7 @@ has usage       => sub { shift->extract_usage };
 sub run {
   my ($self, $class) = (shift, shift || 'MyApp');
 
-  is_package($class) or die "Bad app name[$class]\n";
+  is_package($class) or die "Can't generate application: Bad name[$class]\n";
 
   # Script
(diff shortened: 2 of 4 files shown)


lib/Mojo/Loader.pm
```diff
@@ -32,11 +32,31 @@ sub find_packages {
...
+      (?:
+          (?: :: | ' )  # Package::name eq Package'name
+          [a-zA-Z0-9_]+
+      )*
+      \z
+    /x) {
+      return 1;
+  }
+  return 0;
+}
```
kraih, line 51: That probably fits on one line.

lib/Mojo/Loader.pm
```diff
@@ -150,6 +170,14 @@ Check if embedded file from the C<DATA> section of a class was Base64 encoded.
 
 Search for packages in a namespace non-recursively.
 
+=head2 is_package
+
+  if( is_package('a1::aaa') ){
+    say "package name valid";
+  }
+
+Check package name.
```
kraih, line 179: Inconsistent documentation.

lib/Mojolicious/Command/Author/generate/app.pm
```diff
@@ -2,13 +2,16 @@ package Mojolicious::Command::Author::generate::app;
...
 use Mojo::Util qw(class_to_file class_to_path decamelize);
+use Mojo::Loader qw(is_package);
 
 has description => 'Generate Mojolicious application directory structure';
 has usage       => sub { shift->extract_usage };
 
 sub run {
   my ($self, $class) = (shift, shift || 'MyApp');
 
+  is_package($class) or die "Bad app name[$class]\n";
```
kraih, line 13: Untested.

t/mojo/loader.t
```diff
@@ -72,6 +72,31 @@ subtest 'Search packages' => sub {
   is_deeply [find_packages 'MyLoaderTest::DoesNotExist'], [], 'no packages found';
 };
 
+# is_package
+# To check cases use
+# perl -e"package aaa::a; print __PACKAGE__"
```
kraih, line 77: You forgot to remove this.

t/mojo/loader.t
```diff
@@ -7,8 +7,8 @@ use Test::More;
 use Mojo::File qw(curfile);
 use lib curfile->sibling('lib')->to_string;
 
-use Mojo::Loader qw(data_section file_is_binary find_packages find_modules load_class);
-
+use Mojo::Loader qw(data_section file_is_binary find_packages find_modules load_class is_package);
```
kraih, line 10: Not alphabetically ordered.

lib/Mojo/Loader.pm
```diff
@@ -6,7 +6,7 @@ use Mojo::Exception;
 use Mojo::File qw(path);
 use Mojo::Util qw(b64_decode class_to_path);
 
-our @EXPORT_OK = qw(data_section file_is_binary find_modules find_packages load_class);
+our @EXPORT_OK = qw(data_section file_is_binary find_modules find_packages load_class is_package);
```
kraih, line 9: Not alphabetically ordered.

kraih (changes requested):

lib/Mojo/Loader.pm
Greenfactory: You meant
$package && $package =~ /.../ ? 1 : 0;
or collapse of regexp.

lib/Mojo/Loader.pm
Greenfactory: Sorry, my english isn't good for write pod. Could you please write text here? I'll insert it in request.

lib/Mojolicious/Command/Author/generate/app.pm
kraih: Also pretty bad exception text, make it consistent with other cases in the Mojolicious repo.

lib/Mojo/Loader.pm
kraih: All.

lib/Mojo/Loader.pm
kraih: Not just the text, also the code.

kraih: Simplistic text i don't mind too much, as long as it's phrased correctly. But the rest needs to be consistent with the rest of the docs.

kraih:
Where did you take the regex from?

lib/Mojo/Loader.pm
kiwiroy: This would be more consistent.

```pod
=head2 is_package

  my $bool = is_package "Mojo::Base";
  # "Incorrect name for package"
  say "Incorrect name for package" unless is_package "MyApp.pm";

Check if value is a package name.
```

kraih: No it wouldn't at all @kiwiroy. We never have comments in the calling convention section.

christopherraa: I don't think much tweaking is needed here. How about something like this?
```
=head2 is_package

  say "Package name is valid" if is_package "Mojo::Base";

Check if a provided string is a valid package name. Please note that this only verifies the name of the package. It does not check if the package exists or is loaded.
```

I changed the example to a test returning something truthy instead of something falsey. Also the description text I provided I am still unsure of, specifically the latter sentence. Any input from you @kraih is appreciated.
```

File: lib/Mojolicious/Command/Author/generate/app.pm (hunk 3 above)
Comment (christopherraa) [reply lib/Mojolicious/Command/Author/generate/app.pm:13]: Agreed. A change here would be good.

**Review by christopherraa — CHANGES_REQUESTED** [review]
Mostly a fine change. The issues brought up should be addressed before any approval can be added, at least from my standpoint. I'll check back when the comments have been addressed.

File: lib/Mojo/Loader.pm (hunk 2 above)
Comment (kraih) [reply lib/Mojo/Loader.pm:179]: The convention here is:
```perl
=head2 is_package

  my $bool = is_package 'Mojo::Base';

Check if value is a package name.

  # False
  is_package 'MyApp.pm';
  is_package 'My/App.pm';

  # True
  is_package 'MyApp';
  is_package 'My::App';
```

christopherraa: Yeah, that is _much_ better :+1: 

@greenfactory Could you update the code accordingly?

t/mojo/loader.t
Greenfactory: I wanted to leave it for those who need to add tests in the future

Greenfactory:
> Where did you take the regex from?
Nowhere. I wrote it

t/mojo/loader.t
kraih: We don't leave comments like that around.

lib/Mojo/Loader.pm
```diff
@@ -32,11 +32,31 @@ sub find_packages {
...
+      )*
+      \z
+    /x) {
+      return 1;
+  }
+  return 0;
+}
+
+
+
```
kraih, line 54: Extra newlines.

lib/Mojolicious/Command/Author/generate/app.pm
```diff
@@ -2,13 +2,16 @@ package Mojolicious::Command::Author::generate::app;
...
 use Mojo::Util qw(class_to_file class_to_path decamelize);
+use Mojo::Loader qw(is_package);
 
 has description => 'Generate Mojolicious application directory structure';
 has usage       => sub { shift->extract_usage };
 
 sub run {
   my ($self, $class) = (shift, shift || 'MyApp');
 
+  is_package($class) or die "Can't generate application: Bad name[$class]\n";
```
kraih, line 13: You seem to have dismissed requests to change this error message without addressing them.

lib/Mojo/Loader.pm
kraih: If you don't want to emulate the Mojolicious coding style then this code does not belong in Mojolicious. That regex is not rocket science. We only expand non-trivial regexes with comments.

kraih:
I don't see why this implementation uses a custom regex instead of reusing one that has proven itself already in an established module like [Module::Runtime](https://metacpan.org/release/Module-Runtime/source/lib/Module/Runtime.pm#L176).

lib/Mojo/Loader.pm
kraih: But aside from that, i'm more concerned with the fact that this is a custom regex. It should really have been an existing one copied from a proven module.

kraih (changes requested):
I'm blocking this PR because of the custom regex.

Issue #1517, opened 2020-05-20
$promise1->resolve($promise1) should reject an exception


akarelas:
* Mojolicious version: current `unhandled_promises` branch (`3d91e75`)
* Perl version: 5.30.2
* Operating system: Linux 5.3.0-51

### Steps to reproduce the behavior

```perl
my $p1 = Mojo::Promise->new;
$p1->then(sub {say "p1 resolved: @_"}, sub {say "p1 rejected: @_"});

$p1->resolve($p1);

Mojo::IOLoop->one_tick;
```

The above does not output an exception (or something resembling it), while it should.

### Expected behavior
According to https://tc39.es/ecma262/#sec-promise-resolve-functions (point 7), `$p1` should reject with a result of `selfResolutionError` which is a `TypeError` object.

The script should output something, perhaps something like this:

`p1 rejected: ... at line ...`

### Actual behavior

The above script didn't produce any output.

kraih:
Patch did not pass the vote.

Issue #1515, opened 2020-05-20
Mojo::Promise objects sometimes produce side-effects the 2nd time resolve is called on them


akarelas:
* Mojolicious version: current `unhandled_promises` branch (`3d91e75`)
* Perl version: 5.30.2
* Operating system: Linux 5.3.0-51

### Steps to reproduce the behavior
```perl
say '--- output section 1';
my $p1 = Mojo::Promise->new;
my $p2 = Mojo::Promise->new;
$p2->then(sub {say 'hi 2'});
$p2->resolve($p1);

Mojo::IOLoop->one_tick;

say '--- output section 2';
$p2->resolve(10);

Mojo::IOLoop->one_tick;
```

The above outputs:

```
--- output section 1
--- output section 2
hi 2
```

### Expected behavior

While I'm not sure about section 1 of the output, section 2 should be empty, since the second call to resolve should be a no-op, according to https://tc39.es/ecma262/#sec-promise-resolve-functions bullet points 5 and 6

### Actual behavior

Section 2 of the output was not empty

abbottmg:
I think the example script leaves `$p2` unresolved until the `$p2->resolve(10)` call in the second section; if I'm reading `_settle` correctly, the first call is the equivalent of `$p1->then(sub { $p2->resolve(@__)  }, sub { $p2->reject(@_) }`. This is basically steps 12-14 of your ECMA reference.

The end result is that we're telling `$p2` to depend on `$p1` for resolution but never resolving `$p1`. The second call passes a non-promise value, `$p2` sees it's not resolved (and doesn't know it's already agreed to depend on `$p1`) and so resolves.

Adding `$p1->resolve()` a tick prior to our `resolve(10)` shows behavior we'd expect:

```perl
# example 1
say '--- output section 1';
my $p1 = Mojo::Promise->new;
my $p2 = Mojo::Promise->new;
$p2->then(sub {say 'hi 2'});
$p2->resolve($p1);
$p1->resolve(); # resolve p1 in same time slot as p2
Mojo::IOLoop->one_tick;

say '--- output section 2';
$p2->resolve(10);

Mojo::IOLoop->one_tick;
```

```perl
# example 2
say '--- output section 1';
my $p1 = Mojo::Promise->new;
my $p2 = Mojo::Promise->new;
$p2->then(sub {say 'hi 2'});
$p2->resolve($p1);
Mojo::IOLoop->one_tick; # tick 1

[...]

PR #1561, opened 2020-09-05
$p->resolve($p) rejects with an error message, and only one call to resolve considered

akarelas:
### Summary
Sort of solves bug #1517 (`$p->resolve($p)` should reject an exception). I say "sort of" because the reason of the rejection is not a TypeError object (itself a subclass of Error), but a simple text message, because I don't know how to produce errors (and if they are at all relevant to Perl). Nevertheless, with this PR I provide a solution the maintainers can modify to make an Error if they so wish.

**Edit:** This PR now also solves bug #1515.

### Motivation
This change makes Mojo::Promise objects more in accordance with the spec (https://tc39.es/ecma262/#sec-promise-resolve-functions)

### References
* Bug #1517 
* https://tc39.es/ecma262/#sec-promise-resolve-functions point 7
* https://tc39.es/ecma262/#sec-promise-resolve-functions points 5 & 6

### DIFF
--- a/lib/Mojo/Promise.pm
+++ b/lib/Mojo/Promise.pm
@@ -4,7 +4,7 @@ use Mojo::Base -base;
 use Carp qw(carp);
 use Mojo::Exception;
 use Mojo::IOLoop;
-use Scalar::Util qw(blessed);
+use Scalar::Util qw(blessed refaddr);
 
 use constant DEBUG => $ENV{MOJO_PROMISE_DEBUG} || 0;
 
@@ -214,8 +214,19 @@ sub _settle {
     $self = $self->new;
   }
 
+  return $self if $self->{semi_settled}++;
+
   if ($thenable && $status eq 'resolve') {
-    $results[0]->then(sub { $self->resolve(@_); () }, sub { $self->reject(@_); () });
+    if (refaddr $results[0] == refaddr $self) {
+      undef $self->{semi_settled};
+      $self->reject("Chaining cycle detected for promise $self");
+    }
+    else {
+      $results[0]->then(
+        sub { undef $self->{semi_settled}; $self->resolve(@_); () },
+        sub { undef $self->{semi_settled}; $self->reject(@_); () }
+      );
+    }
   }
 
   elsif (!$self->{results}) {
--- a/t/mojo/promise.t
+++ b/t/mojo/promise.t
@@ -349,7 +349,7 @@ subtest 'All settled (with rejection)' => sub {
   is_deeply \@results, $result, 'promise resolved';
 };
 
-subtest 'Settle with promise' => sub {
+subtest 'Settle with settled promise' => sub {
   my $promise = Mojo::Promise->new->resolve('works');
   my @results;
   my $promise2 = Mojo::Promise->new->resolve($promise)->then(sub { push @results, 'first', @_; @_ });
@@ -366,6 +366,47 @@ subtest 'Settle with promise' => sub {
   is_deeply \@errors, ['first', $promise], 'promises rejected';
   is_deeply \@results, ['second'], 'promises resolved';
   $promise->catch(sub { });
+};
+
+subtest 'Settle with self' => sub {
+  my $promise = Mojo::Promise->new;
+  my (@results, @errors);
+  $promise->resolve($promise)->then(sub { @results = @_ }, sub { @errors = @_ });
+  Mojo::IOLoop->one_tick;
+  is_deeply \@results, [], 'promise not resolved';
+  cmp_ok scalar(@errors), '==', 1, 'correct number of errors';
+  like $errors[0], qr/^Chaining cycle detected for promise \S/, 'correct error';
+
+  $promise = Mojo::Promise->new;
+  (@results, @errors) = ();
+  $promise->reject($promise)->then(sub { @results = @_ }, sub { @errors = @_; () });
+  Mojo::IOLoop->one_tick;
+  is_deeply \@results, [], 'promise not resolved';
+  is_deeply \@errors, [$promise], 'correct errors';
+};
+
+subtest 'Settle with unsettled promise' => sub {
+  my $promise  = Mojo::Promise->new;
+  my $promise2 = Mojo::Promise->new;
+  my @results;
+  $promise->then(sub { @results = @_ });
+  $promise->resolve($promise2);
+  Mojo::IOLoop->one_tick;
+  is_deeply \@results, [], 'promise not resolved';
+  $promise->resolve(123);
+  Mojo::IOLoop->one_tick;
+  is_deeply \@results, [], 'promise not resolved';
+  $promise2->resolve(1_000);
+  Mojo::IOLoop->one_tick;
+  is_deeply \@results, [1_000], 'promise resolved';
+
+  $promise = Mojo::Promise->new;
+  (@results, my @errors) = ();
+  $promise->then(sub { @results = @_ }, sub { @errors = @_ });
+  $promise->reject($promise2);
+  Mojo::IOLoop->one_tick;
+  is_deeply \@results, [], 'promise not resolved';
+  is_deeply \@errors, [$promise2], 'correct errors';
 };
 
 subtest 'Promisify' => sub {
@@ -490,7 +531,7 @@ subtest 'Map (with reject)' => sub {
     1 .. 5
   )->then(sub { @results = @_ }, sub { @errors = @_ })->wait;
   is_deeply \@results, [], 'promise not resolved';
-  is_deeply \@errors, [1], 'correct errors';
+  is_deeply \@errors,  [1], 'correct errors';
   is_deeply \@started, [1, 2, 3], 'only initial batch started';
 };
 


lib/Mojo/Promise.pm
```diff
@@ -215,7 +215,8 @@ sub _settle {
   }
 
   if ($thenable && $status eq 'resolve') {
-    $results[0]->then(sub { $self->resolve(@_); () }, sub { $self->reject(@_); () });
+    if (refaddr $results[0] == refaddr $self) {$self->reject('Chaining cycle detected')}
+    else { $results[0]->then(sub { $self->resolve(@_); () }, sub { $self->reject(@_); () }); }
```
jhthorsen, line 219: I don't think this is run through perltidy. I suggest this formatting:

```
    (refaddr $results[0] == refaddr $self)
      ? $self->reject('Chaining cycle detected.')
      : $results[0]->then(sub { $self->resolve(@_); () }, sub { $self->reject(@_); () });
```

jhthorsen (changes requested):
This PR cannot be accepted without any accommodating tests.

I'm unsure if this is a real issue though or simply an academic one, but I'm positive to the change.

lib/Mojo/Promise.pm
akarelas: I don't have access to perltidy right now. You'd have to run it through perltidy yourself, if that's not a huge problem for you. Thanks.

christopherraa:
I'm also positive to the change, but there needs to be tests. Please add tests for this change and I'll follow up by reviewing it.

akarelas:
This PR now solves two issues (#1515 and #1517), contains tests, and have passed everything (`lib/Mojo/Promise.pm` and `t/mojo/promise.t`) through perltidy.

lib/Mojo/Promise.pm
```diff
@@ -214,8 +214,19 @@ sub _settle {
     $self = $self->new;
   }
 
+  !$self->{_semi_settled}++ or return $self;
```
kraih, line 217: We don'y use underscore names.

kraih: Or this `... or return ...` style.

t/mojo/promise.t
```diff
@@ -516,4 +516,29 @@ subtest 'Wait for stopped loop' => sub {
   is_deeply \@results, ['wait'], 'promise resolved';
 };
 
+subtest 'Resolve self' => sub {
+  my $promise = Mojo::Promise->new;
+  my @results = ([], []);
```
kraih, line 521: Other tests don't use `@results` for both.

t/mojo/promise.t
```diff
@@ -516,4 +516,29 @@ subtest 'Wait for stopped loop' => sub {
...
+  $promise->resolve($promise)->then(sub { $results[0] = \@_ }, sub { $results[1] = \@_ });
+  Mojo::IOLoop->one_tick;
+  is_deeply $results[0], [], 'no resolve results';
```
kraih, line 524: Inconsistent description.

kraih (dismissed):

kraih:
Your commit messages are really bad, to the point where i will have to reject the PR for it. And you did not squash the commits.

lib/Mojo/Promise.pm
```diff
@@ -214,8 +214,19 @@ sub _settle {
...
+
   if ($thenable && $status eq 'resolve') {
-    $results[0]->then(sub { $self->resolve(@_); () }, sub { $self->reject(@_); () });
+    if (refaddr $results[0] == refaddr $self) {
+      undef $self->{_semi_settled};
+      $self->reject('Chaining cycle detected');
```
kraih, line 222: Isn't the JavaScript exception `Chaining cycle detected for promise`?

lib/Mojo/Promise.pm
akarelas: addressed

lib/Mojo/Promise.pm
akarelas: Addressed.

t/mojo/promise.t
akarelas: Addressed.

t/mojo/promise.t
akarelas: Addressed.

lib/Mojo/Promise.pm
akarelas: Addressed.

**node.js** exception is `Chaining cycle detected for promise [Promise]`. Remembering how **node.js** outputs `[object Object]` when you stringify an object (as in `console.log('hi: ' + {})`), it came to my mind that it would be more correct, and helpful to developers who are investigating their own bugs at the same time, if I stringified $self (a promise) in this message too (as in: `"Chaining cycle detected for promise $self"`).

If you don't agree with this I will remove $self from that string.

lib/Mojo/Promise.pm
```diff
@@ -214,8 +214,19 @@ sub _settle {
     $self = $self->new;
   }
 
+  return $self if $self->{semi_settled}++;
+
   if ($thenable && $status eq 'resolve') {
-    $results[0]->then(sub { $self->resolve(@_); () }, sub { $self->reject(@_); () });
+    if (refaddr $results[0] == refaddr $self) {
+      undef $self->{semi_settled};
+      $self->reject("Chaining cycle detected for promise $self");
```
kraih, line 222: What is the use of including `$self` there?

akarelas:
> Your commit messages are really bad, to the point where i will have to reject the PR for it. And you did not squash the commits.

Fixed messages and squashed commits, but now the message takes 2 lines.

t/mojo/promise.t
```diff
@@ -516,4 +516,30 @@ subtest 'Wait for stopped loop' => sub {
   is_deeply \@results, ['wait'], 'promise resolved';
 };
 
+subtest 'Resolve to self' => sub {
+  my $promise = Mojo::Promise->new;
+  my (@results, @errors);
+  $promise->resolve($promise)->then(sub { @results = @_ }, sub { @errors = @_ });
+  Mojo::IOLoop->one_tick;
+  is_deeply \@results, [], 'promise not resolved';
+  cmp_ok scalar(@errors), '==', 1, 'correct number of errors';
```
kraih, line 525: How is that better than `is_deeply \@errors, [], ...`?

lib/Mojo/Promise.pm
akarelas: The use is that this kind of cycling error is only going to happen in a bug. So I included `$self` here to allow the developer to debug and figure out which of their promises took part in the cycle. (I'm assuming the developer can insert debugging code in their own code to display the memory address of their promises, but shouldn't have to tweak the Mojolicious library's code to display the address of the promise that took part in the cycle).

akarelas: Plus, node.js also displays `...for promise [Promise]` in their error message, which kind of reminds how objects are stringified (`[object Object]`), so perhaps including `$self` increases the consistency with node.js.

t/mojo/promise.t
```diff
@@ -516,4 +516,30 @@ subtest 'Wait for stopped loop' => sub {
...
+  like $errors[0], qr/^Chaining cycle detected for promise \S/, 'correct error';
+};
+
+subtest 'Resolve to unsettled promise' => sub {
```
kraih, line 529: Is it resolve "to" or resolve "with" unsettled promise?

akarelas: I do not know. Couldn't find either of the two in past tests. I believe I haven't used both verbs in my commit.

But I believe "resolve to" conveys the message in a more clear way in this case, because "Resolve with unsettled promise" could also mean "calling `resolve` on an unsettled promise" which is not what I meant.

* This promise resolves to 5
* This promise resolves with 5

Maybe both expressions are valid, but I can't say. I'm not that good with English prepositions.

akarelas:
I created this looping test script, which tests 10 settling cases for Promises. e.g. `Mojo::Promise->resolve($p);` and `$p->reject($p)`, etc.

Maybe you might want to keep a copy to test all 10 cases next time you change something in Mojo::Promise. I created it to test for bugs, and to check whether I fixed them in my PR.

https://gist.github.com/akarelas/1cddeda23e6a408c31929b3a1993964f

t/mojo/promise.t
akarelas: Since I saw you were already using the preposition "with" in similar -although non-ambiguous- situations inside `mojo/promise.t`, I decided to keep the same preposition, and addressed the ambiguity by renaming "Resove to unsettled promise" to "Settle with unsettled promise" and added the `$p->reject($p2)` case inside that subtest, too. Also I renamed "Resolve to self" to "Resolve with self", as that was already non-ambiguous.

To make sure the case "Reject with self" is covered, I renamed the subtest "Resolve to self" to "Settle with self" and added the "Reject with self" case in there too.

shadowcat-mst:
If you mean to resolve a promise that itself is unsettled,

  $unsettled->resolve(...)

would be "resolve unsettled promise"

If you mean to resolve a promise by passing another unsettled promise,

  $promise->resolve($unsettled)

would be "resolve with unsettled promise"

t/mojo/promise.t
akarelas: shadowcat-mst below provided the answer to your question: "resolve "to" or resolve "with"" (it is "resolve with"). I have thus renamed the subtests, and included more cases in them (with "reject" instead of "resolve").

Anything else you need to accept this PR?

akarelas:
@christopherraa , @jhthorsen I have addressed all the issues you and kraih had with this bugfix.

Could you please look at it again and approve it, if it passes your criteria?

Thanks.

kraih:
I don't see this PR going anywhere. The complexity is just way too high for the (still questionable) value it would add.

akarelas:
> I don't see this PR going anywhere. The complexity is just way too high for the (still questionable) value it would add.

I'd like to remind that it also solves bug #1515. Anyway, thought it would be cool to solve two bugs with 15 lines of code. Your choice.

t/mojo/promise.t
```diff
@@ -368,6 +368,47 @@ subtest 'Settle with promise' => sub {
...
+  is_deeply \@results, [1_000], 'promise resolved';
+
+  $promise = Mojo::Promise->new;
+  (@results, my @errors) = ();
+  $promise->then(sub { @results = @_ }, sub { @errors = @_ });
+  $promise->reject($promise2);
+  Mojo::IOLoop->one_tick;
+  is_deeply \@results, [], 'promise not resolved';
+  is_deeply \@errors, [$promise2], 'correct errors';
+};
```
kraih, line 410: This block of tests seems completely unrelated to the PR description.

t/mojo/promise.t
```diff
@@ -349,7 +349,7 @@ subtest 'All settled (with rejection)' => sub {
   is_deeply \@results, $result, 'promise resolved';
 };
 
-subtest 'Settle with promise' => sub {
+subtest 'Settle with settled promise' => sub {
```
kraih, line 352: These tests are unrelated to your PR, why would you randomly change their description?

kraih (changes requested):
I'll block this PR because it contains random unexplained changes.

kraih:
Has this PR been abandoned?

jhthorsen:
I cannot remember seeing this problem while writing code, so I'm no longer positive to the change. I consider the PR to have academic value, but not something that is frequent enough in real life to be considered.

t/mojo/promise.t
akarelas: Indeed, this block was related to the second issue this PR solves, which was added later. (For history reasons I thought I shouldn't have modified any of the past comments, including the title & description). Now the PR's description has been updated to address that.

t/mojo/promise.t
akarelas: I was under the impression that a patch needs to leave the repo in a consistent state. It would seem strange to have a subtest that would test "Settle with promise", and another one that would test "Settle with unsettled promise". Because, what would the second subtest's reason for existence be, if its case is already being tested by the first subtest? So I did this modification.

Shall I revert that line?

akarelas:
> Has this PR been abandoned?

No, sorry, I was just busy celebrating Christmas. Now I'm back to working.

akarelas:
> I cannot remember seeing this problem while writing code, so I'm no longer positive to the change. I consider the PR to have academic value, but not something that is frequent enough in real life to be considered.

Indeed, the only reason I made this PR was to improve compliance with the ES spec. None of my code would encounter these issues either, unless my code had a bug. I'm okay with you not merging this PR.

t/mojo/promise.t
kraih: PRs trying to resolve multiple issues at once pretty much never get accepted.

kraih:
It seems nobody has ever actually run into these issues and the code complexity is fairly high. So i'm 👎 on this PR.

jberger (changes requested):
After letting this sit this long, I've got to agree. It is a lot of complexity on top of an already complex (of necessity) codebase. If it isn't addressing a real world need (like warning on unhandled errors) i vote no

kraih:
Thanks for your contribution, but i don't see this PR pass a vote.

PR #1562, opened 2020-09-07
Add Envoy Proxy configuration example

### LINKED ISSUE #1543: Cookbook example for Envoy deployment
With Nginx going out of fashion we should add an example for how to use [Envoy](https://www.envoyproxy.io) as a reverse-proxy with Mojolicious to the [cookbook](https://mojolicious.org/perldoc/Mojolicious/Guides/Cookbook#Nginx).

zakame:
This adds a simple reverse proxy example using Envoy.

Fixes #1543.

### DIFF
--- a/lib/Mojolicious/Guides/Cookbook.pod
+++ b/lib/Mojolicious/Guides/Cookbook.pod
@@ -342,10 +342,10 @@ web features, such as WebSockets, are not available.
 
   ScriptAlias / /home/sri/my_app/script/my_app/
 
-=head2 Envoy Proxy
+=head2 Envoy
 
-For deploying L<Mojolicious> applications on "cloud-native" environments that use L<Envoy|https://www.envoyproxy.io>,
-here is a configuration for a basic reverse proxy similar to Apache and Nginx:
+L<Mojolicious> applications can be deployed on "cloud-native" environments that use L<Envoy|https://www.envoyproxy.io>,
+such as with this reverse proxy configuration similar to the Apache and Nginx ones above.
 
   static_resources:
     listeners:
--- a/lib/Mojolicious/Guides/Cookbook.pod
+++ b/lib/Mojolicious/Guides/Cookbook.pod
@@ -386,6 +386,11 @@ such as with this reverse proxy configuration similar to the Apache and Nginx on
           - endpoint:
               address:
                 socket_address: { address: mojo, port_value: 8080 }
+                
+While this configuration works for simple applications, Envoy's typical use case is for implementing proxies of
+applications as a "service mesh" providing advanced filtering, load balancing, and observability features, such as
+seen in L<Istio|https://istio.io/latest/docs/ops/deployment/architecture/>.  More examples can be found at the
+L<Envoy documentation|https://www.envoyproxy.io/docs/envoy/latest/start/start>.
 
 =head2 PSGI/Plack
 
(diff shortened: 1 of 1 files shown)


lib/Mojolicious/Guides/Cookbook.pod
```diff
@@ -342,6 +342,51 @@ web features, such as WebSockets, are not available.
 
   ScriptAlias / /home/sri/my_app/script/my_app/
 
+=head2 Envoy Proxy
```
kraih, line 345: Why does the header need a "Proxy" qualifier? Are there different ways to deploy Mojolicious apps with Envoy?

lib/Mojolicious/Guides/Cookbook.pod
```diff
@@ -342,6 +342,51 @@ web features, such as WebSockets, are not available.
...
+
+For deploying L<Mojolicious> applications on "cloud-native" environments that use L<Envoy|https://www.envoyproxy.io>,
+here is a configuration for a basic reverse proxy similar to Apache and Nginx:
```
kraih, line 348: Other sections don't use the "...:" format for examples.

zakame: Yeah, I'll need to rephrase this :muscle:

lib/Mojolicious/Guides/Cookbook.pod
kraih: Looking at the [README](https://github.com/envoyproxy/envoy) it's just referred to as Envoy and not Envoy Proxy. I don't think the "Proxy" is actually part of the name.

kraih:
Instead of making the example too large with more context information, perhaps there's official Envoy docs we could just link to in a second sentence after the example config?

lib/Mojolicious/Guides/Cookbook.pod
zakame: Doesn't matter now, its already dropped :muscle:

zakame:
Yep, added pointers to Istio and Envoy docs for more information.

I can probably make a fuller-fledged example, but that'll take a while.

lib/Mojolicious/Guides/Cookbook.pod
```diff
@@ -342,6 +342,56 @@ web features, such as WebSockets, are not available.
...
+        cluster_name: local_service
+        endpoints:
+        - lb_endpoints:
+          - endpoint:
+              address:
+                socket_address: { address: mojo, port_value: 8080 }
+                
+While this configuration works for simple applications, Envoy's typical use case is for implementing proxies of
+applications as a "service mesh" providing advanced filtering, load balancing, and observability features, such as
+seen in L<Istio|https://istio.io/latest/docs/ops/deployment/architecture/>.  More examples can be found at the
```
kraih, line 392: Extra whitespace.

kraih: Is "at the documentation" is correct english?

zakame: Would you prefer
```suggestion
seen in L<Istio|https://istio.io/latest/docs/ops/deployment/architecture/>. For more examples, see the
```

zakame: I'll use the above suggestion then, we can revisit it in the new review pass :muscle:

marcusramberg (approved):
Lgtm

Grinnz (approved):
I don't know anything about Envoy, but more cookbook examples are always good.

zakame:
Thanks! :tada:

PR #1564, opened 2020-09-14
add status_(un)like to Test::Mojo

davidcantrell-bb:
### Summary
Adds two functions to Test::Mojo, `status_like` and `status_unlike`.

### Motivation
While working on a hook that looks for one particular class of error, I can easily test that the hook fires and spits out the appropriate HTTP status code by crafting some dodgy input and checking `status_is(456)`. But I also want to test that it *doesn't* fire when the input is OK (as far as that hook is concerned), and instead control passes on to the next hook or to the controller, which will whine about its input in some different way but I don't care which species of whining I happen to get as long as it's a different error. So I want `status_isnt(456)` but to still check that the status is some species of HTTP error code.

Hence `status_like(qr/^4/)`. I also implemented its inverse `status_unlike` even though I have no immediate need for it.

### DIFF
--- a/lib/Test/Mojo.pm
+++ b/lib/Test/Mojo.pm
@@ -299,6 +299,18 @@ sub status_isnt {
   return $self->test('isnt', $self->tx->res->code, $status, $desc);
 }
 
+sub status_like {
+  my ($self, $status, $desc) = @_;
+  $desc = _desc($desc, $self->tx->res->code . " like $status");
+  return $self->test('like', $self->tx->res->code, $status, $desc);
+};
+
+sub status_unlike {
+  my ($self, $status, $desc) = @_;
+  $desc = _desc($desc, $self->tx->res->code . " unlike $status");
+  return $self->test('unlike', $self->tx->res->code, $status, $desc);
+};
+
 sub test {
   my ($self, $name, @args) = @_;
   local $Test::Builder::Level = $Test::Builder::Level + 3;
@@ -1070,6 +1082,20 @@ Check response status for exact match.
   $t = $t->status_isnt(200, 'different status');
 
 Opposite of L</"status_is">.
+
+=head2 status_like
+
+  $t = $t->status_like(qr/^4/);
+  $t = $t->status_like(qr/^4/, 'some kind of error');
+
+Check response status for regex match.
+
+=head2 status_unlike
+
+  $t = $t->status_unlike(qr/^3/);
+  $t = $t->status_unlike(qr/^3/, 'not any kind of redirect');
+
+Check response status for B<not> matching a regex.
 
 =head2 test
 
--- a/lib/Test/Mojo.pm
+++ b/lib/Test/Mojo.pm
@@ -1095,7 +1095,7 @@ Check response status for regex match.
   $t = $t->status_unlike(qr/^3/);
   $t = $t->status_unlike(qr/^3/, 'not any kind of redirect');
 
-Check response status for B<not> matching a regex.
+Opposite of L</"status_like">.
 
 =head2 test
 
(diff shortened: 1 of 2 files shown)


lib/Test/Mojo.pm
```diff
@@ -1071,6 +1083,20 @@ Check response status for exact match.
...
+  $t = $t->status_like(qr/^4/, 'some kind of error');
+
+Check response status for regex match.
+
+=head2 status_unlike
+
+  $t = $t->status_unlike(qr/^3/);
+  $t = $t->status_unlike(qr/^3/, 'not any kind of redirect');
+
+Check response status for B<not> matching a regex.
```
kraih, line 1098: Seems inconsistent with other method descriptions of the same type.

t/test/mojo.t
```diff
@@ -11,6 +11,8 @@ my $t = Test::Mojo->new;
 subtest 'Basics' => sub {
   isa_ok $t->app, 'Mojolicious', 'right class';
   $t->get_ok('/')->status_is(200)->content_is('Hello Test!');
+  $t->status_like(qr/^2/);
+  $t->status_unlike(qr/^4/);
```
kraih, line 15: Not actual tests.

kraih:
Not really a fan of using a regex to match numbers. Other similar APIs we already have work [differently](https://docs.mojolicious.org/Mojo/Message/Response#is_server_error).

lib/Test/Mojo.pm
DrHyde: Fixed

t/test/mojo.t
DrHyde: How so? They pass or fail depending on what data is in `$t` and the regex you pass in.

DrHyde:
> Not really a fan of using a regex to match numbers. Other similar APIs we already have work [differently](https://docs.mojolicious.org/Mojo/Message/Response#is_server_error).

Those are great if you're only interested in the few pre-selected options, and if you want a Boolean return value instead of something that is a test, which you can chain with all the other testing functions.

t/test/mojo.t
kraih: All new `Test::Mojo` methods require their own `subtest` section below. With tests for all argument variations.

jhthorsen:
I think it sounds like a strange usecase not wanting to check for the exact status code.

t/test/mojo.t
DrHyde: OK

DrHyde:
> I think it sounds like a strange usecase not wanting to check for the exact status code.

I suppose it is a bit strange, but it makes sense in my case.

Grinnz:
I agree with @kraih that it would make more sense to provide tests for the category of status code similar to Mojo::Message::Response; this is better for your provided use case anyway. It would be more verbose and require more test methods, but much clearer.

jberger:
I'll also echo what Grinnz says about categories, but I'll take it a step further and remind everyone in the discussion that it is already possible to just inspect the response directly and use Test::More for it.

```perl
$t->get_ok($url);
like $t->tx->res->code, /^4/;
```

and also if we cannot settle on an api, this is a fine case for a role on cpan.

DrHyde:
> and also if we cannot settle on an api, this is a fine case for a role on cpan

Ah-ha, I didn't know about those. I think that even if we *can* settle on an API it would be a fine case for a role. So I think I'll go away and do that, calling it ... hmmm ... `Test::Mojo::Role::FuzzyStatus`? `...::StatusGroups`? `...::StatusCategories`? ... and implement `status_is_client_error` and so on as well as a more generic status_(un)like.

Anyone have any better ideas for naming it?

Grinnz:
Test::Mojo::Role::HTTPStatus ?

DrHyde:
> Test::Mojo::Role::HTTPStatus ?

Nice and obvious, so easy to find. Thanks!

jberger:
I'm glad I could offer a useful alternative! Cheers!

DrHyde:
FYI https://github.com/DrHyde/perl-modules-Test-Mojo-Role-HTTPstatus

Issue #1563, opened 2020-09-10
Unexpected finish emitted when using subprocess


HEM42:
* Mojolicious version: 8.58
* Perl version: 5.32.0
* Operating system: MacOS v10.15.6

### Steps to reproduce the behavior
I have a route which triggers some longer running IOLoop subprocesses, for which I return promises. I have a subsystem I need to log into at the begining, somewhere along the code I can abort due to other errors, and I always need to logout of this subsystem when the transaction with the user is over.
For this I use `$c->on( finish => sub{} );`

The problem is that `->on( finish` is triggered for each subprocess running, and I dont want to logout of the subsystem before everything is done / or some error occurred.

```perl
#!/usr/bin/env perl

use Mojo::Base -strict;
use Test::Mojo;
use Test::More;

use Mojo::Promise;
use Mojolicious::Lite -signatures;

get '/' => sub ($c) {
    $c->render_later;
    $c->on(
        finish => sub {
            say 'FINISHED';
            # $c->subsystem->logout;
        }
    );

    # $c->subsystem->login; to some subsystem

    my $p1 = another_long_running_subprocess();
    my $p2 = another_long_running_subprocess();
    my $p3 = another_long_running_subprocess();

    Mojo::Promise->all( $p1, $p2, $p3 )->then(
        sub (@results) {
            # manage results
            $c->render( text => 'Hello World!' );
        }
    )->wait;
};

my $t = Test::Mojo->new;

$t->get_ok('/')->status_is(200);

done_testing;

sub another_long_running_subprocess
{
    my $subprocess = Mojo::IOLoop::Subprocess->new;

    my $promise = $subprocess->run_p(
        sub {
            my $result = `ping -c 5 1.1.1.1`;
            return $result;
        },
    );

    return $promise;
}
```

### Expected behavior
I would not expect to see more one `finish` emitted

### Actual behavior
Instead of a single `finish`, one from each subprocess is also emitted

Output from the above code example
```
FINISHED
FINISHED
FINISHED
FINISHED
ok 1 - GET /
ok 2 - 200 OK
1..2
```

kraih:
Why are you not using `Mojo::IOLoop->subprocess`?

HEM42:
> Why are you not using `Mojo::IOLoop->subprocess`?

No specific reason, but changing the code still produce the same result

```perl
sub another_long_running_subprocess
{
    Mojo::IOLoop->subprocess->run_p(
        sub {
[...]

PR #1565, opened 2020-09-15
Unsubscribe close events for streams when resetting ioloop

kiwiroy:
### Summary
No longer trigger close event for streams when resetting the ioloop.

### Motivation
This fixes the bug where child processes have such cleanup events triggered.

### References
- #1563

### DIFF
--- a/lib/Mojo/IOLoop.pm
+++ b/lib/Mojo/IOLoop.pm
@@ -95,6 +95,7 @@ sub remove {
 
 sub reset {
   my $self = _instance(shift)->emit('reset');
+  $self->stream($_)->unsubscribe('close') for (keys %{$self->{in}}, keys %{$self->{out}});
   delete @$self{qw(accepting acceptors events in out stop)};
   $self->reactor->reset;
   $self->stop;
--- a/t/mojo/ioloop.t
+++ b/t/mojo/ioloop.t
@@ -83,18 +83,19 @@ subtest 'Recurring timer' => sub {
 };
 
 subtest 'Handle and reset' => sub {
-  my ($handle, $handle2, $reset);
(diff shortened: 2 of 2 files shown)


lib/Mojo/IOLoop.pm
```diff
@@ -95,6 +95,8 @@ sub remove {
 
 sub reset {
   my $self = _instance(shift)->emit('reset');
+  $self->stream($_)->unsubscribe('close') for keys %{$self->{in}};
+  $self->stream($_)->unsubscribe('close') for keys %{$self->{out}};
```
kraih, line 99: Why is this not just one line?

kiwiroy: Compressed.

kraih:
I'm not sure if this is the right way to deal with the problem, but so far this is the best proposal.

kraih:
The code seems fine, but i don't know if this is the right solution. We'll have to think carefully about what use cases we want `Mojo::IOLoop->reset` for.

jberger:
I've been trying to think of any useful situation wherein you might want to get close events when resetting but none are coming to mind. I was also trying to think of a useful way to invert the logic and have the stream subscribe to the loop's reset event and use that to unsub from close, but there still wouldn't be a good way to target the unsub to override it, plus it would be more convoluted. I think I'm slightly +1 on this but it should have documentation in reset saying that stream close events will not be emitted (and possibly that they'll be removed).

kraih:
This might end up being a case where i let this PR sit for a bit and apply it if nobody finds a good reason not to do it.

kraih:
Looks like nobody objected.

kraih:
I think there might actually be a risk of connection leaks here. So we might have to revert this PR.

PR #1570, opened 2020-10-10
Convert t/mojo/json_pointer.t to use subtests

dz-at-tc:
### Summary
Convert t/mojo/json_pointer.t to use subtests.

### Motivation
One more file converted.

### References
Issue #1520

### DIFF
--- a/t/mojo/json_pointer.t
+++ b/t/mojo/json_pointer.t
@@ -3,94 +3,106 @@ use Mojo::Base -strict;
 use Test::More;
 use Mojo::JSON::Pointer;
 
-# "contains" (hash)
-my $pointer = Mojo::JSON::Pointer->new({foo => 23, '' => 24});
-ok $pointer->contains(''),     'contains ""';
-ok $pointer->contains('/'),    'contains "/"';
-ok $pointer->contains('/foo'), 'contains "/foo"';
-ok !$pointer->contains('foo'),  '"foo" is invalid';
-ok !$pointer->contains('/bar'), 'does not contains "/bar"';
-ok $pointer->new({foo => {bar => undef}})->contains('/foo/bar'), 'contains "/foo/bar"';
-
-# "contains" (string)
-$pointer = Mojo::JSON::Pointer->new('works');
-ok $pointer->contains(''), 'contains ""';
-ok !$pointer->contains('/'),    'does not contain "/"';
-ok !$pointer->contains('/foo'), 'does not contain "/foo"';
-ok $pointer->new('0')->contains(''), 'contains ""';
-
-# "contains" (mixed)
-$pointer = Mojo::JSON::Pointer->new({foo => [0, 1, 2]});
-ok $pointer->contains(''),       'contains ""';
-ok $pointer->contains('/foo/0'), 'contains "/foo/0"';
-ok !$pointer->contains('/foo/9'),   'does not contain "/foo/9"';
-ok !$pointer->contains('/foo/bar'), 'does not contain "/foo/bar"';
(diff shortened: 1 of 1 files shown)


dz-at-tc:
Note that I was not 100% sure how you wanted to handle the various "get" subtests that share the $pointer variable across subtests.  At least for now I left the sharing in place and created nested subtests (though I'm certainly not in love with the name "[shared pointer]"...  patches welcome!).

Actually, as I write this it finally dawned on me that all I really had to do was factor out the variable, not add a nested subtest...  Mea culpa, I will fix that now and re-push shortly...

t/mojo/json_pointer.t
```diff
@@ -3,94 +3,105 @@ use Mojo::Base -strict;
...
-$pointer = Mojo::JSON::Pointer->new({foo => 'bar', '' => 'baz'});
-is_deeply $pointer->get(''), {foo => 'bar', '' => 'baz'}, '"" is "{foo => "bar", "" => "baz"}"';
-is $pointer->get('/'),    'baz', '"/" is "baz"';
-is $pointer->get('/foo'), 'bar', '"/foo" is "bar"';
-is $pointer->get('foo'),  undef, '"foo" is invalid';
-is $pointer->new({foo => {bar => 42}})->get('/foo/bar'), 42, '"/foo/bar" is "42"';
-is_deeply $pointer->new({foo => {23 => {baz => 0}}})->get('/foo/23'), {baz => 0}, '"/foo/23" is "{baz => 0}"';
-is $pointer->new({foo => {'' => 42}})->get('/foo/'), 42, '"/foo/" is "42"';
-is $pointer->new({foo => {'' => {'' => 42}}})->get('/foo//'), 42, '"/foo//" is "42"';
+my $pointer;
```
kraih, line 34: Why is this before the block?

dz-at-tc: That's what I was trying express in my prior comment, namely that the "get" tests appear to share the $pointer variable across different subsections.  For example, note how the first line in the '"get" (mixed)' section runs is_deeply.  Thus, I pulled the variable out to reflect that sharing.  Is there a better way to handle this?

Grinnz: It would be better to have each subtest set up $pointer again as it needs. There isn't anything that actually needs to be shared.

dz-at-tc: Sounds good, I will have another go at this then, thanks!

t/mojo/json_pointer.t
```diff
@@ -3,94 +3,106 @@ use Mojo::Base -strict;
...
 
-# "get" (mixed)
-is_deeply $pointer->new({foo => {bar => [1, 2, 3]}})->get('/foo/bar'), [1, 2, 3], '"/foo/bar" is "[1, 2, 3]"';
-$pointer = Mojo::JSON::Pointer->new({foo => {bar => [0, undef, 3]}});
-is $pointer->get('/foo/bar/0'), 0,     '"/foo/bar/0" is "0"';
-is $pointer->get('/foo/bar/1'), undef, '"/foo/bar/1" is undef';
-is $pointer->get('/foo/bar/2'), 3,     '"/foo/bar/2" is "3"';
-is $pointer->get('/foo/bar/6'), undef, '"/foo/bar/6" is undef';
+subtest '"get" (mixed)' => sub {
+  my $pointer = Mojo::JSON::Pointer->new();
```
kraih, line 55: Parentheses are inconsistent.

dz-at-tc: Drat!  Sorry, fixing now...

kraih (approved):

jhthorsen (approved):

PR #1578, opened 2020-10-17
Replace prettify.js with highlight.js

### LINKED ISSUE #1544: Replace prettify.js with something maintained
pretty

For syntax highlighting on the exception page we still use [prettify.js](https://github.com/googlearchive/code-prettify), which like most Google projects has been abandoned. So we should really replace it with something better, like [highlight.js](https://highlightjs.org).

There was actually [an attempt](https://github.com/mojolicious/mojo/pull/737) before, but it failed because of interdependencies with [mojolicious.org](https://mojolicious.org). That is no longer a problem, because assets are not shared anymore.

For testing you can use the included example app `examples/responses.pl` (`perl -Ilib examples/responses.pl daemon`). The screenshot above was taken of a request for `http://127.0.0.1:3000/res5`.

zakame:
### Summary

This replaces the old prettify.js highlighter with highlight.js.

### Motivation

prettify.js has long been abandoned by Google so we need to replace it.

That said, highlight.js is also somewhat dormant (https://github.com/highlightjs/highlight.js/issues/1678) but still sees enough activity to get around.

### References

This is an update of #737, fixes #1544.

Most of the work was already done by @dotandimet, please give credit where due. I just updated it for latest Mojo and highlight.js :muscle:


kraih:
The commit history is all messed up.

kraih (dismissed):
Blocked until the commit history is fixed.

zakame:
@kraih thanks, I rebased against latest master now and squashed it all to a single commit :muscle:

lib/Mojolicious/resources/public/mojo/highlight.js/highlight-mojo-light.css
```diff
@@ -0,0 +1,141 @@
+/*
```
kraih, line 1: Why does this file exist?

lib/Mojolicious/resources/public/mojo/highlight.js/highlight-mojo-dark.css
```diff
@@ -0,0 +1,107 @@
...
+
+.hljs {
+/*
+  border:0;padding-bottom:0.75em;padding-top:0.75em;
+  display: block;
+  overflow-x: auto;
+  padding: 0.5em;
+  background: #444;
+  -webkit-text-size-adjust: none; */
+}
```
kraih, line 16: Why is this commented out?

lib/Mojolicious/resources/public/mojo/highlight.js/highlight-mojo-dark.css
```diff
@@ -0,0 +1,107 @@
+﻿/*
+
+Based on Dark style by Ivan Sagalaev
+and prettify-mojo-dark.css by Sebastian Reidel
+
+*/
```
kraih, line 6: That comment seems redundant.

lib/Mojolicious/resources/public/mojo/highlight.js/highlight-mojo-light.css
zakame: Yep, this shouldn't be here, removed just now.

lib/Mojolicious/resources/public/mojo/highlight.js/highlight-mojo-dark.css
zakame: Yeah, can probably drop this.

lib/Mojolicious/resources/public/mojo/highlight.js/highlight-mojo-dark.css
zakame: This was from the old PR, we can drop this too.

lib/Mojolicious/resources/public/mojo/highlight.js/highlight-mojo-dark.css
```diff
@@ -0,0 +1,90 @@
...
+.hljs-title,
+.hljs-type,
+.ini .hljs-title,
+.hljs-tag .hljs-value,
+.css .hljs-rules .hljs-value,
+.hljs-preprocessor,
+.hljs-pragma,
+.ruby .hljs-symbol,
+.ruby .hljs-symbol .hljs-string,
+.ruby .hljs-class .hljs-parent,
```
zakame, line 26: I think we can drop these class names for other languages, per the current version and git history of https://github.com/highlightjs/highlight.js/blob/master/src/styles/dark.css

zakame: After a closer look, its probably better to recreate this...

zakame: Updated, it should better highlight some keywords now:

image

t/mojolicious/exception_lite_app.t
```diff
@@ -257,8 +257,9 @@ ok !$snapshot->{exception}, 'no exception in snapshot';
 
 # Bundled static files
 $t->get_ok('/mojo/jquery/jquery.js')->status_is(200)->content_type_is('application/javascript');
-$t->get_ok('/mojo/prettify/run_prettify.js')->status_is(200)->content_type_is('application/javascript');
-$t->get_ok('/mojo/prettify/prettify-mojo-dark.css')->status_is(200)->content_type_is('text/css');
+$t->get_ok('/mojo/highlight.js/highlight.min.js')->status_is(200)->content_type_is('application/javascript');
+$t->get_ok('/mojo/highlight.js/mojolicious.min.js')->status_is(200)->content_type_is('application/javascript');
+$t->get_ok('/mojo/highlight.js/highlight-mojo-dark.css')->status_is(200)->content_type_is('text/css');
```
zakame, line 262: These need updating if #1579 gets ahead.

kraih (approved):
That looks much better now.

lib/Mojolicious/resources/public/mojo/highlight.js/highlight-mojo-dark.css
zakame: The [base theme's history](https://github.com/highlightjs/highlight.js/commits/master/src/styles/dark.css) suggest that the main style classes are already set, what's only changed/removed are for language-specific styles that crept into the base theme during its development.  I doubt this will change any further given the last commit for that base theme is from 2015.

t/mojolicious/exception_lite_app.t
zakame: Updated :muscle:

jhthorsen (commented):
I think this is a good change in general, but I am curious if we need to support all the languages in highlight.js/highlight.js.min. Limiting the supported languages to bash, css, diff, http, javascript, json, makefile, xml, perl, plaintext, shell, and sql slims down the file from 100k to 45k. Maybe we can limit it even further?

zakame:
Good idea, we can probably start from that set then drill down further if need?  Maybe we can trim Makefile and diff out as well (since I suspect we won't be able to see these under debug.html anyway?)

I'll also update again to latest version, 10.3.1 was just released.

jberger (approved):
I'm a fan of highlight.js and use it in other things.

jberger (approved):

jhthorsen (approved):
I think the new slimmed down version of the language pack is great, so I'm in favor of this change now.

zakame:
Thanks again everyone! :tada:

PR #1579, opened 2020-10-18
Convert t/mojolicious/exception_lite_app.t to use subtests

zakame:
### Summary

As the title says.

### Motivation

I was already in the area for #1578 so might as well convert the test!

### References

For #1520.

### DIFF
--- a/t/mojolicious/exception_lite_app.t
+++ b/t/mojolicious/exception_lite_app.t
@@ -126,149 +126,186 @@ get '/dead_helper';
 
 my $t = Test::Mojo->new;
 
-# Missing error
-my $c = $t->app->build_controller;
-$c->reply->exception(undef);
-like $c->res->body, qr/Exception!/, 'right result';
-$c = $t->app->build_controller;
-$c->reply->exception;
-like $c->res->body, qr/Exception!/, 'right result';
-$c = $t->app->build_controller;
-$c->reply->exception(Mojo::Exception->new);
-like $c->res->body, qr/Exception!/, 'right result';
+subtest 'Missing error' => sub {
+  my $c = $t->app->build_controller;
+  $c->reply->exception(undef);
+  like $c->res->body, qr/Exception!/, 'right result';
 
-# Debug
-$t->get_ok('/logger?level=debug&message=one')->status_is(200)->content_is('debug: one');
-like $log, qr/debug:one/, 'right result';
+  $c = $t->app->build_controller;
+  $c->reply->exception;
+  like $c->res->body, qr/Exception!/, 'right result';
 
-# Info
-$t->get_ok('/logger?level=info&message=two')->status_is(200)->content_is('info: two');
-like $log, qr/info:two/, 'right result';
+  $c = $t->app->build_controller;
+  $c->reply->exception(Mojo::Exception->new);
+  like $c->res->body, qr/Exception!/, 'right result';
+};
 
-# Warn
-$t->get_ok('/logger?level=warn&message=three')->status_is(200)->content_is('warn: three');
-like $log, qr/warn:three/, 'right result';
+subtest Debug => sub {
+  $t->get_ok('/logger?level=debug&message=one')->status_is(200)->content_is('debug: one');
+  like $log, qr/debug:one/, 'right result';
+};
 
-# Error
-$t->get_ok('/logger?level=error&message=four')->status_is(200)->content_is('error: four');
-like $log, qr/error:four/, 'right result';
+subtest Info => sub {
+  $t->get_ok('/logger?level=info&message=two')->status_is(200)->content_is('info: two');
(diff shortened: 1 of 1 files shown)


t/mojolicious/exception_lite_app.t
```diff
@@ -126,149 +126,213 @@ get '/dead_helper';
...
+  ok !$snapshot, 'no snapshot';
+
+  $t->get_ok('/reuse/exception')->status_is(500)->content_like(qr/Reusable exception/);
+
+  isa_ok $exception, 'Mojo::Exception', 'right exception';
+
+  like $exception, qr/Reusable exception/, 'right message';
+
+  is $snapshot->{foo}, 'bar', 'right snapshot value';
+
```
kraih, line 305: Those empty lines don't add any value.

zakame: Was just following your exact direction:

>Block comments become the subtest description, and then we add a blank line between the different test cases within the subtest.

I can revert to no spaces, no problem.

zakame: @kraih should I undo the extra newlines under `Bundled static files` subtest as well?

kraih: Updated the sentence, since you've taken it a bit too literal. And yes, blank lines only if there's a pattern to follow, like 3 repetitive groups of similar tests (as is the case in the #1579 example).

zakame: Thanks, I've adjusted the newlines based on context, let me know if that's better or if I should adjust more :muscle:

t/mojolicious/exception_lite_app.t
```diff
@@ -126,149 +126,199 @@ get '/dead_helper';
...
-$t->get_ok('/mojo/pinstripe-dark.png')->status_is(200)->content_type_is('image/png');
-$t->get_ok('/mojo/pinstripe-light.png')->status_is(200)->content_type_is('image/png');
+subtest 'Dead included template' => sub {
+  $t->get_ok('/dead_included_template')->status_is(500)->content_like(qr/dead template!/)->content_like(qr/line 1/);
+};
+
+subtest 'Dead template with layout' => sub {
+  $t->get_ok('/dead_template_with_layout')->status_is(500)->content_like(qr/dead template with layout!/)
+    ->content_like(qr/line 2/)->content_unlike(qr/Green/);
+
```
kraih, line 216: These two tests really belong together.

t/mojolicious/exception_lite_app.t
```diff
@@ -126,149 +126,199 @@ get '/dead_helper';
...
+  like $log, qr/dead template with layout!/, 'right result';
+};
+
+subtest 'Dead action' => sub {
+  $t->get_ok('/dead_action')->status_is(500)->content_type_is('text/html;charset=UTF-8')
+    ->content_like(qr!get &#39;/dead_action&#39;!)->content_like(qr/dead action!/)
+    ->text_is('#error' => "dead action!\n");
+
```
kraih, line 224: Same here.

t/mojolicious/exception_lite_app.t
```diff
@@ -126,149 +126,199 @@ get '/dead_helper';
...
+subtest 'Reuse exception' => sub {
+  ok !$exception, 'no exception';
+  ok !$snapshot, 'no snapshot';
+
+  $t->get_ok('/reuse/exception')->status_is(500)->content_like(qr/Reusable exception/);
+  isa_ok $exception, 'Mojo::Exception', 'right exception';
+  like $exception, qr/Reusable exception/, 'right message';
+
+  is $snapshot->{foo}, 'bar', 'right snapshot value';
+  ok !$snapshot->{exception}, 'no exception in snapshot';
```
kraih, line 302: All of these also belong together.

t/mojolicious/exception_lite_app.t
zakame: Ok, let's get rid of that line...

t/mojolicious/exception_lite_app.t
zakame: I take this is the same for the other `Dead XXX` subtests?

zakame: And perhaps for the logger level subtests as well?

zakame: I went ahead and just removed all extra newlines for those subtests...

t/mojolicious/exception_lite_app.t
zakame: Done, removed newlines.

kraih (approved):

Grinnz (approved):

zakame:
Thanks everyone! :+1:

PR #1590, opened 2020-10-29
Bump Highlight.js to 10.3.2

zakame:
### Summary

While working on https://github.com/mojolicious/mojolicious.org/pull/22 I was alerted about https://github.com/highlightjs/highlight.js/releases/tag/10.3.2

As we're using a custom build that's supposed to include JavaScript, we rebuild this here from https://highlightjs.org/download/ to make sure.

### Motivation

For comparison, here's our current highlight.js version (10.3.1), mistaking the JS snippet for `bash`:

image

And here is the latest (10.3.2) which fixes it:

image

The highlighting itself does not change though, but I figure it is better to ensure code blocks are properly recognized.

### References

#1578 

For additional record, the highlight.js custom build contains support for these languages:

- bash
- CSS
- diff
- HTML, XML
- JSON
- JavaScript
- Perl
- plain text
- SQL

### DIFF
--- a/lib/Mojolicious/resources/public/mojo/highlight.js/highlight.min.js
+++ b/lib/Mojolicious/resources/public/mojo/highlight.js/highlight.min.js
@@ -1,5 +1,5 @@
 /*
-  Highlight.js 10.3.1 (3797c108)
+  Highlight.js 10.3.2 (31e1fc40)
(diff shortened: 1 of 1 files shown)


kraih (approved):
Great PR description.

jberger (approved):

Issue #1488, opened 2020-03-31
Better test coverage for Test::Mojo


kraih:
After [this patch](https://github.com/mojolicious/mojo/compare/4f1fb79883e5...3723973fd500#diff-ee9e1eca670f561769d858a66424416e) has been merged, all the test methods in `Test::Mojo` should be added to `t/test/mojo.t`.

kraih:
You don't have to add tests for all functions at once, but maybe combine at least related methods into one PR. Like all `content_*` methods.

PR #1591, opened 2020-10-30
Added tests for Test::Mojo::content_* methods

rage311:
### Summary
Added Test::Mojo::content_* tests.  If it seems like I'm on the right path here, I can continue adding tests for more methods.

### Motivation
Part of addressing #1488 

### References
#1488

### DIFF
--- a/t/test/mojo.t
+++ b/t/test/mojo.t
@@ -83,6 +83,60 @@ subtest 'content_is' => sub {
   is_deeply \@args, ['is', 'Hello Test!', 'Hello Test!', 'some description'], 'right result';
 };
 
+subtest 'content_isnt' => sub {
+  $t->content_isnt('Goodbye Test!');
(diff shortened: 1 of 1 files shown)


kraih (changes requested):
Please squash your commits.

kraih (approved):

jhthorsen (approved):
Looks good to me 👍

rage311:
Sorry to be a bother, as I realize this isn't your problem, but it seems these pull requests weren't counted toward Hacktoberfest as the hacktoberfest label was removed from the mojo project and they must not have done their validity checks before that happened.  In lieu of re-adding the hacktoberfest label to the project, it seems you can also add a `hacktoberfest-accepted` label to this pull request, which should also make it count.  If there's any chance that someone could add the `hacktoberfest-accepted` label to my 3 pull requests (#1592 and #1593 too), I would be very grateful.

Issue #1599, opened 2020-11-04
Mojo::Util secure_compare still leaks timing information


robrwo:
* Mojolicious version: 8.63
* Perl version: 5.28.1
* Operating system: Linux 4.15.0-122-generic 124-Ubuntu

### Steps to reproduce the behavior

I would expect the cumulative or code to short-circuit comparisons once r is true.
```perl
$r |= ord(substr $one, $_) ^ ord(substr $two, $_)
```

I wrote a rudimentary benchmark:
```perl
use strict;
use warnings;

use Benchmark;

sub secure_compare {
  my ($one, $two) = @_;
  return undef if length $one != length $two;
  my $r = 0;
  $r |= ord(substr $one, $_) ^ ord(substr $two, $_) for 0 .. length($one) - 1;
  return $r == 0;
}


timethese( 5_000_000, {
    diff1 => sub { secure_compare("aaaaaaaaaa", "baaaaaaaaa") },
    diff5 => sub { secure_compare("aaaaaaaaaa", "aaaaabaaaa") },
    diff9 => sub { secure_compare("aaaaaaaaaa", "aaaaaaaaab") },
    same  => sub { secure_compare("aaaaaaaaaa", "aaaaaaaaaa") },
});

```

It seems that the latter a difference occurs in a string, the longer the comparison takes:
```
Benchmark: timing 5000000 iterations of diff1, diff5, diff9, same...
     diff1:  9 wallclock secs ( 8.50 usr +  0.00 sys =  8.50 CPU) @ 588235.29/s (n=5000000)
     diff5:  9 wallclock secs ( 8.57 usr +  0.00 sys =  8.57 CPU) @ 583430.57/s (n=5000000)
     diff9:  8 wallclock secs ( 8.80 usr +  0.00 sys =  8.80 CPU) @ 568181.82/s (n=5000000)
      same: 10 wallclock secs ( 8.84 usr +  0.00 sys =  8.84 CPU) @ 565610.86/s (n=5000000)
```

So this does leak information about the strings.

### Expected behavior

There should be no difference in the timing.

### Actual behavior

The latter a difference occurs in a string, the longer the check takes.

robrwo:
Note running tests on different machines (different Linux versions) and different versions of Perl, the timings are less consistent, but generally diff1 takes less time than diff5 or diff9 or same.

robrwo:
[...]

PR #1601, opened 2020-11-07
Mojo::Util#secure_compare fix leak of string length #1599

robrwo:
By immediately returning when the two strings are not the
same length, the function allows an attacker to guess the
length of the secret string using timing attacks.

This change uses a fix by @jberger, based on [1].

This also updates the documentation to emphasize that the
secret string should be the second argument (although it's
not critical).

[1] https://github.com/vadimdemedes/secure-compare/blob/master/index.js

### Summary

This fixes a potential leak of the length of the secret string using timing attacks.

### Motivation

The secure_compare function is not as secure as claimed.

### References

See discussion #1599.

### DIFF
--- a/lib/Mojo/Util.pm
+++ b/lib/Mojo/Util.pm
@@ -276,8 +276,8 @@ sub scope_guard { Mojo::Util::_Guard->new(cb => shift) }
 
 sub secure_compare {
   my ($one, $two) = @_;
-  return undef if length $one != length $two;
-  my $r = 0;
+  my $r = length $one != length $two;
+  $two = $one if $r;
   $r |= ord(substr $one, $_) ^ ord(substr $two, $_) for 0 .. length($one) - 1;
   return $r == 0;
 }
@@ -792,7 +792,8 @@ Create anonymous scope guard object that will execute the passed callback when t
 
   my $bool = secure_compare $str1, $str2;
 
-Constant time comparison algorithm to prevent timing attacks.
+Constant time comparison algorithm to prevent timing attacks. The secret string should be the second argument, to
+avoid leaking information about the length of the string.
 
 =head2 sha1_bytes
 


Changes
```diff
@@ -1,5 +1,6 @@
 
 8.65  2020-11-06
+  - Improve secure_compare to avoid leaking the length of the string using timing attacks. (jberger, rrwo)
```
kraih, line 3: Please leave the Changes entry to me.

lib/Mojo/Util.pm
```diff
@@ -790,9 +790,9 @@ Create anonymous scope guard object that will execute the passed callback when t
 
 =head2 secure_compare
 
-  my $bool = secure_compare $str1, $str2;
+  my $bool = secure_compare $str, $secret;
```
kraih, line 793: The choice of variable names tells me nothing.

robrwo: I've updated the POD accordingly.

Changes
robrwo: Removed from the PR

lib/Mojo/Util.pm
```diff
@@ -790,9 +790,12 @@ Create anonymous scope guard object that will execute the passed callback when t
...
 
-Constant time comparison algorithm to prevent timing attacks.
+Constant time string comparison algorithm to prevent timing attacks.
+
+The secret string should be the second argument, to avoid leaking
+information about the length of the string.
```
kraih, line 798: That formatting is inconsistent with the rest of the docs i believe.

lib/Mojo/Util.pm
```diff
@@ -794,6 +794,8 @@ Create anonymous scope guard object that will execute the passed callback when t
 
 Constant time comparison algorithm to prevent timing attacks.
 
+The secret string should be the second argument, to avoid leaking information about the length of the string.
```
kraih, line 797: This is still inconsistently formatted.

robrwo: Apologies. I don't understand how you want it formatted. Is there a style guide somewhere, or am I missing something obvious here?

kraih: Just look at the surrounding documentation. There is no newline between sentences.

lib/Mojo/Util.pm
```diff
@@ -792,7 +792,7 @@ Create anonymous scope guard object that will execute the passed callback when t
 
   my $bool = secure_compare $str1, $str2;
 
-Constant time comparison algorithm to prevent timing attacks.
+Constant time comparison algorithm to prevent timing attacks. The secret string should be the second argument, to avoid leaking information about the length of the string.
```
kraih, line 795: The line is longer than 120 characters now.

jberger:
Fix the long line and I'm ok with this.

lib/Mojo/Util.pm
robrwo: Ok. Fixed. 

I normally keep lines under 80 characters but wrongly assumed because of long lines for other sections that you weren't wrapping lines.  I think a style guide would be useful, or (better yet) add a POD test that enforces style.

marcusramberg (approved):
lgtm

jberger (approved):

lib/Mojo/Util.pm
Grinnz: The practical guide is that there is a perltidy config in the repo which should be used for all contributions and code/doc styles in use should be adhered to when appropriate.

PR #1609, opened 2020-11-16
use different return values for min/max validation checks

tim-2:
### Motivation
Sometimes it can be useful to know which range check in num/size validation checks failed.
For example if you want to display a detailed error message to the user in a html form.

```
$v->required('num')->num(1, 3);
if(my $error = $v->error('num')) {
    my $result = $error->[1];
    my $reason = $result == 1 ? 'no number' :
                 $result == 2 ? 'value < min' :
                 $result == 3 ? 'value > max' : die;
    say "num validation failed because: $reason";
}
```


jhthorsen (commented):
I don't like this change too much. If anything, I think we should return `-1` and `1` and maybe `0e0` for `$value !~ /^-?[0-9]+$/`.

For the record: I'm not sure if I like my suggestion too much either.

kraih:
Agree with @jhthorsen. This is a 👎 for me.

kraih:
Thanks for your contribution, but i don't see this PR pass a vote.

Issue #1610, opened 2020-11-17
Inconsistency with commands that use dash.


kiwiroy:
The default [hint](https://github.com/mojolicious/mojo/blob/0a8e6f2a989903a4143a7c5e302c18798499703c/lib/Mojolicious/Commands.pm#L10) `APPLICATION help COMMAND` does not find the underscore named module, since the switch to using a dash rather than an underscore in 863f97b. I feel the following line should be after the `help` detection, rather than before.

https://github.com/mojolicious/mojo/blob/0a8e6f2a989903a4143a7c5e302c18798499703c/lib/Mojolicious/Commands.pm#L38

## Expected

Help is displayed in the terminal for the command.

## Actual

```
./script/myapp help kebab-time
Unknown command "kebab-time", maybe you need to install it?
```
Also, why is the issue template not applied here?

kraih:
This is actually much easier demonstrated with
```
$ mojo generate help lite-app
Unknown command "lite-app", maybe you need to install it?
```

PR #1611, opened 2020-11-17
Display help for dash commands as per hint

kiwiroy:
### Summary
Resolves the bug where help is not shown for commands with a dash.

### Motivation
Consistency

### References
#1610


jhthorsen (approved):
I think this looks like a good change. Appreciate the unit test 👍

Grinnz (approved):

PR #1613, opened 2020-11-17
Fix typo in docs example code

zoffixznet:
Unsure if it's the right way to fix, but it looks wrong when it says "if not error or error *has* a code", because I think, say, a 404 would have a 404 code and a message and needs to be rejected.


Grinnz:
No, all HTTP responses have a code; this is catching the case where a client error has occured and so there is no response.

Grinnz:
This example isn't trying to handle HTTP errors.

Issue #1607, opened 2020-11-16
Preload controller classes


kraih:
For efficiency and to make it easier to detect syntax errors it would make sense for the router to preload as many controler classes as possible from the known namespaces during startup.

kraih:
It appears we screwed up and this is not actually possible. Because the app class also happens to be a [controller namespace by default](https://github.com/mojolicious/mojo/blob/a1f627fd5d4214d8561256a8a9145d2098c8b76d/lib/Mojolicious.pm#L159). That means we don't have clear boundaries defining what modules contain controllers, and would have to preload absolutely everything.

kraih:
I had the feature almost implemented before i noticed. 😭 https://github.com/mojolicious/mojo/compare/controller_preloading

PR #1614, opened 2020-11-28
Add framework for preloading controllers

kraih:
### Summary
This is a less ambitious solution for #1607. It doesn't preload all possible controllers but leaves it up to the developer to configure the right namespaces. The default of `MyApp::Controller` might just work for most of our users though.

### Motivation
More efficient memory usage. Detecting syntax errors in controllers much earlier (during application startup instead of at runtime).

### References
#1607

### DIFF
--- a/lib/Mojo/Loader.pm
+++ b/lib/Mojo/Loader.pm
@@ -6,7 +6,7 @@ use Mojo::Exception;
 use Mojo::File qw(path);
 use Mojo::Util qw(b64_decode class_to_path);
 
-our @EXPORT_OK = qw(data_section file_is_binary find_modules find_packages load_class);
+our @EXPORT_OK = qw(data_section file_is_binary find_modules find_packages load_class load_classes);
 
 my (%BIN, %CACHE);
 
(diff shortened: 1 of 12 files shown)


amenonsen:
For what it's worth, I read through this commit, and it looks good to me.

lib/Mojo/Loader.pm
```diff
@@ -15,12 +15,17 @@ sub data_section { $_[0] ? $_[1] ? _all($_[0])->{$_[1]} : _all($_[0]) : undef }
 sub file_is_binary { keys %{_all($_[0])} ? !!$BIN{$_[0]}{$_[1]} : undef }
 
 sub find_modules {
-  my $ns = shift;
+  my ($ns, $options) = (shift, shift // {});
+
+  my @ns  = split '::', $ns;
```
Grinnz, line 20: `split /::/, $ns;`

Grinnz:
I like this approach especially as it can be easily configured to avoid preloading if someone was so inclined (e.g. CGI deployment).

christopherraa (approved):
:+1: 

No objections from me. Surprised that this new functionality wasn't a bigger change.

christopherraa:
One question as an after-thought: why default to `MyApp::Controller`? Does anyone really use that? If not, would it make more sense to default to whatever namespace the app is in, in the case of a full app? So if anyone has generated a `Corp::App::Foo` application, defaulting to preloading `Corp::App::Foo`. Too difficult or error-prone to detect perhaps?

lib/Mojo/Loader.pm
kraih: We actually use the string form everywhere in Mojolicious.

kraih:
When i say `MyApp::Controller` i of course mean `"@{[ref $self]}::Controller"` in code.

lib/Mojo/Loader.pm
Grinnz: Weird, I'll suggest changing that separately...

Grinnz (approved):

marcusramberg (commented):
I think this is a great addition to mojo. 👍

PR #1615, opened 2020-11-28
Use split /PATTERN/ syntax consistently

Grinnz:
### Summary
Use `split /PATTERN/` rather than `split 'STRING'` where appropriate.

### Motivation
As described in its [documentation](https://perldoc.perl.org/functions/split), the first argument to split is a pattern, not a string. Even if a string is passed, it is used as a pattern. Thus using the `/PATTERN/` syntax makes it clear to the reader that it is a regex and not a string.

Note that `split ' '` (single space string) is an exception where a string must be passed to activate awk compatibility, so we leave this case alone. Also, Mojo::ByteStream's split method cannot take a bare `/PATTERN/` because in that case it would be interpreted as an immediate regex match against `$_`.

### References
https://perldoc.perl.org/functions/split

### DIFF
--- a/lib/Mojo/Exception.pm
+++ b/lib/Mojo/Exception.pm
@@ -65,7 +65,7 @@ sub inspect {
   }
 
   # Search for context in sources
-  $self->_context($files[-1][1], [map { [split "\n"] } @sources]) if @sources;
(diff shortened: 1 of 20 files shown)


kraih (approved):

jhthorsen (approved):
Looks good to me 👍

PR #1616, opened 2020-11-28
Treat type_start in Mojolicious::Routes::Pattern as a character instead of a regex

Grinnz:
### Summary
The type_start character passed to split is currently being treated as a regex but it should be treated as a string to split on.

### Motivation
The type_start attribute is meant to be a single character that delimits a placeholder name from its type. It was being
passed directly to split and thus used as a regex, which means any regex metacharacters would break the parsing.
Interpolate it with \Q to apply quotemeta to the character for use in split, and add tests for custom placeholder
characters.

### DIFF
--- a/t/mojolicious/pattern.t
+++ b/t/mojolicious/pattern.t
@@ -250,4 +250,23 @@ is_deeply $result, {'bar' => 23}, 'right structure';
 is $pattern->render($result, 1), '/foo/23/baz', 'right result';
 ok !$pattern->match('/foo/bar/baz', 1), 'no result';
 
+subtest 'Alternate placeholder characters' => sub {
+  my $pattern = Mojolicious::Routes::Pattern->new->placeholder_start('|')->parse('/|test');
+  my $result  = $pattern->match('/foo');
+  is_deeply $result, {test => 'foo'}, 'right structure';
+  is $pattern->render($result, 1), '/foo', 'right result';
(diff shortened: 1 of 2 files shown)


t/mojolicious/pattern.t
```diff
@@ -250,4 +250,22 @@ is_deeply $result, {'bar' => 23}, 'right structure';
...
+is_deeply $result, {test => 'foo'}, 'right structure';
+is $pattern->render($result, 1), '/foo', 'right result';
+$pattern = Mojolicious::Routes::Pattern->new->wildcard_start('|')->parse('/|test');
+$result  = $pattern->match('/foo');
+is_deeply $result, {test => 'foo'}, 'right structure';
+is $pattern->render($result, 1), '/foo', 'right result';
+$pattern = Mojolicious::Routes::Pattern->new->types({num => qr/\d+/})->type_start('|')->parse('/<test|num>');
+$result  = $pattern->match('/42');
+is_deeply $result, {test => '42'}, 'right structure';
+is $pattern->render($result, 1), '/42', 'right result';
```
kraih, line 269: Make it a subtest (we do that for all new tests).

Grinnz: Done

kraih (approved):

marcusramberg (approved):
lgtm.

PR #1617, opened 2020-11-28
Document and test that Mojo::ByteStream split method uses core split function semantics

Grinnz:
### Summary
The first argument to Mojo::ByteStream's split method has some special treatment as it's passed directly to the split function. Add some documentation and examples to clarify this treatment - it's the pattern argument and is treated as a regex, except in the awk compatibility special case of a single space character, and the special case of an empty pattern which splits on each character.

It also may be worthwhile to suggest and document passing split patterns to this function as `qr//` regex refs rather than strings, thoughts? (this is already tested)

### Motivation
Clarify how this argument is interpreted.

### References
https://perldoc.perl.org/functions/split

### DIFF
--- a/lib/Mojo/ByteStream.pm
+++ b/lib/Mojo/ByteStream.pm
@@ -49,6 +49,9 @@ sub size { length ${$_[0]} }
 
 sub split {
   my ($self, $pat, $lim) = (shift, shift, shift // 0);
+
+  # awk compatible special case on 5.16
+  return Mojo::Collection->new(map { $self->new($_) } split ' ',  $$self, $lim) if !ref $pat and $pat eq ' ';
   return Mojo::Collection->new(map { $self->new($_) } split $pat, $$self, $lim);
 }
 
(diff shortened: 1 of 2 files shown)


kraih (approved):

Grinnz:
The split function only consistently interprets a string containing a single space when it's not a literal since 5.18: https://perldoc.perl.org/perl5180delta#split's-first-argument-is-more-consistently-interpreted

So added a workaround to make this work consistently on 5.16, the other option is not documenting or testing the single space string case at all.

lib/Mojo/ByteStream.pm
```diff
@@ -49,6 +49,9 @@ sub size { length ${$_[0]} }
 
 sub split {
   my ($self, $pat, $lim) = (shift, shift, shift // 0);
+
+  # awk compatible special case on 5.16
```
kraih, line 53: AWK.

kraih (changes requested):
I'm opposed to adding special workarounds for Perl 5.16.

kraih:
For reference. https://docs.mojolicious.org/Mojolicious/Guides/FAQ#Which-versions-of-Perl-are-supported-by-Mojolicious

kraih:
A week should have been enough time for anyone who disagrees to speak up, so i'll close this PR as rejected.

PR #1619, opened 2020-12-01
Randomize request id morbo

DabeDotCom:
### Summary
Running under Morbo, any time it detects a file has changed and reloads itself, it resets `$c->req->request_id` back to its *initial* value, rather than continuing to generate new ones — i.e., it ***reuses*** old Request IDs, which makes it hard to `grep` through logs.  :-\

### Motivation
I tried to limit the blast radius of my changes to where they're actually used — just the `Mojo::Message::Request` package.

I had originally thought of promoting the SEED/COUNTER to `Mojo::EventEmitter`, and then maybe adding some kind of `app->srand()` method to re-seed the RNG, which would get called by `Mojolicious->new()` right before running `startup()`, but I figured keeping it localized, like before, was more prudent...

(PS: I'm also guessing the fact that it uses/increments lexically scoped `my` variables was an optimization to only call `rand()` once [okay, twice] up-front, rather than on every request — even though it still calls `sha1_base64` every time, which is ~25–30x slower on my machine...?  But maybe that's not necessarily true if you've replaced the RNG with something more cryptographically secure?  «shrug»)

### References
If you'd prefer that I create an issue, and then reference *that*, by all means, lemme know...

### DIFF
--- a/lib/Mojo/Message/Request.pm
+++ b/lib/Mojo/Message/Request.pm
@@ -6,13 +6,14 @@ use Mojo::Cookie::Request;
 use Mojo::Util qw(b64_encode b64_decode sha1_sum);
 use Mojo::URL;
 
-my ($SEED, $COUNTER) = ($$ . time . rand, int rand 0xffffff);
-
 has env    => sub { {} };
 has method => 'GET';
 has [qw(proxy reverse_proxy)];
 has request_id => sub {
-  my $b64 = substr(sha1_base64($SEED . ($COUNTER = ($COUNTER + 1) % 0xffffff)), 0, 8);
+  state $seed    = $$ . time . rand;
+  state $counter = int rand 0xffffff;
+
+  my $b64 = substr(sha1_base64($seed . ($counter = ($counter + 1) % 0xffffff)), 0, 8);
   $b64 =~ tr!+/!-_!;
   return $b64;
 };
--- a/t/mojo/morbo.t
+++ b/t/mojo/morbo.t
@@ -117,6 +117,7 @@ EOF
   ok((stat $script)[9] == $mtime, 'modify time has not changed');
   isnt((stat $script)[7], $size, 'size has changed');
   sleep 0.1 until -f $started;
+  $started = $started->sibling('started4.txt');
 
   # Application has been reloaded again
   my $tx = $ua->get("http://127.0.0.1:$port/hello");
@@ -151,6 +152,53 @@ subtest 'Broken symlink' => sub {
   local $SIG{__WARN__} = sub { $warned++ };
   is_deeply $morbo->backend->modified_files, [], 'directory has not changed';
   ok !$warned, 'no warnings';
+};
+
+subtest 'Randomized Request IDs' => sub {
+  my @ids;
+  $script->spurt(<<"EOF");
+use Mojolicious::Lite;
+use Mojo::File qw(path);
+use Mojo::IOLoop;
+
+app->log->level('fatal');
+
+Mojo::IOLoop->next_tick(sub { path('$started')->touch });
+
+# Return "request_id"
+get '/request_id' => sub {
+  my \$c = shift;
+  \$c->render( text => \$c->req->request_id );
+};
+
+app->start;
+EOF
+  sleep 0.1 until -f $started;
+
+  # Fetch initial request_ids
+  my $tx = $ua->get("http://127.0.0.1:$port/request_id");
+  like $tx->res->body, qr/^[\-_0-9a-z]+$/i, 'request_id is valid';
+  $ids[0] = $tx->res->body;
+
+  $tx = $ua->get("http://127.0.0.1:$port/request_id");
+  like $tx->res->body, qr/^[\-_0-9a-z]+$/i, 'request_id is valid';
+  $ids[1] = $tx->res->body;
+
+  isnt $ids[1], $ids[0], 'request_ids differ';
+
+  # Restart Morbo and fetch new request_ids
+  $started->remove;
+  $script->touch;
+  sleep 0.1 until -f $started;
+  $started = $started->sibling('started5.txt');
+
+  $tx = $ua->get("http://127.0.0.1:$port/request_id");
+  like $tx->res->body, qr/^[\-_0-9a-z]+$/i, 'request_id is valid';
+  isnt $tx->res->body, $ids[0], 'request_ids differ';
+
+  $tx = $ua->get("http://127.0.0.1:$port/request_id");
+  like $tx->res->body, qr/^[\-_0-9a-z]+$/i, 'request_id is valid';
+  isnt $tx->res->body, $ids[1], 'request_ids differ';
 };
 
 # Stop


shadowcat-mst:
What about moving $SEED into the 'has request_id' default closure as a state variable? Seems noticeably simpler

Grinnz:
Also much more efficient, and nothing else in mojolicious uses subroutines this way.

DabeDotCom:
> What about moving $SEED into the 'has request_id' default closure as a state variable? Seems noticeably simpler

Mayhaps that code was originally written to target pre-state Perls?  But hey, now that the minimum requirement is 5.16.0 (soon to be 5.20.0) then sure!  :-D

t/mojo/morbo.t
```diff
@@ -136,6 +136,49 @@ subtest 'Broken symlink' => sub {
...
+  my $c = shift;
+  $c->render( text => $c->req->request_id );
+};
+
+app->start;
+EOF
+
+  # Fetch initial request_ids
+  $script->spurt($src);
+  sleep 3;
```
Grinnz, line 157: 3 second sleeps in tests is not acceptable.

Grinnz, line 157: Nevermind, I realized this is a developer test.

Changes
```diff
@@ -1,6 +1,8 @@
 
 8.67  2020-11-30
   - Improved Mojolicious::Commands to throw an exception for invalid commands.
+  - Fixed a bug where request_id attribute in Mojo::Message::Request would revert to previous value(s) whenever Morbo
+    restarted -- e.g., responding to file modification.  (DabeDotCom)
```
kraih, line 5: Please don't add Changes entries in PRs.

lib/Mojo/Message/Request.pm
```diff
@@ -6,13 +6,14 @@ use Mojo::Cookie::Request;
...
 
-my ($SEED, $COUNTER) = ($$ . time . rand, int rand 0xffffff);
-
 has env    => sub { {} };
 has method => 'GET';
 has [qw(proxy reverse_proxy)];
 has request_id => sub {
-  my $b64 = substr(sha1_base64($SEED . ($COUNTER = ($COUNTER + 1) % 0xffffff)), 0, 8);
+  state $seed    = $$ . time . rand;
+  state $counter = int rand 0xffffff;
```
kraih, line 14: Can this be one line?

t/mojo/morbo.t
kraih: I still run those tests and i hate waiting...

t/mojo/morbo.t
```diff
@@ -136,6 +136,49 @@ subtest 'Broken symlink' => sub {
...
+
+  $tx = $ua->get("http://127.0.0.1:$port/request_id");
```
kraih, line 159: You're using a new scope, why is this a global?

lib/Mojo/Message/Request.pm
Grinnz: Sadly no, state declarations don't support list assignment.

Grinnz: Although I guess it could be written `state $seed = ..., state $counter = ...;` but that's kind of weird...

t/mojo/morbo.t
```diff
@@ -136,6 +136,49 @@ subtest 'Broken symlink' => sub {
...
+  like $tx->res->body, qr/^[\-_0-9a-z]+$/i, 'okay content';
```
kraih, line 160: That's an inconsistent generic test description.

t/mojo/morbo.t
```diff
@@ -136,6 +136,49 @@ subtest 'Broken symlink' => sub {
...
+  $ids[0] = $tx->res->body;
+
+  $tx = $ua->get("http://127.0.0.1:$port/request_id");
+  like $tx->res->body, qr/^[\-_0-9a-z]+$/i, 'okay content';
+  $ids[1] = $tx->res->body;
+
+  isnt $ids[1], $ids[0], "different (initial) request_ids: \"$ids[0]\" != \"$ids[1]\"";
```
kraih, line 167: Including the ids in the test description adds no value and makes things less readable.

kraih:
Squash your commits.

kraih:
Without the test i would have been inclined to approve, but making the test so much slower is a bit of a deal breaker for me.

DabeDotCom:
Thanks, everybody, for all your help!  I'll give a blanket *mea culpa* that I wasn't sure if you all preferred `push -f` or the full commit history.

(Just like I'm not sure if you prefer one long reply or a flurry of "Fixed!" one-liners...  I originally wrote one single comment, addressing everything at once, but I'm about to copy-paste the salient bits into their corresponding conversations, above.  Apologies in advance if I guessed wrong — *again!*  «grin»)

Changes
DabeDotCom: *Mea culpa*  I've backed it out.

lib/Mojo/Message/Request.pm
DabeDotCom: Yeah, I tried to do `state ($seed, $counter) = ...` — and even `(state $seed, state $counter) = ...` — but no dice.  Quoth `perldiag`, it's a big ol' (F)atal error:

```
    Initialization of state variables in list currently forbidden
        (F) "state" only permits initializing a single variable, specified
        without parentheses. So "state $a = 42" and "state @a = qw(a b c)"
        are allowed, but not "state ($a) = 42" or "(state $a) = 42". To
        initialize more than one "state" variable, initialize them one at a
        time.
```

And I agree that `state $seed = ..., state $counter = ...` is awkward; you might as well just use a semicolon at that point, hehe!

t/mojo/morbo.t
DabeDotCom: ~~Cargo-culting~~ — er, I mean, *"Standing on the Shoulders of Giants"* — is why I reused the global `$tx` and `$ua`, etc.  I was only using `subtest` as a means of *grouping* the tests, not necessarily because of the new lexical scope.

Would you rather I just leave them at the top level?  (Since going through all that effort again in a subtest seems wasteful... ***That's*** not the part of the system that's actually being tested.  «shrug»)

t/mojo/morbo.t
DabeDotCom: *"Fixed in the Next Release"*

(With `request_id is valid` and 'request_ids differ', respectively)

t/mojo/morbo.t
DabeDotCom: FWIW, I just copied the `sleep 3` from the earlier tests in that file...  If there's an easier way to reliably block until Morbo reloads/restarts, I am absolutely, without a doubt, 100% all for it!

Like, if there were some kind of `*after*_server_start` hook that I could `await` on, or something?  (Does that even make sense?  Create a promise, start/restart the server, await, and then `resolve` it in a hook/handler?) I'm sure the devil is in the details; i.e., making it work for all the various `Mojo::Server::*` implementations surely involves voodoo that I can't even begin to fathom — otherwise I'm sure somebody would have knocked it by now!

† And isn't it great how, what seems like a lifetime ago, I remember `LOAD 1,1`-ing Commodore 64 games from a Cassette Tape in **less than \*10\* minutes!**  ;-D

kraih: I've changed the existing tests in the meantime to make them faster.

t/mojo/morbo.t
kraih: The test has been adjusted in the meantime and should answer all questions. 😁

t/mojo/morbo.t
kraih: If they don't match they will be shown in the output anyway.

t/mojo/morbo.t
DabeDotCom: Ahh!  I see what you did there, thanks!!  :-D

BTW: Not to blaspheme, or anything, but I will say that it felt wrong that the "previous" test was responsible for updating the temporary file, and having to manually remember to increment the number seems like a footgun just waiting to go off...

For instance, to maintain consistency, I had to diddle around up in the *'Update script without changing mtime'* subtest — which wasn't "mine" — adding `$started = $started->sibling('started4.txt')`.

[With faux indignation:]  "And you yelled at ***\*ME\**** for global variables?!?"  (I'm so totally «j/k», BTW!!)  ☮️

Anyway, I've tweaked it to do:

```
my $dir = tempdir;    # This is still global...

subtest 'Blah Blah Blah' => sub {
  my $started = $dir->child('started.txt');
  ok !-f $started, 'started.txt does not exist';

  # $script->spurt(...), etc.
  sleep 0.1 until -f $started;
  ok $started->remove, 'started.txt removed';

  # my $tx = $ua->get(...)
};
```

... which seems to work.  (Would that be a separate PR?  Or a second commit in this one?  Sorry for sounding like such a newb, but *squashing* it into the current commit seems like it violates the "Do One Thing and Do It Well" Single-Responsibility Principle.  But maybe that's okay, in this case?  «shrug»)

Anyway, I'm sure one could ~~waste~~ I mean, "spend" a ton of time making it more robust — e.g., generating random filenames (`File::Temp::tempnam()` ?), and using `mkdir` instead of `touch` 'cuz the former is atomic, and all that — but I'm pretty certain that's beginning to cross into the realm of *"Better Is The Enemy Of Good"* and *"You Ain't Gonna Need It"*, etc.  :-D

Thanks again!

t/mojo/morbo.t
DabeDotCom: I think I've been burned by `ok $one eq $two` (rather than `is $one, $two`) too many times...  :-}

That, and by having tests bail out before they actually print the line number, whereas a unique and descriptive test name makes it at least *possible* to grep for what died.  «shrug»

t/mojo/morbo.t
```diff
@@ -117,6 +117,7 @@ EOF
   ok((stat $script)[9] == $mtime, 'modify time has not changed');
   isnt((stat $script)[7], $size, 'size has changed');
   sleep 0.1 until -f $started;
+  $started = $started->sibling('started4.txt');
```
kraih, line 120: There is no reason for this to be so far from the place where it is actually used.

t/mojo/morbo.t
```diff
@@ -153,6 +154,53 @@ subtest 'Broken symlink' => sub {
   ok !$warned, 'no warnings';
 };
 
+subtest 'Randomized Request IDs' => sub {
+  my @ids;
```
kraih, line 158: The array is not used in this block at all, but much further down.

t/mojo/morbo.t
```diff
@@ -153,6 +154,53 @@ subtest 'Broken symlink' => sub {
...
+get '/request_id' => sub {
+  my \$c = shift;
+  \$c->render( text => \$c->req->request_id );
+};
+
+app->start;
+EOF
+  sleep 0.1 until -f $started;
+
+  # Fetch initial request_ids
```
kraih, line 178: `request_ids` is not a word.

t/mojo/morbo.t
```diff
@@ -153,6 +154,53 @@ subtest 'Broken symlink' => sub {
...
+  my $tx = $ua->get("http://127.0.0.1:$port/request_id");
+  like $tx->res->body, qr/^[\-_0-9a-z]+$/i, 'request_id is valid';
+  $ids[0] = $tx->res->body;
```
kraih, line 181: Why is this not just a variable with a reasonably descriptive name?

t/mojo/morbo.t
```diff
@@ -153,6 +154,53 @@ subtest 'Broken symlink' => sub {
...
+  \$c->render( text => \$c->req->request_id );
+};
+
+app->start;
+EOF
+  sleep 0.1 until -f $started;
+
+  # Fetch initial request_ids
+  my $tx = $ua->get("http://127.0.0.1:$port/request_id");
+  like $tx->res->body, qr/^[\-_0-9a-z]+$/i, 'request_id is valid';
```
kraih, line 180: This does not seem the right place to test the actual format of the request id, which we change every now and then. Only makes that harder in the future.

kraih:
Fyi. Unless you tell us otherwise it will be assumed that this is a work in progress, and if you're unlucky nobody will do a re-review.

kraih:
I'm afraid i couldn't wait any longer for this PR, the problem has a wider scope than assumed and affects `Mojo::Server::Prefork` workers too. Which is a bigger deal, as it affects production systems. https://github.com/mojolicious/mojo/compare/907dd0ec6129...c54ca5254a96

Issue #1620, opened 2020-12-01
Exceeding the max_request_size has no consequences 


pileswasp:
* Mojolicious version: 8.63 and 8.66
* Perl version: v5.30.0
* Operating system: Ubuntu Server 20.04 LTS (Focal Fossa)

### Steps to reproduce the behavior
```perl
#!/usr/bin/env perl
use Mojolicious::Lite -signatures;
app->max_request_size(100);
post '/' => sub ($c) {
    $c->log->warn( 'Still processing....' );
  $c->render(text => 'Still processing');
};
app->start;
```
$ `./request_limit get -v -mode development --method post --form long=111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111 /`
[2020-12-01 21:00:34.68342] [39407] [debug] **Maximum message size exceeded**
[2020-12-01 21:00:34.68556] [39407] [debug] [30DScmhj] POST "/"
[2020-12-01 21:00:34.68640] [39407] [debug] [30DScmhj] Routing to a callback
[2020-12-01 21:00:34.68678] [39407] [warn] [30DScmhj] **Still processing....**
[2020-12-01 21:00:34.68761] [39407] [debug] [30DScmhj] 200 OK (0.002009s, 497.760/s)
POST / HTTP/1.1
Accept-Encoding: gzip
Host: 127.0.0.1:36349
User-Agent: Mojolicious (Perl)
Content-Type: application/x-www-form-urlencoded
Content-Length: 107

HTTP/1.1 200 OK
Server: Mojolicious (Perl)
Content-Length: 16
Content-Type: text/html;charset=UTF-8
Connection: close
Date: Tue, 01 Dec 2020 21:00:34 GMT

Still processing

### Expected behavior
Explosions and fire.
The [main docs](https://docs.mojolicious.org/Mojolicious#max_request_size) don't actually say what's supposed to happen.
The FAQ suggests it should ["automatically stop accepting new data, and in most cases force the connection to be closed"](https://docs.mojolicious.org/Mojolicious/Guides/FAQ#What-does-the-error-Maximum-message-size-exceeded-mean).
Something along the lines of the FAQ and a non-200 HTTP response is what I expected would happen.

### Actual behavior
All I get is a debug-level message in the log (so not even that in production mode) and the code runs as though nothing is amiss.

[...]

PR #1623, opened 2020-12-05
Additional docs to clarify use of max_request_size

pileswasp:
### Summary, Motivation & References
https://github.com/mojolicious/mojo/issues/1620

### DIFF
--- a/lib/Mojo/Message.pm
+++ b/lib/Mojo/Message.pm
@@ -553,7 +553,7 @@ Check if message parser/generator is finished.
   my $bool = $msg->is_limit_exceeded;
 
 Check if message has exceeded L</"max_line_size">, L</"max_message_size">, L<Mojo::Content/"max_buffer_size"> or
-L<Mojo::Headers/"max_line_size">.
+L<Mojo::Headers/"max_line_size">. L</"max_message_size"> may be set by L<Mojolicious/"max_request_size">.
 
 =head2 json
 
--- a/lib/Mojolicious.pm
+++ b/lib/Mojolicious.pm
@@ -425,6 +425,9 @@ will allow requests of indefinite size. Note that increasing this value can also
 should you for example attempt to parse an excessively large request body with the methods L<Mojo::Message/"dom"> or
 L<Mojo::Message/"json">.
 
+If a request exceeds this limit it may be truncated, but your action will still process. Your action should test for
+requests which exceed this limit with L<Mojo::Message/"is_limit_exceeded">.
+
 =head2 mode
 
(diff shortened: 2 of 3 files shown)


lib/Mojo/Message.pm
```diff
@@ -553,7 +553,7 @@ Check if message parser/generator is finished.
   my $bool = $msg->is_limit_exceeded;
 
 Check if message has exceeded L</"max_line_size">, L</"max_message_size">, L<Mojo::Content/"max_buffer_size"> or
-L<Mojo::Headers/"max_line_size">.
+L<Mojo::Headers/"max_line_size">. L</"max_message_size"> may be set by L<Mojolicious/"max_request_size">.
```
kraih, line 556: That's a very random reference that doesn't really add value.

lib/Mojolicious.pm
```diff
@@ -425,6 +425,9 @@ will allow requests of indefinite size. Note that increasing this value can also
 should you for example attempt to parse an excessively large request body with the methods L<Mojo::Message/"dom"> or
 L<Mojo::Message/"json">.
 
+If a request exceeds this limit it may be truncated, but your action will still process. Your action should test for
+requests which exceed this limit with L<Mojo::Message/"is_limit_exceeded">.
+
```
kraih, line 430: That's a guide style description, this seems like the wrong place for it.

lib/Mojo/Message.pm
pileswasp: My intent was to tie max_request_size and is_limit_exceeded together, such that someone googling "mojolicious max_request_size" has a good chance of figuring out that they have to manually check (rather than assuming that setting max_request_size is fire and forget), but it could probably work just as well in reverse by explaining at max_request_size that it sets max_message_size.

lib/Mojolicious.pm
pileswasp: Suggestions for the right place welcome.

pileswasp:
Rebased to latest master.

lib/Mojolicious.pm
kraih: Additionally we will likely change how this is handled in the future. Parser errors ideally shouldn't reach the router in the first place. So new documentation is probably counter productive in that regard.

pileswasp: Heh! I got here via the faulty assumption that exceeding the max_request_size would just give a 413 and no service like Apache's LimitRequestBody. Cancelling this request for reasons given here: https://github.com/mojolicious/mojo/issues/1620

PR #1624, opened 2020-12-08
Document how to change the HTTP status code for an exception

### LINKED ISSUE #1622: Document how to add optional HTTP code code to exception
* Mojolicious version: 8.66
* Perl version: N/A
* Operating system: N/A

It's desirable to throw an exception in code for bad input and return a custom error code, e.g. HTTP 400 for bad parameters.

This also helps to identify the error when looking at HTTP logs, without having to look at an application log (which can be noisier).

### Expected behavior

It would be helpful to use something like
```perl
return $c->reply->exception('Division by zero!', 400)
   if $divisor == 0;
```

### Actual behavior

The only option is to return HTTP 404 (Not Found) or HTTP 500 (Server Error).

robrwo:
### Summary

Explicitly document how to change the HTTP return code for exceptions

### Motivation

It's not as obvious to newcomers.

### References

This closes #1622.

### DIFF
--- a/lib/Mojolicious/Guides/Rendering.pod
+++ b/lib/Mojolicious/Guides/Rendering.pod
@@ -425,6 +425,10 @@ L<Mojolicious::Plugin::DefaultHelpers/"reply-E<gt>not_found">.
 
   app->start;
 
+To change the HTTP status code of the exception, you can use L<Mojolicious::Controller/"rendered">.
+
+  return $c->reply->exception('Division by zero!')->rendered(400) if $divisor == 0;
+
 You can also change the templates of those pages, since you most likely want to show your users something more closely
 related to your application in production. The renderer will always try to find C<exception.$mode.$format.*> or
 C<not_found.$mode.$format.*> before falling back to the built-in default templates.


jhthorsen (dismissed):
I'm neutral to this change, but wanted to mention that the indentation is off.

lib/Mojolicious/Guides/Rendering.pod
```diff
@@ -425,6 +425,11 @@ L<Mojolicious::Plugin::DefaultHelpers/"reply-E<gt>not_found">.
 
   app->start;
 
+You can change the HTTP return code in an exception
```
kraih, line 428: That's not a complete sentence. Also it's not a "return code".

lib/Mojolicious/Guides/Rendering.pod
```diff
@@ -425,6 +425,11 @@ L<Mojolicious::Plugin::DefaultHelpers/"reply-E<gt>not_found">.
 
   app->start;
 
+You can change the HTTP status code of the exception:
+
+  return $c->reply->exception('Division by zero!')->rendered(400)
+    if $divisor == 0;
```
kraih, line 431: The line break is unnecessary.

lib/Mojolicious/Guides/Rendering.pod
```diff
@@ -425,6 +425,10 @@ L<Mojolicious::Plugin::DefaultHelpers/"reply-E<gt>not_found">.
 
   app->start;
 
+You can change the HTTP status code of the exception:
```
kraih, line 428: The next sentence also start with `You can ...`, that doesn't look nice. How about:
```
To change the HTTP status code of the exception you can use L<Mojolicious::Controller/"rendered">.
```

robrwo:
Is this alright?

kraih (approved):

PR #1633, opened 2020-12-22
Fixed a bug in the synopsis section of the documentation

Dudley5000:
### Summary
Updated the very first example in the synopsis section of the Mojolicious::Lite perldoc.

### Motivation
It can be discouraging for newcomers to Mojolicious who are eager to try it out only to discover that their Hello World script throws an error **immediately**.

### References
Since the `-signatures` flag is not present in the first example, anyone who tries to copy/paste and run that code snippet will encounter an error.


kraih:
Thanks, that example should have used `-signatures`, like all the others do.

Dudley5000:
Glad to help. Happy holidays.

PR #1635, opened 2020-12-22
Add workflow that triggers the website to rebuild

jberger:
Now that as of https://github.com/mojolicious/mojolicious.org/pull/23 the website can be automatically built and deployed, this PR adds the ability to trigger the website rebuild when pushes are made to the master branch of mojolicious itself.

### DIFF
--- a/.github/workflows/rebuild-website.yml
+++ b/.github/workflows/rebuild-website.yml
@@ -1,0 +1,18 @@
+name: Rebuild Website
+on:
+  push:
+    branches:
+      - master
+jobs:
+  rebuild_website:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Trigger website workflow
+        run: |
+          curl \
+            -X POST \
+            -u "${{ secrets.WORKFLOW_DISPATCH_USERINFO }}" \
+            -H "Accept: application/vnd.github.everest-preview+json" \
+            -H "Content-Type: application/json" \
+            --data '{"ref": "master"}' \
+            https://api.github.com/repos/mojolicious/mojolicious.org/actions/workflows/docker-publish.yml/dispatches


jhthorsen (approved):
LGTM 👍

.github/workflows/rebuild-website.yml
```diff
@@ -0,0 +1,20 @@
+name: Docker
```
kraih, line 1: I don't get the name.

.github/workflows/rebuild-website.yml
```diff
@@ -0,0 +1,20 @@
...
+
+on:
+  push:
+    branches:
+      - master
+
```
kraih, line 7: Our other workflows do not use empty lines as separators.

.github/workflows/rebuild-website.yml
```diff
@@ -0,0 +1,20 @@
...
+jobs:
+  rebuild_website:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Trigger website workflow
+        run: |
+          curl \
+            -XPOST \
```
kraih, line 15: `-X POST` would be easier to read.

.github/workflows/rebuild-website.yml
jberger: I'll admit, I didn't think about this much from a style perspective. The one in the website probably needs a cleanup in that way too then.

kraih: Yes, it does.

.github/workflows/rebuild-website.yml
```diff
@@ -0,0 +1,18 @@
+name: Rebuild Website
```
kraih, line 1: All the examples seem to suggest names like `rebuild-website`.

kraih (approved):

PR #1644, opened 2020-12-30
Trusted proxy handling to existing reverse proxy functionality

jberger:
Both when using self-hosted kubernetes with ingresses and using external cloud proxy services, it is useful to have the ability mark certain proxies as trusted, ignoring their entries in the X-Forwarded-For header and getting the true client's ip address. While there are plugins on CPAN that can do this, those are hard to add to existing pre-built applications which would have to load them, and their mechanism to do so (hooks and rewriting) is heavier that what can be done in core (during message parsing).

This PR adds this ability to core, keeping the implementation as simple as possible and using only core libraries.

### DIFF
--- a/lib/Mojo/Message/Request.pm
+++ b/lib/Mojo/Message/Request.pm
@@ -16,8 +16,9 @@ has request_id => sub {
   $b64 =~ tr!+/!-_!;
   return $b64;
 };
-has url       => sub { Mojo::URL->new };
-has via_proxy => 1;
+has trusted_proxies => sub { [] };
+has url             => sub { Mojo::URL->new };
+has via_proxy       => 1;
 
 sub clone {
   my $self = shift;
@@ -300,6 +301,13 @@ Proxy URL for request.
   $req     = $req->reverse_proxy($bool);
 
 Request has been performed through a reverse proxy.
+
+=head2 trusted_proxies
+
+  my $proxies = $req->trusted_proxies;
+  $req        = $req->trusted_proxies(['10.0/8', '127.0.0.1', '172.16.0/12', '192.168.0/16', 'fc00::/7']);
+
+Trusted reverse proxies, addresses or networks in CIDR form.
 
 =head2 request_id
 
--- a/lib/Mojo/Server.pm
+++ b/lib/Mojo/Server.pm
@@ -8,8 +8,9 @@ use Mojo::Util qw(md5_sum);
 use POSIX ();
 use Scalar::Util qw(blessed);
 
-has app           => sub { shift->build_app('Mojo::HelloWorld') };
-has reverse_proxy => sub { $ENV{MOJO_REVERSE_PROXY} };
+has app             => sub { shift->build_app('Mojo::HelloWorld') };
+has reverse_proxy   => sub { $ENV{MOJO_REVERSE_PROXY} || !!@{shift->trusted_proxies} };
+has trusted_proxies => sub { [split /\s*,\s*/, ($ENV{MOJO_TRUSTED_PROXIES} // '')] };
 
 sub build_app {
   my ($self, $app) = (shift, shift);
@@ -21,6 +22,7 @@ sub build_app {
 sub build_tx {
   my $self = shift;
   my $tx   = $self->app->build_tx;
+  push @{$tx->req->trusted_proxies}, @{$self->trusted_proxies};
   $tx->req->reverse_proxy(1) if $self->reverse_proxy;
   return $tx;
 }
@@ -130,7 +132,17 @@ Application this server handles, defaults to a L<Mojo::HelloWorld> object.
   my $bool = $server->reverse_proxy;
   $server  = $server->reverse_proxy($bool);
 
-This server operates behind a reverse proxy, defaults to the value of the C<MOJO_REVERSE_PROXY> environment variable.
+This server operates behind a reverse proxy, defaults to the value of the C<MOJO_REVERSE_PROXY> environment variable
+or true if L</trusted_proxies> is not empty.
+
+=head2 trusted_proxies
+
+  my $proxies = $server->trusted_proxies;
+  $server     = $server->trusted_proxies(['10.0/8', '127.0.0.1', '172.16.0/12', '192.168.0/16', 'fc00::/7']);
+
+This server expects requests from trusted reverse proxies, defaults to the value of the C<MOJO_TRUSTED_PROXIES>
+environment variable split on commas with optional whitespace. These proxies should be addresses or networks in CIDR
+form.
 
 =head1 METHODS
 
--- a/lib/Mojo/Transaction.pm
+++ b/lib/Mojo/Transaction.pm
@@ -2,9 +2,10 @@ package Mojo::Transaction;
 use Mojo::Base 'Mojo::EventEmitter';
 
 use Carp qw(croak);
+use List::Util qw(any);
 use Mojo::Message::Request;
 use Mojo::Message::Response;
-use Mojo::Util qw(deprecated);
+use Mojo::Util qw(deprecated network_contains);
 
 has [qw(kept_alive local_address local_port original_remote_address remote_port)];
 has req => sub { Mojo::Message::Request->new };
@@ -36,7 +37,15 @@ sub remote_address {
   return $self->original_remote_address unless $self->req->reverse_proxy;
 
   # Reverse proxy
-  return ($self->req->headers->header('X-Forwarded-For') // '') =~ /([^,\s]+)$/ ? $1 : $self->original_remote_address;
+  my @addrs   = split /\s*,\s*/, ($self->req->headers->header('X-Forwarded-For') // '');
+  my $trusted = $self->req->trusted_proxies;
+  return @addrs ? $addrs[-1] : $self->original_remote_address unless @$trusted;
+
+  push @addrs, $self->original_remote_address;
+  for my $addr (reverse @addrs) {
+    return $addr unless any { network_contains($_, $addr) } @$trusted;
+  }
+  return $addrs[0];
 }
 
 sub result {
@@ -242,8 +251,11 @@ False, this is not a L<Mojo::Transaction::WebSocket> object.
   my $address = $tx->remote_address;
   $tx         = $tx->remote_address('127.0.0.1');
 
-Same as L</"original_remote_address"> or the last value of the C<X-Forwarded-For> header if L</"req"> has been
-performed through a reverse proxy.
+Same as L</"original_remote_address"> unless L</"req"> has been performed via a L<Mojo::Message::Request/reverse_proxy>.
+If so then the last value of C<X-Forwarded-For> header is returned. Additionally if
+L<Mojo::Message::Request/trusted_proxies> are also provided then the original address must be trusted and any
+C<X-Forwarded-For> entries that are trusted are ignored, returning the last untrusted address or the first address if
+all are trusted.
 
 =head2 result
 
--- a/lib/Mojo/Util.pm
+++ b/lib/Mojo/Util.pm
@@ -15,6 +15,7 @@ use IO::Uncompress::Gunzip;
 use List::Util qw(min);
 use MIME::Base64 qw(decode_base64 encode_base64);
 use Pod::Usage qw(pod2usage);
+use Socket qw(inet_aton inet_pton AF_INET6);
 use Sub::Util qw(set_subname);
 use Symbol qw(delete_package);
 use Time::HiRes        ();
@@ -65,9 +66,9 @@ my (%ENCODING, %PATTERN);
 our @EXPORT_OK = (
   qw(b64_decode b64_encode camelize class_to_file class_to_path decamelize decode deprecated dumper encode),
   qw(extract_usage getopt gunzip gzip hmac_sha1_sum html_attr_unescape html_unescape humanize_bytes md5_bytes md5_sum),
-  qw(monkey_patch punycode_decode punycode_encode quote scope_guard secure_compare sha1_bytes sha1_sum slugify),
-  qw(split_cookie_header split_header steady_time tablify term_escape trim unindent unquote url_escape url_unescape),
-  qw(xml_escape xor_encode)
+  qw(monkey_patch network_contains punycode_decode punycode_encode quote scope_guard secure_compare sha1_bytes),
+  qw(sha1_sum slugify split_cookie_header split_header steady_time tablify term_escape trim unindent unquote),
+  qw(url_escape url_unescape xml_escape xor_encode)
 );
 
 # Aliases
@@ -181,6 +182,27 @@ sub monkey_patch {
   no strict 'refs';
   no warnings 'redefine';
   *{"${class}::$_"} = set_subname("${class}::$_", $patch{$_}) for keys %patch;
+}
+
+sub network_contains {
+  my ($cidr, $addr) = @_;
+  return undef unless length $cidr && length $addr;
+
+  # Parse inputs
+  my ($net, $mask) = split m!/!, $cidr, 2;
+  my $v6 = $net =~ /:/;
+  return undef if $v6 xor $addr =~ /:/;
+
+  # Convert addresses to binary
+  return undef unless $net  = $v6 ? inet_pton(AF_INET6, $net)  : inet_aton($net);
+  return undef unless $addr = $v6 ? inet_pton(AF_INET6, $addr) : inet_aton($addr);
+  my $length = $v6 ? 128 : 32;
+
+  # Apply mask if given
+  $addr &= pack "B$length", '1' x $mask if defined $mask;
+
+  # Compare
+  return 0 == unpack "B$length", ($net ^ $addr);
 }
 
 # Direct translation of RFC 3492
@@ -754,6 +776,23 @@ Punycode decode string as described in L<RFC 3492|https://tools.ietf.org/html/rf
 
   # "bücher"
   punycode_decode 'bcher-kva';
+
+=head2 network_contains
+
+  my $bool = network_contains $network, $address;
+
+Check that a given address is contained within a network in CIDR form. If the network is a single address, the
+addresses must be equivalent.
+
+  # True
+  network_contains('10.0.0.0/8', '10.10.10.10');
+  network_contains('10.10.10.10', '10.10.10.10');
+  network_contains('fc00::/7', 'fc::c0:ff:ee');
+
+  # False
+  network_contains('10.0.0.0/29', '10.10.10.10');
+  network_contains('10.10.10.12', '10.10.10.10');
+  network_contains('fc00::/7', '::1');
 
 =head2 punycode_encode
 
--- a/lib/Mojolicious/Command/daemon.pm
+++ b/lib/Mojolicious/Command/daemon.pm
@@ -7,7 +7,7 @@ use Mojo::Util qw(getopt);
 has description => 'Start application with HTTP and WebSocket server';
 has usage       => sub { shift->extract_usage };
 
-sub run {
+sub build_server {
   my ($self, @args) = @_;
 
   my $daemon = Mojo::Server::Daemon->new(app => $self->app);
@@ -18,12 +18,16 @@ sub run {
     'i|inactivity-timeout=i' => sub { $daemon->inactivity_timeout($_[1]) },
     'k|keep-alive-timeout=i' => sub { $daemon->keep_alive_timeout($_[1]) },
     'l|listen=s'             => \my @listen,
-    'p|proxy'                => sub { $daemon->reverse_proxy(1) },
+    'p|proxy:s'              => \my @proxy,
     'r|requests=i'           => sub { $daemon->max_requests($_[1]) };
 
   $daemon->listen(\@listen) if @listen;
-  $daemon->run;
+  $daemon->reverse_proxy(1) if @proxy;
+  $daemon->trusted_proxies([grep {length} @proxy]);
+  return $daemon;
 }
+
+sub run { shift->build_server(@_)->run }
 
 1;
 
@@ -38,10 +42,11 @@ Mojolicious::Command::daemon - Daemon command
   Usage: APPLICATION daemon [OPTIONS]
 
     ./myapp.pl daemon
-    ./myapp.pl daemon -m production -l http://*:8080
+    ./myapp.pl daemon -m production -p -l http://*:8080
     ./myapp.pl daemon -l http://127.0.0.1:8080 -l https://[::]:8081
     ./myapp.pl daemon -l 'https://*:443?cert=./server.crt&key=./server.key'
     ./myapp.pl daemon -l http+unix://%2Ftmp%2Fmyapp.sock
+    ./myapp.pl daemon -l http://127.0.0.1:8080 -p 127.0/8 -p fc00::/7
 
   Options:
     -b, --backlog <size>                 Listen backlog size, defaults to
@@ -62,9 +67,11 @@ Mojolicious::Command::daemon - Daemon command
     -m, --mode <name>                    Operating mode for your application,
                                          defaults to the value of
                                          MOJO_MODE/PLACK_ENV or "development"
-    -p, --proxy                          Activate reverse proxy support,
+    -p, --proxy [<network>]              Activate reverse proxy support,
                                          defaults to the value of
-                                         MOJO_REVERSE_PROXY
+                                         MOJO_REVERSE_PROXY, optionally takes
+                                         one or more trusted proxy addresses or
+                                         networks
     -r, --requests <number>              Maximum number of requests per
                                          keep-alive connection, defaults to 100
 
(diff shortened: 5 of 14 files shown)


lib/Mojo/Message/Request.pm
```diff
@@ -301,6 +302,13 @@ Proxy URL for request.
 
 Request has been performed through a reverse proxy.
 
+=head2 trusted_proxies
+
+  my $proxies = $req->trusted_proxies;
+  $req        = $req->trusted_proxies(['10.0/8', '127.0/8', '172.16.0/12', '192.168.0/16']);
```
kraih, line 308: One of the examples in the middle should probably be a single address?

lib/Mojo/Server.pm
```diff
@@ -21,6 +22,7 @@ sub build_app {
 sub build_tx {
   my $self = shift;
   my $tx   = $self->app->build_tx;
+  if (my $p = $self->trusted_proxies) { $tx->req->trusted_proxies($p) }
```
kraih, line 25: Looks like this is always true.

lib/Mojo/Server.pm
```diff
@@ -130,7 +132,16 @@ Application this server handles, defaults to a L<Mojo::HelloWorld> object.
...
-This server operates behind a reverse proxy, defaults to the value of the C<MOJO_REVERSE_PROXY> environment variable.
+This server operates behind a reverse proxy, defaults to the value of the C<MOJO_REVERSE_PROXY> environment variable
+or true if L</trusted_proxies> is not empty.
+
+=head2 trusted_proxies
+
+  my $proxies = $server->trusted_proxies;
+  $server     = $server->trusted_proxies(['10.0/8', '127.0/8', '172.16.0/12', '192.168.0/16']);
+
+This server expects requests from trusted reverse proxies, defaults to the value of C<MOJO_TRUSTED_PROXIES> environment
```
kraih, line 143: `... value of the ...`.

lib/Mojo/Transaction.pm
```diff
@@ -242,8 +252,11 @@ False, this is not a L<Mojo::Transaction::WebSocket> object.
   my $address = $tx->remote_address;
   $tx         = $tx->remote_address('127.0.0.1');
 
-Same as L</"original_remote_address"> or the last value of the C<X-Forwarded-For> header if L</"req"> has been
-performed through a reverse proxy.
+Same as L</"original_remote_address"> unless the L</"req"> has been performed via a
```
kraih, line 255: `unless the L</"req">` sounds weird, `unless L</"req">` would be similar to how it was before.

lib/Mojo/Util.pm
```diff
@@ -758,6 +783,21 @@ Punycode decode string as described in L<RFC 3492|https://tools.ietf.org/html/rf
...
   punycode_decode 'bcher-kva';
 
+=head2 network_contains
+
+  my $bool = network_contains $network, $address;
+
+Check that a given address is contained within a network in CIDR form. If the network is a single address, the
+addresses must be equivalent.
+
+  # true
```
kraih, line 793: Think that should be `# True`.

lib/Mojo/Util.pm
```diff
@@ -758,6 +783,21 @@ Punycode decode string as described in L<RFC 3492|https://tools.ietf.org/html/rf
...
+  network_contains('10.0.0.0/8', '10.10.10.10');
+  network_contains('10.10.10.10', '10.10.10.10');
+
+  # false
```
kraih, line 797: `# False`.

lib/Mojolicious/Command/daemon.pm
```diff
@@ -42,6 +44,7 @@ Mojolicious::Command::daemon - Daemon command
     ./myapp.pl daemon -l http://127.0.0.1:8080 -l https://[::]:8081
     ./myapp.pl daemon -l 'https://*:443?cert=./server.crt&key=./server.key'
     ./myapp.pl daemon -l http+unix://%2Ftmp%2Fmyapp.sock
+    ./myapp.pl prefork -l http://127.0.0.1:8080 -t 127.0/8 -t 192.168.0/16
```
kraih, line 47: This is `daemon`, not `prefork`.

lib/Mojo/Server.pm
jberger: Yeah, I wondered about that last night. Should I then also check `@$p` or just assign it always?

lib/Mojo/Transaction.pm
jberger: I was reading it as "unless the request" but as code losing the "the" is probably better. I'm fine with either.

lib/Mojolicious/Command/daemon.pm
jberger: good catch, copypasta

t/mojo/psgi.t
```diff
@@ -241,4 +241,76 @@ subtest 'Reverse proxy' => sub {
...
+  my %headers = @{$res->[1]};
+  is $headers{'Content-Length'}, 5,                         'right "Content-Length" value';
+  is $headers{'Content-Type'},   'text/html;charset=UTF-8', 'right "Content-Type" value';
+  my $body = '';
+  while (defined(my $chunk = $res->[2]->getline)) { $body .= $chunk }
+  is $body, 'https', 'right content';
+  is $ENV{MOJO_HELLO}, undef, 'finish event has not been emitted';
+  $res->[2]->close;
+  is delete $ENV{MOJO_HELLO}, 'world', 'finish event has been emitted';
+};
```
kraih, line 315: Missing newline.

lib/Mojo/Server.pm
kraih: 🤷

jberger: Does `if (my $p = $self->trusted_proxies) { $tx->req->trusted_proxies($p) if @$p }` seem weird? It seems weird to me.

kraih: Very weird.

lib/Mojo/Server.pm
```diff
@@ -130,7 +132,17 @@ Application this server handles, defaults to a L<Mojo::HelloWorld> object.
...
   $server  = $server->reverse_proxy($bool);
 
-This server operates behind a reverse proxy, defaults to the value of the C<MOJO_REVERSE_PROXY> environment variable.
+This server operates behind a reverse proxy, defaults to the value of the C<MOJO_REVERSE_PROXY> environment variable
+or true if L</trusted_proxies> is not empty.
+
+=head2 trusted_proxies
+
+  my $proxies = $server->trusted_proxies;
+  $server     = $server->trusted_proxies(['10.0/8', '127.0/8', '172.16.0/12', '192.168.0/16']);
```
kraih, line 141: Shouldn't the examples be consistent?

jberger: oh, yes.

Sidenote: i had originally included these 4 original cidrs are they are the private network space defined for these purposes. I'm changing 128.0/8 to 127.0.0.1 because it is probably the most common usage of a private space and I rarely if ever see anyone use anything but .1

lib/Mojo/Server.pm
jberger: How about `push @{$tx->req->trusted_proxies}, @{$self->trusted_proxies};`? It has the advantage of shallow cloning (which we probably want anyway) and it avoids lots of tests.

jberger: Also, since this happens after the build tx hook, it is possible they will have set something on the req object.

jberger: I've gone with this approach.

jberger:
I know we're into the New Year's holiday so I don't expect it to move quickly but I think this PR is ready, I'm calling the vote: @mojolicious/maintainers

lib/Mojo/Message/Request.pm
```diff
@@ -301,6 +302,13 @@ Proxy URL for request.
 
 Request has been performed through a reverse proxy.
 
+=head2 trusted_proxies
+
+  my $proxies = $req->trusted_proxies;
+  $req        = $req->trusted_proxies(['10.0/8', '127.0.0.1', '172.16.0/12', '192.168.0/16']);
```
jhthorsen, line 308: Could there be a IPv6 address in this example just to make it more obvious v6 supported?

lib/Mojo/Util.pm
```diff
@@ -183,6 +184,30 @@ sub monkey_patch {
   *{"${class}::$_"} = set_subname("${class}::$_", $patch{$_}) for keys %patch;
 }
 
+sub network_contains {
```
jhthorsen, line 187: I'm not sure if I understand what this function does by looking at the name. Could it be cidr_contains($cidr, $addr), ip_within($addr, $cidr) or something else that implies that it's working with IP addresses?

lib/Mojolicious/Command/daemon.pm
```diff
@@ -19,9 +19,11 @@ sub run {
     'k|keep-alive-timeout=i' => sub { $daemon->keep_alive_timeout($_[1]) },
     'l|listen=s'             => \my @listen,
     'p|proxy'                => sub { $daemon->reverse_proxy(1) },
-    'r|requests=i'           => sub { $daemon->max_requests($_[1]) };
+    'r|requests=i'           => sub { $daemon->max_requests($_[1]) },
+    't|trusted-proxy=s'      => \my @trusted;
```
jhthorsen, line 23: Could we reuse proxy instead? Something like `p|proxy:s{,1}`?

jhthorsen (commented):
I'm positive in general to this change. Seems to solve a common problem.

lib/Mojo/Util.pm
kraih: I find the name descriptive enough.

jberger: A colleague suggested `cidr_contains` as well and I could be convinced, but I personally think that's too "wonky" of a name. You have to know what it is to know what it is in a way. `network_contains` is similarly accurate and more descriptive to the outside reader imo.

lib/Mojolicious/Command/daemon.pm
jberger: Having just reverse proxy mode without trusted ips still has meaning, namely a simple "skip one level". Indeed -t still requires (and implicitly implies) -r because of how the scheme handling works. I can also imagine an extension to -r that would say how many levels to skip, like `-r 2`. All that just goes to say these aren't exactly the same concept and thus I don't want to conflate them at this outermost layer when they aren't internally.

lib/Mojo/Message/Request.pm
jberger: done

kraih:
Like i said on IRC, i'm neutral on this PR.

lib/Mojolicious/Command/daemon.pm
jhthorsen: -r or -p..? -p is a boolean flag, which I can’t imagine need to be a supported IP address. That’s why I suggested keeping both. I haven’t tried {,} myself though, so I’m unsure how it works with zero arguments.

lib/Mojo/Util.pm
jhthorsen: I suggested ip_within(), since NetAddr::IP has a “within” method. Not a deal breaker for me, but I think it would be nice if we could steal a name from somewhere else.

lib/Mojolicious/Command/daemon.pm
kraih: With zero arguments an empty string is passes to the closure.

marcusramberg (approved):
I think this is a good approach to a common problem 👍

lib/Mojolicious/Command/daemon.pm
jberger: Apologies, I meant -p not -r

lib/Mojo/Util.pm
jberger: There is no consistency on CPAN. Net::IP calls it "overlaps" with some exported constants to check for different types of overlap, Net::IP::Lite(::Net) calls it "contains" which is where I got my name.

lib/Mojolicious/Command/daemon.pm
kraih: Fwiw. I don't see us adding `-p 2` when `-p 127.0.0.1` already works.

lib/Mojolicious/Command/daemon.pm
```diff
@@ -18,10 +18,13 @@ sub run {
...
     'k|keep-alive-timeout=i' => sub { $daemon->keep_alive_timeout($_[1]) },
     'l|listen=s'             => \my @listen,
-    'p|proxy'                => sub { $daemon->reverse_proxy(1) },
+    'p|proxy:s'              => \my @proxy,
     'r|requests=i'           => sub { $daemon->max_requests($_[1]) };
 
   $daemon->listen(\@listen) if @listen;
+  $daemon->reverse_proxy(1) if @proxy;
+  my @trusted = grep {length} @proxy;
+  $daemon->trusted_proxies(\@trusted) if @trusted;
```
kraih, line 27: Pretty sure this can be done with less code too.

kraih (dismissed):
Considering this is a security relevant feature, i think we might need a test in `t/mojolicious/commands.t` that ensures `-p` options end up in the right place.

lib/Mojolicious/Command/prefork.pm
```diff
@@ -23,12 +23,21 @@ sub run {
...
-    'p|proxy'                => sub { $prefork->reverse_proxy(1) },
+    'p|proxy:s'              => \my @proxy,
     'r|requests=i'           => sub { $prefork->max_requests($_[1]) },
     's|spare=i'              => sub { $prefork->spare($_[1]) },
     'w|workers=i'            => sub { $prefork->workers($_[1]) };
 
   $prefork->listen(\@listen) if @listen;
+  $prefork->reverse_proxy(1) if @proxy;
+  my @trusted = grep {length} @proxy;
+  $prefork->trusted_proxies(\@trusted) if @trusted;
```
kraih, line 34: Why not just `$prefork->trusted_proxies([grep {length} @proxy]);`?

lib/Mojolicious/Command/prefork.pm
```diff
@@ -23,12 +23,21 @@ sub run {
     'k|keep-alive-timeout=i' => sub { $prefork->keep_alive_timeout($_[1]) },
     'l|listen=s'             => \my @listen,
     'P|pid-file=s'           => sub { $prefork->pid_file($_[1]) },
-    'p|proxy'                => sub { $prefork->reverse_proxy(1) },
+    'p|proxy:s'              => \my @proxy,
```
kraih, line 26: Something like this would probably work too (with less code):
```
'p|proxy:s' => sub { $_[1] eq '' ? $daemon->reverse_proxy(1) : push @{$daemon->trusted_proxies}, $_[1] },
```

kraih:
This PR has been broken for a few days, so the vote is off for now and we'll consider it a work in progress.

jberger:
I may get some time to work on it again today, but yes, I've considered the vote on hold since you blocked it several days ago

t/mojolicious/commands.t
```diff
@@ -426,19 +455,48 @@ my $inflate = Mojolicious::Command::Author::inflate->new;
...
+subtest 'prefork' => sub {
+  require Mojolicious::Command::prefork;
+
+  subtest 'description' => sub {
+    my $command = Mojolicious::Command::prefork->new;
+    ok $command->description, 'has a description';
+    like $command->usage, qr/prefork/, 'has usage information';
+  };
+
+  subtest 'unknown option' => sub {
```
kraih, line 467: Wrong capitalization (same for the subtests below).

lib/Mojolicious/Command/daemon.pm
```diff
@@ -18,10 +18,19 @@ sub run {
...
+  return $daemon;
+}
+
+sub run {
+  my ($self, @args) = @_;
+  my $daemon = $self->build_server(@args);
   $daemon->run;
```
kraih, line 34: `$self->build_server(@args)->run`?

kraih: Or actually:
```perl
sub run { shift->build_server(@_)->run }
```

kraih:
Looks like this is still unfinished, so i'll leave it out of the next release.

lib/Mojolicious/Command/daemon.pm
jhthorsen: How about just `$daemon->trusted_proxies([grep {length} @proxy])` ?

lib/Mojolicious/Command/prefork.pm
jberger: I guess I was saving a method call if it wasn't needed, but this isn't going to be a hot code path, so it doesn't matter to me

jberger: But I guess that first one handles all of this so, ok

lib/Mojolicious/Command/prefork.pm
jberger: this ends up being too long

jberger:
Assuming @kraih doesn't block on security concerns, I'm calling the vote again @mojolicious/maintainers

kraih (commented):
I'm neutral on this.

marcusramberg (dismissed):
I wish other web frameworks I have to deal with supported this. Am 👍

t/mojolicious/lite_app.t
```diff
@@ -662,21 +662,52 @@ $t->get_ok('/.html')->status_is(200)->header_exists_not('Servers')
...
+
+subtest 'Reverse proxy with "X-Forwarded-For" and trusted proxies (all addresses trusted)' => sub {
+  local $ENV{MOJO_TRUSTED_PROXIES} = '0/0';
+  my $t = Test::Mojo->new;
+  $t->get_ok('/0' => {'X-Forwarded-For' => '192.0.2.2, 192.0.2.1'})->status_is(200)
+    ->header_unlike('X-Original' => qr/192\.0\.2\.(?:2|1)/)
+    ->content_like(qr!http://127\.0\.0\.1:\d+/0-192\.0\.2\.2-0$!);
+};
+
+
```
kraih, line 704: Double newline!

lib/Mojo/Util.pm
```diff
@@ -183,6 +184,30 @@ sub monkey_patch {
...
+  return undef if $v6 xor $addr =~ /:/;
+
+  # Convert addresses to binary
+  return undef unless $net  = $v6 ? inet_pton(AF_INET6, $net)  : inet_aton($net);
+  return undef unless $addr = $v6 ? inet_pton(AF_INET6, $addr) : inet_aton($addr);
+  my $length = $v6 ? 128 : 32;
+
+  # Apply mask if given
+  if (defined $mask) {
+    $mask = pack "B$length", "1" x $mask;
```
kraih, line 203: `"1"` should be `'1'` i guess.

marcusramberg (dismissed):
I hereby rebless this pr

shadowcat-mst:
LGTM

t/mojo/cgi.t
```diff
@@ -179,4 +179,53 @@ subtest 'Reverse proxy' => sub {
...
+  open STDOUT, '>', \$msg;
+  local %ENV = (
+    PATH_INFO                => '/proxy',
+    REQUEST_METHOD           => 'GET',
+    SCRIPT_NAME              => '/',
+    HTTP_HOST                => 'localhost:8080',
+    REMOTE_ADDR              => '127.0.0.1',
+    SERVER_PROTOCOL          => 'HTTP/1.0',
+    'HTTP_X_Forwarded_For'   => '10.10.10.10, 192.0.2.2, 192.0.2.1',
+    'HTTP_X_Forwarded_Proto' => 'https'
```
jhthorsen, line 194: No need for quotes. (This is a recurring typo for all the HTTP_X headers)

shadowcat-mst (approved):
LGTM

lib/Mojo/Transaction.pm
```diff
@@ -36,7 +37,16 @@ sub remote_address {
...
   # Reverse proxy
-  return ($self->req->headers->header('X-Forwarded-For') // '') =~ /([^,\s]+)$/ ? $1 : $self->original_remote_address;
+  my @addrs   = split /\s*,\s*/, ($self->req->headers->header('X-Forwarded-For') // '');
+  my $trusted = $self->req->trusted_proxies;
+  return @addrs ? $addrs[-1] : $self->original_remote_address unless @$trusted;
+
+  push @addrs, $self->original_remote_address;
+  for my $addr (reverse @addrs) {
+    next if any { network_contains($_, $addr) } @$trusted;
+    return $addr;
```
jhthorsen, line 47: I think you can save one line here by doing `return $addr unless...`

lib/Mojo/Util.pm
```diff
@@ -183,6 +184,30 @@ sub monkey_patch {
...
+  # Convert addresses to binary
+  return undef unless $net  = $v6 ? inet_pton(AF_INET6, $net)  : inet_aton($net);
+  return undef unless $addr = $v6 ? inet_pton(AF_INET6, $addr) : inet_aton($addr);
+  my $length = $v6 ? 128 : 32;
+
+  # Apply mask if given
+  if (defined $mask) {
+    $mask = pack "B$length", '1' x $mask;
+    $addr &= $mask;
+  }
```
jhthorsen, line 205: Can this be golfed into `$addr &= ($mask = pack "B$length", '1' x $mask) if defined $mask`?

jhthorsen (commented):
I'm not going to "Request changes", since I think my comments are very minor.

lib/Mojo/Util.pm
jberger: wow

lib/Mojo/Server.pm
```diff
@@ -8,8 +8,9 @@ use Mojo::Util qw(md5_sum);
 use POSIX ();
 use Scalar::Util qw(blessed);
 
-has app           => sub { shift->build_app('Mojo::HelloWorld') };
-has reverse_proxy => sub { $ENV{MOJO_REVERSE_PROXY} };
+has app             => sub { shift->build_app('Mojo::HelloWorld') };
+has reverse_proxy   => sub { $ENV{MOJO_REVERSE_PROXY} || @{shift->trusted_proxies} };
```
jberger, line 12: should this be `sub { $ENV{MOJO_REVERSE_PROXY} || !!@{shift->trusted_proxies} };`

lib/Mojo/Util.pm
jberger: Actually I don't even need to keep the updated `$mask` variable so it doesn't need the extra complexity of the assignment inside the parens

tianon (commented):
All the tests I see here seem to be testing for mostly sane/valid/benign `X-Forwarded-For` header values -- do you think it would be valuable to include at least one instance of the type of malicious header this is intended to guard against returning malicious values for?

ie, instead of just `10.10.10.10, 192.0.2.2, 192.0.2.1`, perhaps at least one instance of something more malicious like `172.16.6.6, 127.0.0.1, 10.10.10.10, 192.0.2.2, 192.0.2.1`?
(where `10.10.10.10` is the "correct" answer given `127.0/8, 192.0/8`)

jhthorsen (approved):
LGTM 👍

marcusramberg (approved):
Third time lucky!

lib/Mojolicious/Guides/Cookbook.pod
```diff
@@ -106,7 +106,10 @@ instead of something meaningful for the client, like this:
 
 To solve these problems, you can configure your reverse proxy to send the missing data (see L</Nginx> and
 L</"Apache/mod_proxy">) and tell your application about it by setting the environment variable C<MOJO_REVERSE_PROXY>.
-For finer control, L</Rewriting> includes examples of how the changes could be implemented manually.
+In more complex situations, usually involving multiple proxies or proxies that live outside your network, it can be
+necessary to tell the application from which ip addresses to expect proxy requests by setting C<MOJO_TRUSTED_PROXIES>
+to a list of comma separated addresses or CIDR networks. For even finer control, L</Rewriting> includes examples of how
+the changes could be implemented manually.
```
tianon, line 112: I think my own misunderstanding of this PR (which I resolved via discussion on IRC) would've benefited from a concrete example here similar to the ones above this paragraph; essentially explicitly pointing out that `X-Forwarded-For: 7.7.7.7, 192.0.2.2, 192.0.2.1` + `MOJO_TRUSTED_PROXIES='127.0/8, 192.0.2.1'` = `192.0.2.2`.

```pod
In more complex situations, usually involving multiple proxies or proxies that live outside your network, it can be
necessary to tell the application from which ip addresses to expect proxy requests by setting C<MOJO_TRUSTED_PROXIES>
to a list of comma separated addresses or CIDR networks.

For example, setting C<MOJO_TRUSTED_PROXIES> to C<127.0/8, 192.0.2.1> will resolve an C<X-Forwarded-For> header value of
C<7.7.7.7, 192.0.2.2, 192.0.2.1> to C<192.0.2.2>.

For even finer control, L</Rewriting> includes examples of how the changes could be implemented manually.
```

PR #1647, opened 2021-01-06
Use a constant for empty arrayref defaults

ilmari:
### Summary

This micro-optimises cases where the arraref is immediately
dereferenced and not passed to things that might subsequently modify it.

### Motivation

This replaces the anonlist+pushmark of [] with a const RV in the
optree, which is nearly three times as fast, according to this
microbenchmark:

    $ perl -MBenchmark::Dumb=cmpthese -e '
      use constant empty => []; my $foo;
      cmpthese(0.001, {
        const => sub { @{ $foo // empty } },
        array => sub { @{ $foo // [] } },
      })'
                         Rate       array  const
    array 1.2869e+07+-22000/s          -- -75.1%
    const 5.166e+07+-220000/s 301.4+-1.8%     --

### DIFF
--- a/lib/Mojo/IOLoop/Subprocess.pm
+++ b/lib/Mojo/IOLoop/Subprocess.pm
@@ -6,6 +6,7 @@ use Mojo::IOLoop;
 use Mojo::IOLoop::Stream;
 use Mojo::JSON;
 use Mojo::Promise;
+use Mojo::Util;
 use POSIX ();
 
 has deserialize => sub { \&Mojo::JSON::decode_json };
@@ -50,7 +51,7 @@ sub _start {
   unless ($pid) {
     eval {
       $self->ioloop->reset({freeze => 1});
-      my $results = eval { [$self->$child] } // [];
+      my $results = eval { [$self->$child] } // Mojo::Util::_EMPTY_ARRAY;
       print {$self->{writer}} '0-', $self->serialize->([$@, @$results]);
       $self->emit('cleanup');
     } or warn $@;
@@ -81,7 +82,7 @@ sub _start {
       waitpid $pid, 0;
       $self->{exit_code} = $? >> 8;
       substr $buffer, 0, 2, '';
-      my $results = eval { $self->deserialize->($buffer) } // [];
+      my $results = eval { $self->deserialize->($buffer) } // Mojo::Util::_EMPTY_ARRAY;
       $self->$parent(shift(@$results) // $@, @$results);
     }
   );
--- a/lib/Mojo/Util.pm
+++ b/lib/Mojo/Util.pm
@@ -23,6 +23,12 @@ use Unicode::Normalize ();
 # Check for monotonic clock support
 use constant MONOTONIC => eval { !!Time::HiRes::clock_gettime(Time::HiRes::CLOCK_MONOTONIC()) };
 
+# Speed up @{ $foo // [] } and similar by avoiding having to construct a new empty arrayref each time
+use constant _EMPTY_ARRAY => [];
+
+# Make sure nobody accidentally modifies the referenced array
+BEGIN { Internals::SvREADONLY(@{+_EMPTY_ARRAY}, 1) }
+
 # Punycode bootstring parameters
 use constant {
   PC_BASE         => 36,
--- a/lib/Mojo/DOM.pm
+++ b/lib/Mojo/DOM.pm
@@ -12,6 +12,7 @@ use overload
 use Mojo::Collection;
 use Mojo::DOM::CSS;
 use Mojo::DOM::HTML;
+use Mojo::Util;
 use Scalar::Util qw(blessed weaken);
 use Storable qw(dclone);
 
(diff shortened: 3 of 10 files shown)


lib/Mojo/Util.pm
```diff
@@ -23,6 +23,11 @@ use Unicode::Normalize ();
 # Check for monotonic clock support
 use constant MONOTONIC => eval { !!Time::HiRes::clock_gettime(Time::HiRes::CLOCK_MONOTONIC()) };
 
+# For speeding up @{ $foo // [] }
+use constant _empty => [];
+# make sure nobody accidentally modifies the referenced array
```
kraih, line 28: Inconsistent comment.

lib/Mojo/Util.pm
```diff
@@ -23,6 +23,11 @@ use Unicode::Normalize ();
 # Check for monotonic clock support
 use constant MONOTONIC => eval { !!Time::HiRes::clock_gettime(Time::HiRes::CLOCK_MONOTONIC()) };
 
+# For speeding up @{ $foo // [] }
+use constant _empty => [];
```
kraih, line 27: We don't use lowercase constants.

kraih: Something like `_EMPTY_ARRAY` would also be more descriptive.

kraih:
Are there any benchmarks for a real world use case where this makes a meaningful difference?

lib/Mojo/Util.pm
ilmari: Reworded (also expanded the first comment a bit)

lib/Mojo/Util.pm
ilmari: Fixed.

ilmari:
> Are there any benchmarks for a real world use case where this makes a meaningful difference?

No, but it might help @jberger's [proposed implementation](http://paste.debian.net/1179902/) of `->on('*' => sub { … })`

lib/Mojo/Util.pm
```diff
@@ -23,6 +23,12 @@ use Unicode::Normalize ();
 # Check for monotonic clock support
 use constant MONOTONIC => eval { !!Time::HiRes::clock_gettime(Time::HiRes::CLOCK_MONOTONIC()) };
 
+# Speed up @{ $foo // [] } and similar by avoiding having to construct
+# a new empty arrayref each time
```
kraih, line 27: Wrong line length.

lib/Mojo/Util.pm
```diff
@@ -23,6 +23,12 @@ use Unicode::Normalize ();
...
+use constant _EMPTY_ARRAY => [];
+# Make sure nobody accidentally modifies the referenced array
+BEGIN { Internals::SvREADONLY(@{+_EMPTY_ARRAY}, 1) }
```
kraih, line 30: Did that really run through perltidy?

lib/Mojo/IOLoop/Subprocess.pm
```diff
@@ -50,7 +51,7 @@ sub _start {
   unless ($pid) {
     eval {
       $self->ioloop->reset;
-      my $results = eval { [$self->$child] } || [];
+      my $results = eval { [$self->$child] } // Mojo::Util::_EMPTY_ARRAY;
```
kraih, line 54: Isn't defined or slower than or?

lib/Mojo/Util.pm
ilmari: The only thing perltidy did was add an empty line after the `use constant`, which seems silly. I've instead moved both the comment lines together before the block, and joined the with a `, and`.  Breaking the line after the comma seemed more sensible than forcing the first line to the full 120 characters specified in `.perltidyrc`.

lib/Mojo/IOLoop/Subprocess.pm
ilmari: No, `//` only has to check the `SVf_OK` flag, while `||` has to check the actual value to see if it's truthy.

lib/Mojo/Util.pm
kraih: If you want to start an argument about our formatting rules here, i have to block this PR until that is resolved.

kraih (dismissed):

lib/Mojo/Util.pm
ilmari: Not really. I've split it back into to separate blocks now, with an empty line between, per perltidy.

kraih:
Ok, on its own i'm 👎 on this PR, since there's no real gain. We'll have to see about `Mojo::EventEmitter` changes.

jhthorsen:
I feel this change is a bit dangerous, since I think it will be difficult to spot if we by accident start returning `Mojo::Util::_EMPTY_ARRAY`. Based on that, I have to give this PR 👎.

kraih:
Thanks for doing the research though, that will probably be very helpful with optimizations in the future.

ilmari:
> I feel this change is a bit dangerous, since I think it will be difficult to spot if we by accident start returning `Mojo::Util::_EMPTY_ARRAY`.

That's why I made it `SvREADONLY`, so that at least modifications get caught.  We could also make a policy that it is only allowed to be used directly inside an array dereference, i.e. `@{ ... // Mojo::Util::_EMPTY_ARRAY}`. It shouldn't be too hard to rejig the few places that don't currently.

> Based on that, I have to give this PR -1.

I don't have a particularly strong opinion on this, so feel free to close the PR.

kraih:
We'll have to see if this gets added as part of a future `Mojo::EventEmitter` patch.

Issue #1499, opened 2020-04-23
Don't stringify logged Mojo::Exceptions


jrubinator:
* Mojolicious version: 8.40
* Perl version: 5.18
* Operating system: Centos6

### Short overview of this request
We're using https://metacpan.org/pod/Mojolicious::Plugin::Log::Any to reroute app logs through `Log::Any` to an external bug tracker (Sentry). In 8.24, `Mojo::Log` began stringifying requests earlier (https://metacpan.org/source/SRI/Mojolicious-8.24/lib/Mojo/Log.pm#L72) in a way that prevents us from getting metadetails about what went wrong (such as the stack trace).

### Steps to reproduce the behavior
Logging a `Mojo::Exception` results in premature stringification of the exception when the app log has a context. This occurs by default on exceptions as a combination of
https://metacpan.org/source/SRI/Mojolicious-8.40/lib/Mojo/Log.pm#L72
and https://metacpan.org/source/SRI/Mojolicious-8.40/lib/Mojolicious/Plugin/DefaultHelpers.pm#L176

To see this in action:
```
$ cat log.t
use strict;
use warnings;

use Log::Any::Adapter;
use Test::Mojo;
use Test::More;

{
    package Log::Any::Adapter::Mojo::Test; # this package wants an error object
    use base qw(Log::Any::Adapter::Base);

    use Test::More;

    sub structured {
        my ($self, $level, $category, @log_args) = @_;

        # This fails!
        isa_ok($log_args[0], 'Mojo::Exception')
            or diag explain $log_args[0];
        # Do something with Mojo::Exception
    }

    sub is_error { 1 }
    sub is_debug { 0 }
}

{
    package Mojolicious::Test::Log; # an app that dies
    use Mojolicious::Lite;
    use Mojolicious::Plugin::Log::Any;

    plugin 'Log::Any';

    get '/error' => sub { die "Log Me" };
}

subtest test_exceptions => sub {
    my $t = Test::Mojo->new('Mojolicious::Test::Log');
    Log::Any::Adapter->set('Mojo::Test');

    # this succeeds, but the logging itself fails
    $t->get_ok('/error')
      ->status_is(500 => 'threw exception');
};

done_testing();
$ prove log.t
```

### Expected behavior

The test should pass.
That would mean that we're not stringifying a `Mojo::Exception`, and whatever log mechanism is in use can eg. extract a stack trace from the exception.

[...]

PR #1650, opened 2021-01-08
Delay app log stringification

jrubinator:
### Summary
This tweaks `Mojo::Log` to delay `Mojo::Exception` stringification until *after* it emits the message.

### Motivation
The primary motivation is to give non-default loggers (like [Mojolicious::Plugin::Log::Any](https://metacpan.org/pod/Mojolicious%3A%3APlugin%3A%3ALog%3A%3AAny)) direct access to the `Mojo::Exception` being logged (which allows logging of a stack trace via eg. [Log::Any::Adapter::Sentry::Raven](https://metacpan.org/pod/Log%3A%3AAny%3A%3AAdapter%3A%3ASentry%3A%3ARaven)).

A secondary motivation is to provide consistency in what non-default loggers receive, regardless of whether the sending logger has a context or not.

### References
This was originally reported in #1499.
There has been some discussion about this on IRC. One concern raised was that of performance. To that end, https://gist.github.com/jrubinator/387dcac4635942641bc85b82e91f41e4 details what I saw regarding performance. Quoting,

> I saw little, if any, performance regression.
> ...
> 
> This table contains the net requests/second (across all threads)
> The first one is from the one-minute run.
> The second is from the five-minute run.
> ...
>
> TL;DR
> Setup | Regular-Call | Mojo-Exception | Custom-Exception
> ------------ | -------------|------------ | -------------
> Old Code+Development   |  809.63 / 803.89 |    187.36/246.77  | 776.23/764.71
> Old Code+Production   |  1098.18/1102.15 |    442.26/420.58  | 935.83/916.78
> New Code+Development |     805.10/ 803.28 |    181.07/247.63  | 761.39/759.00
> New Code+Production    | 1105.91/1108.10    | 429.07/415.89  | 909.28/910.12

### DIFF
--- a/lib/Mojo/Log.pm
+++ b/lib/Mojo/Log.pm
@@ -5,6 +5,7 @@ use Carp qw(croak);
 use Fcntl qw(:flock);
 use Mojo::File;
 use Mojo::Util qw(encode);
+use Scalar::Util qw(blessed);
 use Time::HiRes qw(time);
 
 has format => sub { shift->short ? \&_short : \&_default };
@@ -66,7 +67,20 @@ sub _default {
 sub _log {
   my ($self, $level) = (shift, pop);
   my @msgs = ref $_[0] eq 'CODE' ? $_[0]() : @_;
-  $msgs[0] = "$self->{context} $msgs[0]" if $self->{context};
+  if ($self->{context}) {
+    if (@msgs == 1 && blessed($msgs[0]) && $msgs[0]->isa('Mojo::Exception')) {
+      $msgs[0] = Mojo::Exception
+        ->new("$self->{context} ".$msgs[0]->message)
+        ->frames($msgs[0]->frames)
+        ->line($msgs[0]->line)
+        ->lines_after($msgs[0]->lines_after)
+        ->lines_before($msgs[0]->lines_before)
+        ->verbose($msgs[0]->verbose);
+    }
+    else {
+      $msgs[0] = "$self->{context} $msgs[0]";
+    }
+  }
   ($self->{parent} || $self)->emit('message', $level, @msgs);
 }
 
--- a/lib/Mojo/Log.pm
+++ b/lib/Mojo/Log.pm
@@ -69,13 +69,9 @@ sub _log {
   my @msgs = ref $_[0] eq 'CODE' ? $_[0]() : @_;
   if ($self->{context}) {
     if (@msgs == 1 && blessed($msgs[0]) && $msgs[0]->isa('Mojo::Exception')) {
-      $msgs[0] = Mojo::Exception
-        ->new("$self->{context} ".$msgs[0]->message)
-        ->frames($msgs[0]->frames)
-        ->line($msgs[0]->line)
-        ->lines_after($msgs[0]->lines_after)
-        ->lines_before($msgs[0]->lines_before)
-        ->verbose($msgs[0]->verbose);
+      $msgs[0]
+        = Mojo::Exception->new("$self->{context} " . $msgs[0]->message)->frames($msgs[0]->frames)->line($msgs[0]->line)
+        ->lines_after($msgs[0]->lines_after)->lines_before($msgs[0]->lines_before)->verbose($msgs[0]->verbose);
     }
     else {
       $msgs[0] = "$self->{context} $msgs[0]";
--- a/t/mojolicious/log_delayed_stringification.t
+++ b/t/mojolicious/log_delayed_stringification.t
@@ -5,10 +5,11 @@ use Test::Mojo;
 use Test::More;
 
 {
-    package Mojolicious::Test::Log; # an app that dies
-    use Mojolicious::Lite;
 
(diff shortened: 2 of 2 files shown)


kraih:
This actually seems less consistent to me.

lib/Mojo/Log.pm
```diff
@@ -66,7 +67,20 @@ sub _default {
...
-  $msgs[0] = "$self->{context} $msgs[0]" if $self->{context};
+  if ($self->{context}) {
+    if (@msgs == 1 && blessed($msgs[0]) && $msgs[0]->isa('Mojo::Exception')) {
+      $msgs[0] = Mojo::Exception
+        ->new("$self->{context} ".$msgs[0]->message)
+        ->frames($msgs[0]->frames)
+        ->line($msgs[0]->line)
+        ->lines_after($msgs[0]->lines_after)
+        ->lines_before($msgs[0]->lines_before)
+        ->verbose($msgs[0]->verbose);
```
kraih, line 78: Why so complicated?

lib/Mojo/Log.pm
```diff
@@ -66,7 +67,20 @@ sub _default {
 sub _log {
   my ($self, $level) = (shift, pop);
   my @msgs = ref $_[0] eq 'CODE' ? $_[0]() : @_;
-  $msgs[0] = "$self->{context} $msgs[0]" if $self->{context};
+  if ($self->{context}) {
+    if (@msgs == 1 && blessed($msgs[0]) && $msgs[0]->isa('Mojo::Exception')) {
+      $msgs[0] = Mojo::Exception
+        ->new("$self->{context} ".$msgs[0]->message)
```
kraih, line 73: There is no way this ever went through perltidy.

t/mojolicious/log_delayed_stringification.t
```diff
@@ -0,0 +1,35 @@
...
+        isa_ok($lines[0], 'Mojo::Exception') or diag explain $lines[0];
+    }
+});
+
+# this succeeds, but the isa_ok above fails
+$t->get_ok('/error')
+  ->status_is(500 => 'threw exception');
+
+# sanity check
+ok $logged_an_error, "ran error testing above";
```
kraih, line 33: This whole test is completely inconsistent with our other tests.

kraih:
You also didn't explain what makes `Mojo::Exception` so special that it deserves to be treated differently, while any other object or data structure is still stringified.

kraih:
It seems to me like there is probably something here we can improve, but you got so focused on your one specific `Mojo::Exception` case that the root problem got lost along the way.

lib/Mojo/Log.pm
jrubinator: I don't know a simpler way to copy all the attributes of a `Mojo::Exception` (or `Mojo::Base`) object.

lib/Mojo/Log.pm
jrubinator: Oops, missed that. Should be fixed now.

t/mojolicious/log_delayed_stringification.t
jrubinator: By this, you mean you'd like to see this look more like eg. `t/mojolicious/log_lite_app.t`?

jrubinator:
> You also didn't explain what makes `Mojo::Exception` so special that it deserves to be treated differently, while any other object or data structure is still stringified.

I would love to extend this protection for other objects as well, but don't know how to go about doing so. Is there a way?

I rationalized treating `Mojo::Exception` differently because it's the only object (I know of) that `Mojolicious` expects to pass through logging out-of-the-box.

kraih:
Looking through the `Mojo::Log` code, i don't think this is the right approach. What we should have done in the first place is join multiple messages (`$log->debug('first', 'second')`) with a whitespace instead of newline. That would have made contexts much easier (`$log->debug('[12345]', 'some message with context')`). Perhaps there's a way we can do that in the future somehow.

jrubinator:
Self-directed sigh: I rubber-ducked IRC about that, and talked myself out of it. I couldn't see a non-breaking path forward and assumed the newlines were desired for multiple logs. I agree that would be much cleaner though.

It would be convenient to change with a major version change, though correct me if I'm wrong: I doubt we would want this to drive a major version change.

If we really wanted to use deprecation cycles, maybe something like:
* add an attribute that turns on space-joining. Its initial - and deprecated - default would be off
* (3+ months later) switch the attribute to default to on. Deprecate the attribute itself
* (3+ months later) Remove the attribute

That isn't a pretty workflow.

Does this match up with your perception of where things stand?

I suppose there's also an argument that https://metacpan.org/pod/Mojo::Log#message does not promise a different line per message, and changing that could be a bug. I defer to wiser heads' opinion on this though.

kraih:
We will have a major version change very soon, so a `Mojo::Log` change could be included if a patch is ready.

kraih:
This PR has been obsoleted by #1660.

PR #1659, opened 2021-01-16
Prevent unintended subprocess side effects by freezing the current state

kraih:
This is a possible solution for #1563. It sidesteps the original problem by keeping all the original state active, but hidden away. The event loop is reset with a fresh reactor instance. In the case of `Mojo::Reactor::EV` this means that all subprocesses will end up with `Mojo::Reactor::Poll`. Which is probably an advantage.

### DIFF
--- a/lib/Mojo/IOLoop.pm
+++ b/lib/Mojo/IOLoop.pm
@@ -94,10 +94,17 @@ sub remove {
 }
 
 sub reset {
-  my $self = _instance(shift)->emit('reset');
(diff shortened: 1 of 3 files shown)


christopherraa (approved):
Seems like a sensible approach. Liked the tests. They were easy to read and understand.

jhthorsen (commented):
Guess I'm a bit late, but after getting to know that EV is not used in child processes, then I get how this PR works.

So I'm giving it 👍 even though it doesn't matter now.

PR #1661, opened 2021-01-18
t/mojo/dynamic_methods.t now uses subtests.

kberov:
### Summary
t/mojo/dynamic_methods.t now uses subtests.

### Motivation
Contribute to the Subtests iniciative
### References
#1520

### DIFF
--- a/t/mojo/dynamic_methods.t
+++ b/t/mojo/dynamic_methods.t
@@ -22,22 +22,24 @@ sub BUILD_DYNAMIC {
 }
 
 package main;
+my ($t1, $t2, $dyn_methods);
+subtest 'Basics' => sub {
+  ($t1, $t2) = (Mojo::TestDynamic->new, Mojo::TestDynamic->new);
+  Mojo::DynamicMethods::register 'Mojo::TestDynamic', $t1->hashref, 'foo', sub { };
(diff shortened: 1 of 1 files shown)


t/mojo/dynamic_methods.t
```diff
@@ -22,22 +22,24 @@ sub BUILD_DYNAMIC {
...
-ok !Mojo::TestDynamic->can('foo'), 'dynamic method is hidden';
-ok eval { $t1->foo; 1 }, 'foo called ok';
-cmp_ok $called_foo, '==', 1, 'called dynamic method';
-ok !eval { $t2->foo; 1 }, 'error calling foo on wrong object';
-
-# Garbage collection
-undef($t1);
-undef($t2);
-ok(!keys(%$dyn_methods), 'dynamic methods expired');
+my ($t1, $t2, $dyn_methods);
```
kraih, line 25: No, just no.

kberov: What would be expected to be done in this case?

kraih (changes requested):
The point of subtests is to make the code cleaner and more readable.

kberov:
Then this file does not need subtests. It is already readable enough.

PR #1668, opened 2021-01-25
Accept MOJO_TRUSTED_PROXIES when -p is not given

jberger:
The golfing of the daemon and prefork commands was too aggressive and it meant that command line invocations essentially could not use the environment variables.


jhthorsen (approved):
LGTM

kraih (approved):

PR #1669, opened 2021-01-25
Add subtests in tls.t and enable TEST_TLS on Linux.

kiwiroy:
### Summary
Continuous integration testing TLS is not currently done. The built-in certificates use only 1024 bits and result in `ee key too small` error. These are regenerated before the tests step here and TLS testing is enabled, but new certs could equally be committed and the extra step removed. The latter should probably be done by a trusted party so I've chosen the former for this PR.

### Motivation
Increasing coverage of CI by testing TLS and continuing move to subtests (`tls.t`)

### References
#1520

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -29,10 +29,39 @@ jobs:
       - name: Install dependencies
         run: |
           cpanm -n --installdeps .
-          cpanm -n Cpanel::JSON::XS EV Role::Tiny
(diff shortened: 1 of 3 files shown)


kraih:
PRs containing multiple changes almost never get enough reviews for merge.

kiwiroy:
The migration to subtests in `tls.t` has been split out. Happy to reduce the size of this change set further if certs with >=2048bits are committed.

kraih:
I think other certs we ship are 4096 bits already.

kraih:
This PR is trying to do too many things at once.

Issue #1466, opened 2020-01-30
Mojo::Request::_parse_env should add the port from SERVER_PORT


EvanCarroll:
### Steps to reproduce the behavior

host mojo on non-port 80 in cgi mode.
calculate a url with url_for()->to_abs->to_string;
see the port is not there.

### Expected behavior
port should be listed.

### Actual behavior
port is not listed.

On current examination what we're doing is parsing `HTTP_HOST`

```
    # Extract headers
    my $headers = $self->headers;
    my $url     = $self->url;
    my $base    = $url->base;
    for my $name (keys %$env) {
      my $value = $env->{$name};
      next unless $name =~ s/^HTTP_//i;
      $name =~ y/_/-/;
      $headers->header($name => $value);

      # Host/Port
      $value =~ s/:(\d+)$// ? $base->host($value)->port($1) : 
  $base->host($value)
        if $name eq 'HOST';
    }
```

But we shouldn't be parsing that at all, because it's non-standard and not mentioned anywhere. What we're supposed to be parsing is `SERVER_PORT` which is actually in the spec, and it's in the tests.

I believe we created the implementation from dumps that offered more than the spec required. But in doing so we shouldn't ignore the spec which says where that information should be sourced.

----

Spec RFC 3875 https://tools.ietf.org/html/rfc3875

> 4.1.15.  SERVER_PORT
> 
>    The SERVER_PORT variable MUST be set to the TCP/IP port number on
>    which this request is received from the client.  This value is used
>    in the port part of the Script-URI.
> 
>       SERVER_PORT = server-port
>       server-port = 1*digit
> 
>    Note that this variable MUST be set, even if the port is the default
>    port for the scheme and could otherwise be omitted from a URI.

kraih:
We need a reproducible test case to investigate. See [existing tests](https://github.com/mojolicious/mojo/blob/master/t/mojo/request_cgi.t) for examples.

kraih:
And in cases where RFCs are involved references to the relevant spec sections. If that's not possible we need as much information as possible about the environment (all the server specific stuff).

EvanCarroll:
@kraih I can provide a new test, but for the record this should be sufficient because it shows

* when there is a difference the one we want -- SERVER_PORT -- is taken.
* implementations that break the spec, we still function and parse the port from http_host.

EvanCarroll:
What exactly is requested? A totally new test that doesn't use common code from the current tests?

kraih:
[...]

PR #1703, opened 2021-02-14
Fixes CGI port

EvanCarroll:
Source the port in CGI from `SERVER_PORT`, as per spec rather than just expecting to find it in the non-started `HTTP_HOST`. Detailed in #1466

### Summary
Previously we ignored `SERVER_PORT`. Now we source it as the primary information and only if it doesn't exist do we fall back to `HTTP_HOST` (which is non-standard). Not in all of the tests prior to this patch `SERVER_PORT` was provided and should have been used per the CGI spec, that it wasn't was a bug. However, the `SERVER_PORT` in all of the tests was also the same as `HTTP_HOST`, so the bug never caught by the tests. Moreover, the bug only occurs when either (a) only `SERVER_PORT` is provided, or (b) the `SERVER_PORT` is different from the `HTTP_HOST`.

### Motivation
This is simply a broken version of CGI as-is. Though it probably only effects you if you run CGI on a non-standard port outside of Apache (which also provides the non-standard `HTTP_HOST` that we accommodate). At cPanel we have our own web server, `cpservd` which does not set `HTTP_HOST`. It does set `SERVER_PORT`. Without this patch url generation in Mojo is broken.

### References
#1466 (issue referencing)
#1467 (last pull request)

### DIFF
--- a/lib/Mojo/Message/Request.pm
+++ b/lib/Mojo/Message/Request.pm
@@ -153,6 +153,9 @@ sub _parse_env {
   my $headers = $self->headers;
   my $url     = $self->url;
   my $base    = $url->base;
+
+  my $port = $env->{SERVER_PORT};
+
   for my $name (keys %$env) {
     my $value = $env->{$name};
     next unless $name =~ s/^HTTP_//i;
@@ -160,8 +163,13 @@ sub _parse_env {
     $headers->header($name => $value);
 
     # Host/Port
-    $value =~ s/:(\d+)$// ? $base->host($value)->port($1) : $base->host($value) if $name eq 'HOST';
+    if ($name eq 'HOST') {
+      $value =~ s/:(\d+)$//;
+      $base->host($value);
+      $port //= $1;
+    }
   }
+  $base->port($port) if $port;
 
   # Content-Type is a special case on some servers
   $headers->content_type($env->{CONTENT_TYPE}) if $env->{CONTENT_TYPE};
--- a/t/mojo/request_cgi.t
+++ b/t/mojo/request_cgi.t
@@ -291,6 +291,38 @@ subtest 'Parse Apache 2.2.14 CGI environment variables and body (root)' => sub {
   is $req->url->to_abs->to_string, 'http://127.0.0.1:13028/upload/', 'right absolute URL';
 };
 
+subtest 'SERVER_PORT (spec) superior to HTTP_HOST (not-spec)' => sub {
+  my $req = Mojo::Message::Request->new;
+  $req->parse({
+    SCRIPT_NAME       => '/upload',
+    SERVER_NAME       => '127.0.0.1',
+    SERVER_ADMIN      => '[no address given]',
+    PATH_INFO         => '/upload',
+    HTTP_CONNECTION   => 'Keep-Alive',
+    REQUEST_METHOD    => 'POST',
+    CONTENT_LENGTH    => '11',
+    SCRIPT_FILENAME   => '/tmp/SnLu1cQ3t2/test.fcgi',
+    SERVER_SOFTWARE   => 'Apache/2.2.14 (Unix) mod_fastcgi/2.4.2',
+    QUERY_STRING      => '',
+    REMOTE_PORT       => '58232',
+    HTTP_USER_AGENT   => 'Mojolicious (Perl)',
+    SERVER_PORT       => '13042',                                              # NOTE: When this line is present
+    SERVER_SIGNATURE  => '',
+    REMOTE_ADDR       => '127.0.0.1',
+    CONTENT_TYPE      => 'application/x-www-form-urlencoded; charset=UTF-8',
+    SERVER_PROTOCOL   => 'HTTP/1.1',
+    REQUEST_URI       => '/upload',
+    GATEWAY_INTERFACE => 'CGI/1.1',
+    SERVER_ADDR       => '127.0.0.1',
+    DOCUMENT_ROOT     => '/tmp/SnLu1cQ3t2',
+    PATH_TRANSLATED   => '/tmp/test.fcgi/upload',
+    HTTP_HOST         => '127.0.0.1:13028'                                     # NOTE: Ignore this PORT.
+  });
+  $req->parse('hello=world');
+  is $req->url->base->port,        13042,                            'right base port';       # from SERVER_PORT
+  is $req->url->to_abs->to_string, 'http://127.0.0.1:13042/upload/', 'right absolute URL';    # from SERVER_PORT
+};
+
 subtest 'Parse Apache 2.2.11 CGI environment variables and body (HTTPS=ON)' => sub {
   my $req = Mojo::Message::Request->new;
   $req->parse({


lib/Mojo/Message/Request.pm
```diff
@@ -153,15 +153,23 @@ sub _parse_env {
...
+
   for my $name (keys %$env) {
     my $value = $env->{$name};
     next unless $name =~ s/^HTTP_//i;
     $name =~ y/_/-/;
     $headers->header($name => $value);
 
     # Host/Port
-    $value =~ s/:(\d+)$// ? $base->host($value)->port($1) : $base->host($value) if $name eq 'HOST';
+    if ($name eq 'HOST') {
```
rmzg, line 166: Would this be easier to understand as `if ( not defined $port and $name eq 'HOST' )`

EvanCarroll: @rmzg No, because that wouldn't set `$base->host` if `defined $port`

kraih:
Looks like this breaks the [PSGI](https://metacpan.org/pod/PSGI) spec:
> SERVER_NAME, SERVER_PORT: When combined with SCRIPT_NAME and PATH_INFO, these keys can be used to complete > the URL. Note, however, that HTTP_HOST, if present, should be used in preference to SERVER_NAME for reconstructing the > request URL. SERVER_NAME and SERVER_PORT MUST NOT be empty strings, and are always required.

kraih:
Our PSGI backend has many more users than the CGI one, so that spec will always get precedence.

kraih:
And before you start arguing, which i know you enjoy. I'll also mention that `SERVER_PORT` is already used for `$c->tx->local_port`, and it is merely URL reconstruction where `HTTP_HOST` gets used instead. And this behaviour is also consistent with our custom HTTP servers. And of course since `Host` is a very important header in the HTTP spec, that is very much spec compliant there too.

EvanCarroll:
You got me wrong, I don't care to argue. I dropped it for a year before you closed it and told me exactly what you wanted when I promptly did it with a one day turnaround.

That said, there is a problem here as we're reading the spec's differently. So I've asked [PSGI to refine this](https://www.python.org/dev/peps/pep-0333/#url-reconstruction). The Plack code follows the conventions in my patch (which is my own reading). But interestingly it breaks the [Python WSGI spec](https://www.python.org/dev/peps/pep-0333/#url-reconstruction)

jberger:
I think you meant to link to the PSGI issue you opened https://github.com/plack/psgi-specs/issues/45

t/mojo/request_cgi.t
```diff
@@ -291,6 +291,38 @@ subtest 'Parse Apache 2.2.14 CGI environment variables and body (root)' => sub {
   is $req->url->to_abs->to_string, 'http://127.0.0.1:13028/upload/', 'right absolute URL';
 };
 
+subtest 'SERVER_PORT (spec) superior to HTTP_HOST (not-spec)' => sub {
```
kraih, line 294: With what kind of configuration would Apache/2.2.14 have different values in `SERVER_PORT` and `HTTP_HOST`?

t/mojo/request_cgi.t
```diff
@@ -291,6 +291,38 @@ subtest 'Parse Apache 2.2.14 CGI environment variables and body (root)' => sub {
...
+    PATH_INFO         => '/upload',
+    HTTP_CONNECTION   => 'Keep-Alive',
+    REQUEST_METHOD    => 'POST',
+    CONTENT_LENGTH    => '11',
+    SCRIPT_FILENAME   => '/tmp/SnLu1cQ3t2/test.fcgi',
+    SERVER_SOFTWARE   => 'Apache/2.2.14 (Unix) mod_fastcgi/2.4.2',
+    QUERY_STRING      => '',
+    REMOTE_PORT       => '58232',
+    HTTP_USER_AGENT   => 'Mojolicious (Perl)',
+    SERVER_PORT       => '13042',                                              # NOTE: When this line is present
```
kraih, line 309: Inconsistent comment.

t/mojo/request_cgi.t
```diff
@@ -291,6 +291,38 @@ subtest 'Parse Apache 2.2.14 CGI environment variables and body (root)' => sub {
...
+    SERVER_SIGNATURE  => '',
+    REMOTE_ADDR       => '127.0.0.1',
+    CONTENT_TYPE      => 'application/x-www-form-urlencoded; charset=UTF-8',
+    SERVER_PROTOCOL   => 'HTTP/1.1',
+    REQUEST_URI       => '/upload',
+    GATEWAY_INTERFACE => 'CGI/1.1',
+    SERVER_ADDR       => '127.0.0.1',
+    DOCUMENT_ROOT     => '/tmp/SnLu1cQ3t2',
+    PATH_TRANSLATED   => '/tmp/test.fcgi/upload',
+    HTTP_HOST         => '127.0.0.1:13028'                                     # NOTE: Ignore this PORT.
```
kraih, line 319: Inconsistent comment.

t/mojo/request_cgi.t
```diff
@@ -291,6 +291,38 @@ subtest 'Parse Apache 2.2.14 CGI environment variables and body (root)' => sub {
...
+  });
+  $req->parse('hello=world');
+  is $req->url->base->port,        13042,                            'right base port';       # from SERVER_PORT
+  is $req->url->to_abs->to_string, 'http://127.0.0.1:13042/upload/', 'right absolute URL';    # from SERVER_PORT
```
kraih, line 323: Comments that make no sense.

kraih (changes requested):

kraih:
And to be clear again, we do not actually follow RFC 3875 strictly. Because historically many web servers have not done that either. So any changes regarding CGI will require real world test cases from actual web servers. It's much easier with PSGI, since that spec can be followed strictly.

miyagawa:
> The Plack code follows the conventions in my patch (which is my own reading). But interestingly it breaks the Python WSGI spec. 

It does not conflict with the Python WSGI spec. The amount of parenthesis in the code might have confused you. See my comment https://github.com/plack/psgi-specs/issues/45#issuecomment-778816944

> That is to say there is clearly legitimate confusion here.

I'm happy to refine the spec to clarify that part, though.

kraih:
That settles it, `HTTP_HOST` has precedence for URL reconstruction and this PR is incorrect.

t/mojo/request_cgi.t
EvanCarroll: SetEnv SERVER_PORT 8080
SetEnv HTTP_HOST "http://google.com:80"

But I'm not using Apache, so not sure. Using a different webserver.

t/mojo/request_cgi.t
EvanCarroll: Are you asking that I reorder or these lines so they're closer to each other or that they stand on their own rather than in relationship?

EvanCarroll:
In agreement with @kraih . One question though, because essentially what's happening here is that the PSGI/WSGI spec is in compatible with the rules of CGI, should we (a) provide two different mechanisms one for CGI, one for PSGI -- or at the least document that this breaks with the CGI spec if HTTP_HOST is set?

kraih:
No, separate CGI code is not an option. If anything we might deprecate and remove CGI support if it becomes a maintenance burden.

kraih:
Thanks for your effort, but since this PR conflicts with PSGI it cannot be applied.

Issue #1706, opened 2021-02-16
Trusted proxies feature fails on Windows


kraih:
* Mojolicious version: master
* Perl version: all
* Operating system: Windows

### Steps to reproduce the behavior
Our CI tests. https://github.com/mojolicious/mojo/runs/1912645557

### Expected behavior
Pass.

### Actual behavior
```
t\mojo\cache.t ............................. ok
    #   Failed test 'right "Content-Length" value'
    #   at t\mojo\cgi.t line 202.
    #          got: '15'
    #     expected: '17'
    #   Failed test 'right content'
    #   at t\mojo\cgi.t line 204.
    #          got: '127.0.0.1:https'
    #     expected: '10.10.10.10:https'
    # Looks like you failed 2 tests of 6.

#   Failed test 'Trusted proxies'
#   at t\mojo\cgi.t line 205.
# Looks like you failed 1 test of 9.
t\mojo\cgi.t ............................... 
Dubious, test returned 1 (wstat 256, 0x100)
Failed 1/9 subtests 
t\mojo\collection.t ........................ ok
...
t\mojo\proxy.t ............................. ok

    #   Failed test 'right "Content-Length" value'
    #   at t\mojo\psgi.t line 271.
    #          got: '15'
    #     expected: '17'

    #   Failed test 'right content'
    #   at t\mojo\psgi.t line 275.
    #          got: '127.0.0.1:https'
    #     expected: '10.10.10.10:https'
    # Looks like you failed 2 tests of 6.

#   Failed test 'Trusted proxies'
#   at t\mojo\psgi.t line 279.
# Looks like you failed 1 test of 9.
t\mojo\psgi.t .............................. 
Dubious, test returned 1 (wstat 256, 0x100)
Failed 1/9 subtests 
t\mojo\reactor_detect.t .................... ok
...
t\mojo\user_agent_unix.t ................... skipped: set TEST_UNIX to enable this test (developer only!)

    #   Failed test 'v4 network contains addresss'
    #   at t\mojo\util.t line 495.

    #   Failed test 'v4 network contains addresss'
    #   at t\mojo\util.t line 496.
    # Looks like you failed 2 tests of 62.

#   Failed test 'network_contains'
#   at t\mojo\util.t line 552.
# Looks like you failed 1 test of 61.
t\mojo\util.t .............................. 
Dubious, test returned 1 (wstat 256, 0x100)
Failed 1/61 subtests 
t\mojo\websocket.t ......................... ok
...
[...]

PR #1707, opened 2021-02-16
Disallow short-form ipv4 networks

jberger:
As seen in https://github.com/mojolicious/mojo/runs/1912645557
apparently a change to Socket on windows meant that these addresses were
no longer parsing correctly. While they're convenient they're not
required. This change no longer allows them on any platform.


jberger:
This addresses #1706

kraih (approved):

Grinnz (approved):
LGTM

PR #1709, opened 2021-02-18
Issue 1708 return reject from finally

FGasper:
### Summary
Fix a nonstandard behaviour of Mojo::Promise::finally()

### Motivation
The library advertises compatibility with ES6 promises; this fixes one area of incompatibility.

NOTE: This is a breaking change; there was previously a test (now commented out) that verified ES6-incompatible behaviour regarding finally().

### DIFF
--- a/lib/Mojo/Promise.pm
+++ b/lib/Mojo/Promise.pm
@@ -56,7 +56,20 @@ sub catch { shift->then(undef, shift) }
 
 sub clone { $_[0]->new->ioloop($_[0]->ioloop) }
 
-sub finally { shift->_finally(1, @_) }
+sub finally {
+  my $class = ref $_[0];
+  my $cb    = $_[1];
+
+  my $big_cb = sub {
+    my @vals = @_;
+
+    $class->resolve($cb->())->then(sub {@vals});
+  };
+
+  $_[0]->catch($big_cb);
+
+  return $_[0]->then($big_cb);
+}
 
 sub map {
   my ($class, $options, $cb, @items) = (shift, ref $_[0] eq 'HASH' ? shift : {}, @_);
--- a/t/mojo/promise.t
+++ b/t/mojo/promise.t
@@ -160,15 +160,20 @@ subtest 'Double finally' => sub {
   is_deeply \@results, ['finally1', 'finally2'], 'promise not resolved';
 };
 
-subtest 'Resolved nested with finally' => sub {
-  my $promise  = Mojo::Promise->new;
-  my $promise2 = Mojo::Promise->new;
-  my @results;
-  $promise->finally(sub {$promise2})->finally(sub { @results = ('finally') });
-  $promise->resolve('pass');
-  Mojo::IOLoop->one_tick;
-  is_deeply \@results, ['finally'], 'promise already resolved';
-};
+# The expected behaviour here mismatches ES6: the 2nd finally() can’t
+# settle until the first one does, which in turn can’t settle until
+# $promise2 does. Since $promise2 doesn’t settle here, the 2nd finally()
+# should NOT fire.
+#
+#subtest 'Resolved nested with finally' => sub {
+#  my $promise  = Mojo::Promise->new;
+#  my $promise2 = Mojo::Promise->new;
+#  my @results;
+#  $promise->finally(sub {$promise2})->finally(sub { @results = ('finally') });
+#  $promise->resolve('pass');
+#  Mojo::IOLoop->one_tick;
+#  is_deeply \@results, ['finally'], 'promise already resolved';
+#};
 
 subtest 'Exception in finally' => sub {
   my $promise = Mojo::Promise->new;
@@ -514,6 +519,14 @@ subtest 'Wait for stopped loop' => sub {
   });
   $promise->then(sub { @results = @_ })->wait;
   is_deeply \@results, ['wait'], 'promise resolved';
+};
+
+subtest 'rejection returned from finally() callback' => sub {
+  my @fate;
+  Mojo::Promise->resolve(123)->finally(sub { Mojo::Promise->reject(666) })
+    ->then(sub { @fate = (0, @_) }, sub { @fate = (1, @_) },)->wait();
+
+  is_deeply(\@fate, [1, 666], 'reject returned from finally propagates');
 };
 
 done_testing();


kraih:
This PR is very broken, so we'll consider it a work in progress. Please make sure to inform us when it's ready for review.

FGasper:
@kraih Tests pass.

kraih:
@FGasper That's not the same as ready for review. Is it ready in your opinion? I still see a lot of style inconsistencies.

t/mojo/promise.t
```diff
@@ -160,15 +160,20 @@ subtest 'Double finally' => sub {
...
+#
+#subtest 'Resolved nested with finally' => sub {
+#  my $promise  = Mojo::Promise->new;
+#  my $promise2 = Mojo::Promise->new;
+#  my @results;
+#  $promise->finally(sub {$promise2})->finally(sub { @results = ('finally') });
+#  $promise->resolve('pass');
+#  Mojo::IOLoop->one_tick;
+#  is_deeply \@results, ['finally'], 'promise already resolved';
+#};
```
kraih, line 176: We never keep dead code around. That's a blocker. If you want to discuss this put it in the PR description (and perhaps explain it in the commit message).

kraih: Also, when discussing specs it helps a lot to link to/quote the section you are talking about. JavaScript examples also simplify this significantly.

lib/Mojo/Promise.pm
```diff
@@ -56,7 +56,20 @@ sub catch { shift->then(undef, shift) }
 
 sub clone { $_[0]->new->ioloop($_[0]->ioloop) }
 
-sub finally { shift->_finally(1, @_) }
+sub finally {
+  my $class = ref $_[0];
+  my $cb    = $_[1];
```
kraih, line 61: We don't unroll `@_`like that.

lib/Mojo/Promise.pm
```diff
@@ -56,7 +56,20 @@ sub catch { shift->then(undef, shift) }
 
 sub clone { $_[0]->new->ioloop($_[0]->ioloop) }
 
-sub finally { shift->_finally(1, @_) }
```
kraih, line 59: The `_finally(1,` case it not used at all anymore, so you are leaving dead code behind.

lib/Mojo/Promise.pm
```diff
@@ -56,7 +56,20 @@ sub catch { shift->then(undef, shift) }
...
+sub finally {
+  my $class = ref $_[0];
+  my $cb    = $_[1];
+
+  my $big_cb = sub {
```
kraih, line 63: `$big_cb`? What does that mean?

lib/Mojo/Promise.pm
```diff
@@ -56,7 +56,20 @@ sub catch { shift->then(undef, shift) }
...
+    my @vals = @_;
```
kraih, line 64: `@vals` is inconsistent.

lib/Mojo/Promise.pm
```diff
@@ -56,7 +56,20 @@ sub catch { shift->then(undef, shift) }
...
+
+    $class->resolve($cb->())->then(sub {@vals});
+  };
+
+  $_[0]->catch($big_cb);
```
kraih, line 69: `$_[0]` use for no apparent reason.

lib/Mojo/Promise.pm
```diff
@@ -56,7 +56,20 @@ sub catch { shift->then(undef, shift) }
...
+
+  return $_[0]->then($big_cb);
```
kraih, line 71: `$_[0]` use for no apparent reason again.

lib/Mojo/Promise.pm
```diff
@@ -56,7 +56,20 @@ sub catch { shift->then(undef, shift) }
...
 
-sub finally { shift->_finally(1, @_) }
+sub finally {
+  my $class = ref $_[0];
+  my $cb    = $_[1];
+
+  my $big_cb = sub {
+    my @vals = @_;
+
+    $class->resolve($cb->())->then(sub {@vals});
```
kraih, line 66: Pretty sure `$class->resolve` does not consider the current `ioloop` being used.

t/mojo/promise.t
```diff
@@ -516,4 +521,12 @@ subtest 'Wait for stopped loop' => sub {
   is_deeply \@results, ['wait'], 'promise resolved';
 };
 
+subtest 'rejection returned from finally() callback' => sub {
```
kraih, line 524: Lowercase and inconsistent description.

t/mojo/promise.t
```diff
@@ -516,4 +521,12 @@ subtest 'Wait for stopped loop' => sub {
...
+  my @fate;
```
kraih, line 525: `@fate` is inconsistent.

t/mojo/promise.t
```diff
@@ -516,4 +521,12 @@ subtest 'Wait for stopped loop' => sub {
...
+  Mojo::Promise->resolve(123)->finally(sub { Mojo::Promise->reject(666) })
```
kraih, line 526: Why `666`? If it's a joke i don't get it.

kraih (changes requested):

FGasper:
> @FGasper That's not the same as ready for review. Is it ready in your opinion? I still see a lot of style inconsistencies.

@kraih Sorry, I wasn’t sure what all is expected prior to review. I’ll address your style notes.

As to the PR’s readiness (style points notwithstanding): are you OK with the behaviour change? That seems to me the most pressing question.

lib/Mojo/Promise.pm
FGasper: Are you able to suggest a correction here? I don’t know what else would be the proper syntax; I’m looking for the equivalent of `Promise.resolve(…)`, which I thought was Mojo::Promise->resolve().

kraih:
> As to the PR’s readiness (style points notwithstanding): are you OK with the behaviour change? That seems to me the most pressing question.

I don't know yet. You've provided very little references to the specs. Those will have to be reviewed first before we can actually discuss behaviour changes.

kraih:
I do however suspect that you have made false assumptions here regarding the spec. While error handling is missing i don't believe returned promises need to be fulfilled to continue the chain.

FGasper:
> i don't believe returned promises need to be fulfilled to continue the chain.

```
var res, rej;
var p = new Promise( (y,n) => { res = y; rej = n } );
Promise.resolve(123).finally(() => { return p }).finally( () => console.log(999) )
```
^^^ `999` will only print once you resolve `p`.

kraih:
And please be aware that MDN is not an actual spec, it simplifies explanations for the user. For breaking changes we'll need very specific spec sections to justify the change.

lib/Mojo/Promise.pm
kraih: `Mojo::Promise` instances unfortunately carry an `ioloop` property around, that needs to be inherited by all chained promises.

kraih: Very unlike JavaScript promises, in Mojolicious applications there can be multiple event loops active at the same time.

kraih: And note that i don't believe that's a good idea, in fact i voted against it... but now it just is what it is and we have to deal with it.

kraih:
> ```
> var res, rej;
> var p = new Promise( (y,n) => { res = y; rej = n } );
> Promise.resolve(123).finally(() => { return p }).finally( () => console.log(999) )
> ```
> 
> ^^^ `999` will only print once you settle `p`.

You are indeed correct. I'm having a hard time finding a description for the behaviour in the spec however.

FGasper:
> I'm having a hard time finding a description for the behaviour in the spec however.

It’s here:

4. Let result be ? Call(onFinally, undefined).
5. Let C be F.[[Constructor]].
6. Assert: IsConstructor(C) is true.
7. Let promise be ? PromiseResolve(C, result).
8. Let thrower be equivalent to a function that throws reason.
9. Return ? Invoke(promise, "then", « thrower »).

So, `result` is a promise that rejects. That means promise (step 7) also rejects. That means the `thrower` is never called, because promise.then(throwser) is calling then() on an already-rejected promise.

lib/Mojo/Promise.pm
kraih: Not even sure if `_finally` is even needed anymore.

kraih:
Think i've fixed it with a rewrite of `_finally`. https://github.com/mojolicious/mojo/commit/ce3ad3300960b353c2f5e6c53e331797b5af616e

FGasper:
@kraih Cool … thanks!

kraih:
Detecting unhandled rejected promises does make everything a but clunky unfortunately.

kraih:
Thanks for investigating the issue! Your information was very helpful.

PR #1713, opened 2021-02-22
reuse $stash variable in dispatch

tim-2:
### Summary
Just a small change to reuse $stash variable in dispatch() I found while reading through the code.
I hope it is not too annoying for the mojo team to submit a PR for this simple change.


jhthorsen (approved):
I’m not sure if this makes any real difference, but I don’t mind merging it either.

jberger (approved):
My understanding of how closures work implies that we're already closing over $stash because we're closing over other values in the same pad. In which case we may as well save a method call, sure.

PR #1714, opened 2021-02-22
Support MOJO_LOG_LEVEL in Mojo::Log

pmb0:
### Summary

This pull request introduces support for the `MOJO_LOG_LEVEL` environment variable in `Mojo::Log`.

### Motivation

The variable is currently only supported in the context of `Mojolicious` or `Mojolicious::Lite`. However, `Mojo::Log` can also be used independently and therefore support for `MOJO_LOG_LEVEL` was previously missing.

### References

### DIFF
--- a/t/mojolicious/log_lite_app.t
+++ b/t/mojolicious/log_lite_app.t
@@ -7,6 +7,9 @@ use Mojolicious::Lite;
 use Mojo::Log;
 use Test::Mojo;
 
+# Test::Mojo sets MOJO_LOG_LEVEL to fatal
+local $ENV{MOJO_LOG_LEVEL} = 'debug';
+
 hook before_dispatch => sub {
   my $c = shift;
   $c->req->request_id('17a60115');
--- a/t/mojolicious/log_lite_app.t
+++ b/t/mojolicious/log_lite_app.t
@@ -7,9 +7,6 @@ use Mojolicious::Lite;
 use Mojo::Log;
 use Test::Mojo;
 
-# Test::Mojo sets MOJO_LOG_LEVEL to fatal
-local $ENV{MOJO_LOG_LEVEL} = 'debug';
-
 hook before_dispatch => sub {
   my $c = shift;
   $c->req->request_id('17a60115');
@@ -29,7 +26,7 @@ my $t = Test::Mojo->new;
 # Simple log messages with and without context
 my $buffer = '';
 open my $handle, '>', \$buffer;
-$t->app->log(Mojo::Log->new(handle => $handle));
+$t->app->log(Mojo::Log->new(handle => $handle, level => 'debug'));
 $t->get_ok('/simple')->status_is(200)->content_is('Simple!');
 like $buffer, qr/First.*Second.*Third.*No context!.*Fourth.*Fifth/s,    'right order';
 like $buffer, qr/\[.+\] \[\d+\] \[debug\] \[17a60115\] First!/,         'message with request id';
--- a/lib/Mojo/Log.pm
+++ b/lib/Mojo/Log.pm
@@ -19,7 +19,7 @@ has handle => sub {
   return Mojo::File->new($path)->open('>>');
 };
 has history          => sub { [] };
-has level            => 'debug';
+has level            => sub { $ENV{MOJO_LOG_LEVEL} || 'debug' };
 has max_history_size => 10;
 has 'path';
 has short => sub { $ENV{MOJO_LOG_SHORT} };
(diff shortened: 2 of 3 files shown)


t/mojolicious/log_lite_app.t
```diff
@@ -7,6 +7,9 @@ use Mojolicious::Lite;
 use Mojo::Log;
 use Test::Mojo;
 
+# Test::Mojo sets MOJO_LOG_LEVEL to fatal
+local $ENV{MOJO_LOG_LEVEL} = 'debug';
```
kraih, line 11: Needing a workaround for the test that's supposed to actually cover logging seems like a very bad sign for this PR.

kraih:
I'm not opposed to this change, but it should work cleanly without the need for workarounds anywhere. Perhaps this would have been better as a breaking change in the 9.0 release.

pmb0:
As far as I can see, this is not a breaking change and no workarounds are necessary.

kraih:
```perl
# Test::Mojo sets MOJO_LOG_LEVEL to fatal
local $ENV{MOJO_LOG_LEVEL} = 'debug';
```
That's a workaround for the fact that `MOJO_LOG_LEVEL` used to be much less powerful. And it suggests to me that there will be other unintended side effects. With a more powerful `MOJO_LOG_LEVEL` we simply shouldn't rely on it for the `Test::Mojo` feature.

kraih:
This PR needs to clean up and streamline all uses of `MOJO_LOG_LEVEL` to be considered i'm afraid.

kraih:
Using two separate environment variables might have made sense actually. Not that i can think of good descriptive names though. 😁

pmb0:
I don‘t think more than one variable makes sense: my use case is to control a whole deployment‘s log level, regardless if logging used in the web application or in other parts or classes. Introducing another var feels like a workaround.

kraih:
It's very different scopes though, one is `Mojolicious` applications, the other all `Mojo::Log` users.

kraih:
This is more what i would expect from a version with only one environment variable:
```diff
diff --git a/lib/Mojo/Log.pm b/lib/Mojo/Log.pm
index 413109088..4a6f89f3f 100644
--- a/lib/Mojo/Log.pm
+++ b/lib/Mojo/Log.pm
@@ -19,7 +19,7 @@ has handle => sub {
   return Mojo::File->new($path)->open('>>');
 };
 has history          => sub { [] };
-has level            => 'debug';
+has level            => sub { $ENV{MOJO_LOG_LEVEL} || ($ENV{HARNESS_IS_VERBOSE} ? 'debug' : 'fatal') };
 has max_history_size => 10;
 has 'path';
 has short => sub { $ENV{MOJO_LOG_SHORT} };
diff --git a/lib/Mojolicious.pm b/lib/Mojolicious.pm
index cb9c7c56d..6ca730ede 100644
--- a/lib/Mojolicious.pm
+++ b/lib/Mojolicious.pm
@@ -29,8 +29,7 @@ has log              => sub {
 
   # Reduced log output outside of development mode
   my $log = Mojo::Log->new;
-  return $log->level($ENV{MOJO_LOG_LEVEL}) if $ENV{MOJO_LOG_LEVEL};
-  return $self->mode eq 'development' ? $log : $log->level('info');
+  return $ENV{MOJO_LOG_LEVEL} || ($self->mode eq 'development') ? $log : $log->level('info');
 };
 has 'max_request_size';
 has mode               => sub { $ENV{MOJO_MODE} || $ENV{PLACK_ENV} || 'development' };
diff --git a/lib/Test/Mojo.pm b/lib/Test/Mojo.pm
index b02c52620..472121d70 100644
--- a/lib/Test/Mojo.pm
+++ b/lib/Test/Mojo.pm
@@ -20,9 +20,6 @@ has handler => sub { \&_handler };
 has [qw(message success tx)];
 has ua => sub { Mojo::UserAgent->new(insecure => 1)->ioloop(Mojo::IOLoop->singleton) };
 
-# Silent or loud tests
-$ENV{MOJO_LOG_LEVEL} ||= $ENV{HARNESS_IS_VERBOSE} ? 'debug' : 'fatal';
-
 sub app {
   my ($self, $app) = @_;
   return $self->ua->server->app unless $app;
```

pmb0:
Ok — I can update my PR accordingly later today.

pmb0:
Hello @kraih, I tried your suggestion. Again, some tests failed and I failed to get the overall system to work. I have tested various constellations. If one test runs, the other fails. It is really tricky 😄

I have just pushed a small change, which is actually not a big change. It no longer sets `MOJO_LOG_LEVEL` in `mojolicious/log_lite_app.t`, but instead passes the level along with the overridden handler when creating the log object. This is of course the same effect in the end. I find it difficult to judge what the correct default level should be. Since it is executed in the context of `Test::Mojo` and the test does not run verbosely: Actually `fatal`. But so far `debug` was assumed. So I think it is appropriate to set the log level here to `debug`.

Anyway, I'm running out of ideas, and if you think these changes will break things, we'll just close this MR. Thanks for your time. 😎

kraih:
Yes, what i suggested is a breaking change and requires significant test adjustments too.

kraih:
We'll consider this PR a work in progress. Please make sure to let us know when it is ready for review again.

kraih:
Or close it of course if you don't want to make any more changes. As it stands i'm afraid the PR did not pass the vote. (Even though i'm be very much in favour of simplifying the `MOJO_LOG_LEVEL` handling in general)

Issue #1406, opened 2019-09-15
Lite-applications and config override


christopherraa:
* Mojolicious version: 8.24
* Perl version: 5.26.0
* Operating system: Ubuntu 18.04

### Steps to reproduce the behavior
I've compiled a short demonstration that illustrates the issue.

**myapp**
```perl
#!/usr/bin/env perl
use Mojolicious::Lite;

my $config = plugin 'Config';
my $foo    = $config->{foo};

get '/config' => sub { shift->render(text => $foo) };

app->start;
```

**myapp.conf**
```
{
  foo => 'baz',
}
```

**t/myapp.t**
```perl
use Mojo::Base -strict;
use Test::Mojo;
use Test::More;
use Mojo::File qw/path/;

my $t = Test::Mojo->new(path(__FILE__)->dirname->sibling('myapp'), { foo => 'bar' });

is($t->app->config->{foo}, 'bar');
$t->get_ok('/config')->content_is('bar');

done_testing;
```
### Expected behavior
The documentation for `Test::Mojo->new()` has the following example:
```perl
my $t = Test::Mojo->new(Mojo::File->new('/path/to/myapp.pl'), {foo => 'bar'});
```
It was then my expectation that the overridden configuration would be present immediately when loading the Config-plugin so that any code run during startup would use the correct configuration parameters.

### Actual behavior
Insted the overridden configuration is available only after the app has started, as demonstrated by the tests.

dotandimet:
@christopherraa - the difference between a Lite app and a full app here is that without a startup() method to wrap your code, it doesn't run when the app "starts", it run when the app is **loaded** (at compile time). So you are defining a route based on the value of 'foo' when the file was first parsed and compiled, before either `new` or `startup` methods get called.

You can use config overriding when testing a lite app, but only if you make sure that config lookup happens after the app is started, not when it is loaded. So the example above can be re-written like this:

```perl
get '/config' => sub { my $c = shift; $c->render(text => $c->config('foo')) };
```

kraih:
I welcome proposals for improving the config override feature to work better with lite apps.

kraih:
Sadly, still no proposals.

kraih:
Fixed. 😀

christopherraa:
Awesome! This is great. :smiley:

PR #1734, opened 2021-03-10
Make config overrides work with Mojolicious::Lite

kraih:
This is a solution for #1406.


lib/Mojolicious.pm
```diff
@@ -57,6 +57,8 @@ has validator => sub { Mojolicious::Validator->new };
 our $CODENAME = 'Waffle';
 our $VERSION  = '9.07';
 
+our @ARGS_OVERRIDE;
```
kraih, line 60: This is pretty ugly, and i would welcome suggestions for a cleaner solution.

christopherraa (commented):
The overall change does look good I think. It seems like a pretty minimal change and solves the problem at hand in a good way. Consider this an approval, but I am holding off on actually approving until the thread asking for a better solution has been resolved (so as to avoid auto-merging prematurely). If you feel that the solution is OK as it is then I'll add my approval.

kraih:
To prevent auto-merging a PR can also just be labeled work in progress.

kraih:
My main issue was bad encapsulation with the global variable, `Mojo::Server` had to require `Mojolicious` to be able to set `@Mojolicious::ARGS_OVERRIDE`. Now i've reversed that and the global variable is `@Mojo::Server::ARGS_OVERRIDE`. That is at least a little better.

christopherraa (approved):
I am now :+1: to this. Agree that it much better with the args-move.

marcusramberg (approved):
Looks pretty clean now, and useful feature

PR #1747, opened 2021-03-24
Import fewer symbols in morbo.t

oalders:
Import fewer symbols in morbo.t

With this change, the tests should continue to pass, but the number of
symbols found goes from 612 to 444 according to:

```perl
use Symbol::Get ();
my @symbols = Symbol::Get::get_names();
diag 'found ' . @symbols . ' symbols';
```

### Motivation

As far as I can tell, there's no need to:

```perl
use IO::Socket::INET;
use Socket qw(SO_REUSEPORT SOL_SOCKET);
```

`IO::Socket::INET` isa `IO::Socket` and does not have its own `import()`. If
`IO::Socket::import()` is called without any arguments, it exports
directly from `Socket`

https://metacpan.org/source/TODDR/IO-1.45/lib/IO/Socket.pm#L38

`Socket` exports a large number of symbols by default, including
`SO_REUSEPORT` and `SOL_SOCKET`. So, what appears to be happening is that
the `use Socket` statement avoids importing a lot of unneeded symbols,
but this is actually undone by the bare import of `IO::Socket::INET`.

### DIFF
--- a/t/mojo/morbo.t
+++ b/t/mojo/morbo.t
@@ -9,7 +9,7 @@ plan skip_all => 'set TEST_MORBO to enable this test (developer only!)' unless $
 use Mojo::File qw(curfile);
 use lib curfile->sibling('lib')->to_string;
 
-use IO::Socket::INET;
+use IO::Socket::INET ();
 use Mojo::File qw(tempdir);
 use Mojo::IOLoop::Server;
 use Mojo::Server::Morbo::Backend;


Grinnz:
IO::Socket::INET does not document such an export interface. I would prefer continuing to import these values from Socket, and simply prevent importing from IO::Socket::INET.

Grinnz:
Also the PR title needs to be fixed.

oalders:
I created the PR using `hub pull-request`. I guess it doesn't like the template.

oalders:
force-pushed and updated the uninformative PR name.

kraih (approved):

Grinnz (approved):

Issue #1694, opened 2021-02-07
Helper process support for Mojo::Server::Daemon/Prefork


kraih:
The motivation for this is https://github.com/mojolicious/minion/issues/76. But there are many other use cases for running arbitrary code in a persistent helper process next to the Mojolicious application server.

While forking such a process from a `before_server_start` hook is fairly trivial, making it work with `Mojo::Server::Daemon` as well as `Mojo::Server::Prefork` is not. Especially stopping the process together with the main application server process is tricky. It will have to work with `Mojo::Server::Morbo` and `Mojo::Server::Hypnotoad` too. There also needs to be some way to handle failures in the helper process (at the very least logging the exit code/signal as an error).

API wise, we don't need anything fancy, a simple plugin should do just fine for now.
```
plugin HelperProcess => {run => sub ($app) {
  while (1) {
    sleep 1;
    $app->log->debug("Helper process $$ is still running");
  }
}};
```
Process management is the main priority and needs to be rock solid. With very good test coverage.

kraih:
There will probably have to be two completely separate code paths in the plugin for daemon and prefork. Since both are so different.

kraih:
I've been thinking a bit more about this feature, and actually most real deployments are just `prefork` these days. And `prefork` already has process management features (for the workers) we could expand to include helper processes. The only real blocker here is `morbo`, which needs the feature for development and uses `daemon`. Perhaps we could switch `morbo` to `prefork` and sidestep the problem that way?

jberger:
I still think there's a place for hypnotoad in the modern stack because k8s can update configuration but your app must know hot to respond to it. I can see a blend of hypnotoad and morbo hot restart on config changes so you don't have to lose connections with `kubectl rollout restart`. Here was my first attempt at it a few days ago https://gist.github.com/jberger/e7a5ef096e6637facdc196eea7a063d8

PR #1755, opened 2021-04-04
Add a 'worker_starting' event.

rmanfredi:
This event (`before_app_start`) is always called before the application starts, regardless
of whether it is invoked via `morbo`, is started in daemon mode from the
command line or via `hypnotoad` in pre-forking mode.

It avoids having to say:

    Mojo::IOLoop->next_tick(sub ($ioloop) {
        app->log->info("Worker $$ star...ALL GLORY TO THE HYPNOTOAD!");
    });

to get this special behaviour.  It is way nicer to be able to say:

    app->hook(before_app_start => sub ($app) {
        $app->log->info("App in $$ star...ALL GLORY TO MOJOLICIOUS!");
    });

to get that behaviour.

It is more consistent with the over hooks, and also it is a better
abstraction than having to know the internals of the IOLoop to be
able to get such a hooking point in pre-forked processes.

Now the application has a unique and consistent entry point where it can
initialize its global state.

### DIFF
--- a/lib/Mojo/Server/CGI.pm
+++ b/lib/Mojo/Server/CGI.pm
@@ -3,6 +3,14 @@ use Mojo::Base 'Mojo::Server';
 
 has 'nph';
 
+sub new {
+  my $class = shift;
+  my $self  = $class->SUPER::new(@_);
+  my $app   = $self->app;
+  $app->app_starting if ref $app;
+  return $self;
+}
+
 sub run {
   my $self = shift;
 
@@ -117,6 +125,14 @@ Activate non-parsed header mode.
 =head1 METHODS
 
 L<Mojo::Server::CGI> inherits all methods from L<Mojo::Server> and implements the following new ones.
+
+=head2 new
+
+  my $app    = Mojolicious->new;
+  my $server = Mojo::Server::CGI->new(app => $app);
+  my $server = Mojo::Server::CGI->new(app => $app, nph => 1);
+
+Creates a C<Mojo::Server::CGI> object.
 
 =head2 run
 
--- a/lib/Mojo/Server/Daemon.pm
+++ b/lib/Mojo/Server/Daemon.pm
@@ -35,6 +35,7 @@ sub run {
   my $loop = $self->ioloop;
   my $int  = $loop->recurring(1 => sub { });
   local $SIG{INT} = local $SIG{TERM} = sub { $loop->stop };
+  $self->app->app_starting;
   $self->start->ioloop->start;
   $loop->remove($int);
 }
@@ -53,6 +54,7 @@ sub start {
   # Start listening
   elsif (!@{$self->acceptors}) {
     $self->app->server($self);
+    $self->app->app_starting;
     $self->_listen($_) for @{$self->listen};
   }
 
--- a/lib/Mojolicious.pm
+++ b/lib/Mojolicious.pm
@@ -175,6 +175,8 @@ sub plugin {
 
 sub server { $_[0]->plugins->emit_hook(before_server_start => @_[1, 0]) }
 
+sub app_starting { $_[0]->plugins->emit_hook(before_app_start => $_[0]) }
+
 sub start {
   my $self = shift;
   $_->warmup for $self->static, $self->renderer;
@@ -259,6 +261,17 @@ built-in ones (except for L<Mojo::Server::CGI>).
 
 Useful for reconfiguring application servers dynamically or collecting server diagnostics information. (Passed the
 server and application objects)
+
+=head2 before_app_start
+
+Emitted right before the application starts.  It works with all the built-in web
+servers.  In the case of L<Mojo::Server::Prefork>, it will run in each spawned
+worker, not in the master server.
+
+  $app->hook(before_app_start => sub ($app) {...});
+
+This can be used to initialize the global state that the application will be
+perusing during its processing. (Passed the application object)
 
 =head2 after_build_tx
 
@@ -723,6 +736,12 @@ subclass.
   $app->warmup;
 
 Preload classes from L</"preload_namespaces"> for future use.
+
+=head2 app_starting
+
+  $app->app_starting;
+
+Emits the L</"before_app_start"> hook.
 
 =head1 HELPERS
 
--- a/lib/Mojolicious.pm
+++ b/lib/Mojolicious.pm
@@ -175,7 +175,15 @@ sub plugin {
 
 sub server { $_[0]->plugins->emit_hook(before_server_start => @_[1, 0]) }
 
-sub app_starting { $_[0]->plugins->emit_hook(before_app_start => $_[0]) }
+sub app_starting {
+  my $self = shift;
+
+  # Ensure we run 'before_app_start' only once per process
+  $self->{_started} = {} unless defined $self->{_started};
+  return if exists $self->{_started}->{$$};
+  $self->{_started}->{$$} = undef;
+  $self->plugins->emit_hook(before_app_start => $self);
+}
 
 sub start {
   my $self = shift;
--- a/lib/Mojolicious.pm
+++ b/lib/Mojolicious.pm
@@ -272,14 +272,13 @@ server and application objects)
 
 =head2 before_app_start
 
-Emitted right before the application starts.  It works with all the built-in web
-servers.  In the case of L<Mojo::Server::Prefork>, it will run in each spawned
-worker, not in the master server.
+Emitted right before the application starts.  It works with all the built-in web servers.  In the case of
+L<Mojo::Server::Prefork>, it will run in each spawned worker, not in the master server.
 
   $app->hook(before_app_start => sub ($app) {...});
 
-This can be used to initialize the global state that the application will be
-perusing during its processing. (Passed the application object)
+This can be used to initialize the global state that the application will be perusing during its processing. (Passed
+the application object)
 
 =head2 after_build_tx
 
--- a/lib/Mojolicious.pm
+++ b/lib/Mojolicious.pm
@@ -179,7 +179,7 @@ sub app_starting {
   my $self = shift;
 
   # Ensure we run 'before_app_start' only once per process
-  $self->{_started} = {} unless defined $self->{_started};
+  $self->{_started} //= {};
   return if exists $self->{_started}->{$$};
   $self->{_started}->{$$} = undef;
   $self->plugins->emit_hook(before_app_start => $self);
--- a/lib/Mojolicious/Guides/Cookbook.pod
+++ b/lib/Mojolicious/Guides/Cookbook.pod
@@ -178,12 +178,10 @@ worker is allowed to handle (often as low as C<1>).
   $ ./script/my_app prefork -m production -w 10 -c 1
   Web application available at http://127.0.0.1:3000
 
-During startup your application is preloaded in the manager process
-but it will not start any processing there.
+During startup your application is preloaded in the manager process but it will not start any processing there.
 
-You can use the hook L<Mojolicious/"before_worker_start"> to run code in each
-child process (the worker processes that are being forked) before it is about
-to start processing incoming requests:
+You can use the hook L<Mojolicious/"before_worker_start"> to run code in each child process (the worker processes
+that are being forked) before it is about to start processing incoming requests:
 
   use Mojolicious::Lite;
 
@@ -195,7 +193,7 @@ to start processing incoming requests:
 
   app->start;
 
-That can be handy to setup global state once.
+That can be handy to setup global state, once.
 
 And to manage the pre-forking web server with systemd, you can use a unit configuration file like this.
 
--- a/lib/Mojolicious/Guides/Cookbook.pod
+++ b/lib/Mojolicious/Guides/Cookbook.pod
@@ -178,18 +178,24 @@ worker is allowed to handle (often as low as C<1>).
   $ ./script/my_app prefork -m production -w 10 -c 1
   Web application available at http://127.0.0.1:3000
 
-During startup your application is preloaded in the manager process, which does not run an event loop, so you can use
-L<Mojo::IOLoop/"next_tick"> to run code whenever a new worker process has been forked and its event loop gets started.
+During startup your application is preloaded in the manager process
+but it will not start any processing there.
+
+You can use the hook L<Mojolicious/"before_worker_start"> to run code in each
+child process (the worker processes that are being forked) before it is about
+to start processing incoming requests:
 
   use Mojolicious::Lite;
 
-  Mojo::IOLoop->next_tick(sub ($ioloop) {
-    app->log->info("Worker $$ star...ALL GLORY TO THE HYPNOTOAD!");
+  app->hook(before_worker_start => sub ($app) {
+    $app->log->info("Worker $$ star...ALL GLORY TO THE HYPNOTOAD!");
   });
 
   get '/' => {text => 'Hello Wor...ALL GLORY TO THE HYPNOTOAD!'};
 
   app->start;
+
+That can be handy to setup global state once.
 
 And to manage the pre-forking web server with systemd, you can use a unit configuration file like this.
 
--- a/t/mojo/cgi.t
+++ b/t/mojo/cgi.t
@@ -9,7 +9,13 @@ use Mojolicious::Lite;
 # Silence
 app->log->level('fatal');
 
-get '/' => {text => 'Your Mojo is working!'};
+my $working = "NOT working";
+app->hook(before_app_start => sub { $working = "working" });
+
+get '/' => sub {
+  my $c = shift;
+  $c->render(text => "Your Mojo is $working!");
+};
 
 post '/chunked' => sub {
   my $c = shift;
--- a/t/mojo/daemon.t
+++ b/t/mojo/daemon.t
@@ -112,8 +112,8 @@ my $ua  = Mojo::UserAgent->new(ioloop => Mojo::IOLoop->singleton);
 my $whatever = 'Failed!';
 $app->hook(
   before_app_start => sub {
-    my ($app) = shift;
-    my $ok = ref $app && $app->isa('Mojolicious');
+    my $app = shift;
+    my $ok  = ref $app && $app->isa('Mojolicious');
     $whatever = $ok ? 'Whatever!' : 'Wrong type';
   }
 );
--- a/t/mojo/daemon.t
+++ b/t/mojo/daemon.t
@@ -109,6 +109,15 @@ subtest 'Transaction' => sub {
 my $app = Mojolicious->new;
 my $ua  = Mojo::UserAgent->new(ioloop => Mojo::IOLoop->singleton);
 
+my $whatever = 'Failed!';
+$app->hook(
+  before_app_start => sub {
+    my ($app) = shift;
+    my $ok = ref $app && $app->isa('Mojolicious');
+    $whatever = $ok ? 'Whatever!' : 'Wrong type';
+  }
+);
+
 subtest 'Moniker' => sub {
   is $ua->server->app($app)->app->moniker, 'mojolicious', 'right moniker';
 };
@@ -161,7 +170,12 @@ $app->routes->any(
   }
 );
 
-$app->routes->any('/*whatever' => {text => 'Whatever!'});
+$app->routes->any(
+  '/*whatever' => sub {
+    my ($c) = shift;
+    $c->render(text => $whatever);
+  }
+);
 
 subtest 'Normal request' => sub {
   my $tx = $ua->get('/normal/');
--- a/t/mojo/morbo.t
+++ b/t/mojo/morbo.t
@@ -71,6 +71,7 @@ app->log->level('fatal');
 Mojo::IOLoop->next_tick(sub { path('$started')->touch });
 
 get '/hello' => {text => 'Hello World!'};
+get '/start' => {text => 'Started Morbo!'};
 
 app->start;
 EOF
@@ -90,6 +91,12 @@ EOF
   ok $tx->is_finished, 'transaction is finished';
   is $tx->res->code, 200,            'right status';
   is $tx->res->body, 'Hello World!', 'right content';
+
+  # Hook has been properly called
+  $tx = $ua->get("http://127.0.0.1:$port/started");
+  ok $tx->is_finished, 'transaction is finished';
+  is $tx->res->code, 200,              'right status';
+  is $tx->res->body, 'Started Morbo!', 'right content';
 };
 
 subtest 'Update script without changing mtime' => sub {
@@ -108,7 +115,11 @@ Mojo::IOLoop->next_tick(sub { path('$started')->touch });
 my \$message = 'Failed!';
 hook before_server_start => sub { \$message = 'Hello!' };
 
+my \$started = 'Failed!';
+hook before_app_start => sub { \$started = 'YES!' };
+
 get '/hello' => sub { shift->render(text => \$message) };
+get '/start' => sub { shift->render(text => \$started) };
 
 app->start;
 EOF
@@ -129,6 +140,12 @@ EOF
   ok $tx->is_finished, 'transaction is finished';
   is $tx->res->code, 200,      'right status';
   is $tx->res->body, 'Hello!', 'right content';
+
+  # This new one is there
+  $tx = $ua->get("http://127.0.0.1:$port/start");
+  ok $tx->is_finished, 'transaction is finished';
+  is $tx->res->code, 200,    'right status';
+  is $tx->res->body, 'YES!', 'right content';
 };
 
 subtest 'New file(s)' => sub {
--- a/t/mojo/morbo.t
+++ b/t/mojo/morbo.t
@@ -71,7 +71,6 @@ app->log->level('fatal');
 Mojo::IOLoop->next_tick(sub { path('$started')->touch });
 
 get '/hello' => {text => 'Hello World!'};
-get '/start' => {text => 'Started Morbo!'};
 
 app->start;
 EOF
@@ -91,12 +90,6 @@ EOF
   ok $tx->is_finished, 'transaction is finished';
   is $tx->res->code, 200,            'right status';
   is $tx->res->body, 'Hello World!', 'right content';
-
-  # Hook has been properly called
-  $tx = $ua->get("http://127.0.0.1:$port/started");
-  ok $tx->is_finished, 'transaction is finished';
-  is $tx->res->code, 200,              'right status';
-  is $tx->res->body, 'Started Morbo!', 'right content';
 };
 
 subtest 'Update script without changing mtime' => sub {
@@ -115,11 +108,7 @@ Mojo::IOLoop->next_tick(sub { path('$started')->touch });
 my \$message = 'Failed!';
 hook before_server_start => sub { \$message = 'Hello!' };
 
-my \$started = 'Failed!';
-hook before_app_start => sub { \$started = 'YES!' };
-
 get '/hello' => sub { shift->render(text => \$message) };
-get '/start' => sub { shift->render(text => \$started) };
 
 app->start;
 EOF
@@ -140,12 +129,6 @@ EOF
   ok $tx->is_finished, 'transaction is finished';
   is $tx->res->code, 200,      'right status';
   is $tx->res->body, 'Hello!', 'right content';
-
-  # This new one is there
-  $tx = $ua->get("http://127.0.0.1:$port/start");
-  ok $tx->is_finished, 'transaction is finished';
-  is $tx->res->code, 200,    'right status';
-  is $tx->res->body, 'YES!', 'right content';
 };
 
 subtest 'New file(s)' => sub {
--- a/t/mojo/prefork.t
+++ b/t/mojo/prefork.t
@@ -50,7 +50,7 @@ my $counter = Mojolicious->new;
 my ($counter_events, $counter_spawned, $counter_server);
 $counter->hook(
   before_app_start => sub {
-    my ($app) = @_;
+    my $app = shift;
     $counter_events  = {};
     $counter_spawned = {};
     $counter_server  = {};
@@ -146,7 +146,7 @@ $prefork->app->hook(
 );
 $prefork->app->hook(
   before_app_start => sub {
-    my ($app) = @_;
+    my $app = shift;
     Mojo::UserAgent->new->post("$durl}triggered/$$");
   }
 );
@@ -211,7 +211,7 @@ $prefork->app->hook(
 );
 $prefork->app->hook(
   before_app_start => sub {
-    my ($app) = @_;
+    my $app = shift;
     Mojo::UserAgent->new->post("$durl/triggered/$$");
   }
 );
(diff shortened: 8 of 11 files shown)


lib/Mojo/Server/Daemon.pm
```diff
@@ -35,6 +35,7 @@ sub run {
   my $loop = $self->ioloop;
   my $int  = $loop->recurring(1 => sub { });
   local $SIG{INT} = local $SIG{TERM} = sub { $loop->stop };
+  $self->app->worker_starting;
```
jhthorsen, line 38: Isn't this "before_worker_start" ? And I wonder if we need a method for this...

@kraih: Is it bad to emit events from outside the app? As in `$self->app->plugins->emit_hook(...)`

t/mojo/daemon.t
```diff
@@ -18,13 +18,21 @@ use Mojolicious;
 package TestApp;
 use Mojo::Base 'Mojolicious';
 
+my $message = 'Test';
+
 sub handler {
   my ($self, $tx) = @_;
   $tx->res->code(200);
-  $tx->res->body('Hello TestApp!');
+  $tx->res->body("Hello $message!");
```
jhthorsen, line 26: Why are the quotes changed?

t/mojo/daemon.t
```diff
@@ -18,13 +18,21 @@ use Mojolicious;
...
   $tx->resume;
 }
 
+# Ensure the 'startup' routine it called
+sub startup {
+	my $self = shift;
+	$message = 'TestApp';
```
jhthorsen: Not run through perltidy. Either way, I'm not sure if this is testing what you want.

t/mojo/daemon.t
```diff
@@ -161,6 +169,13 @@ $app->routes->any(
   }
 );
 
+my $whatever = 'Failed!';
+
+$app->hook(worker_starting => sub {
+  my ($app) = shift;
+  $whatever = 'Whatever!';
```
jhthorsen: Maybe test that the $app is actually the first argument?

t/mojo/morbo.t
```diff
@@ -90,6 +91,12 @@ EOF
   ok $tx->is_finished, 'transaction is finished';
   is $tx->res->code, 200,            'right status';
   is $tx->res->body, 'Hello World!', 'right content';
+
+  # This new one is there
```
jhthorsen, line 95: I don't understand this comment. Same below.

jhthorsen (commented):
I think I'm leaning towards thinking this might be a good addition since it is a question that comes up again and again. I don't like the event name though.

My review is indeed early, so I will come back to this PR when it has matured more.

lib/Mojo/Server/Daemon.pm
rmanfredi: Could rename the event as "before_worker_start" to be consistent with the other hooks, indeed!
I like it better myself.

t/mojo/daemon.t
rmanfredi: Because we need variable interpolation now!

t/mojo/daemon.t
rmanfredi: I adjusted indentation afterwards.
This is not checking the hook indeed, I can remove this test.

t/mojo/daemon.t
rmanfredi: Yes, will add a check for that.

t/mojo/morbo.t
rmanfredi: Ooops.  It was my fiddling with code I am discovering! Will remove.

t/mojo/daemon.t
rmanfredi: Anyway I'm going to remove this code as it is not needed.

rmanfredi:
Done with all the changes following the initial review.
Thank you @jhthorsen.

As of now, the hook is called "before_worker_start".  However, I'm offering a new suggestion: "before_app_start".

Indeed, this is not a hook just for pre-forked servers, it is a generic hook that applications can use to initialize global state before they start processing requests.  It just happens to "do the right thing" in pre-forked environments.

kraih:
Normally i'd be against this, since `Mojo::IOLoop->next_tick` works fine and has been the documented solution forever. But this also includes PSGI, so there might be enough original use cases here, but i'd really like to hear about at least two independent ones (same as we've done for every other hook before).

lib/Mojo/Server/Daemon.pm
kraih, line 38: This looks like the wrong place, not all `Mojo::Server::Daemon` instances are started with `run`.

lib/Mojolicious.pm
```diff
@@ -175,6 +175,10 @@ sub plugin {
 
 sub server { $_[0]->plugins->emit_hook(before_server_start => @_[1, 0]) }
 
+sub worker_starting {
+  $_[0]->plugins->emit_hook(before_worker_start => $_[0]);
+}
```
kraih, line 180: The consistent name would be `worker`. And there is no need to use more than one line.

lib/Mojolicious.pm
```diff
@@ -260,6 +264,19 @@ built-in ones (except for L<Mojo::Server::CGI>).
 Useful for reconfiguring application servers dynamically or collecting server diagnostics information. (Passed the
 server and application objects)
 
+=head2 before_worker_start
+
+Emitted right before the application starts, regardless of whether it
+is invoked via C<morbo>, via the command line in plain daemon mode or
+through pre-forking via C<hypnotoad>.
```
kraih, line 271: There is no need to list individual servers, the `before_server_start` hook doesn't do that either. Just follow its example. This hook is almost the same with different timing.

kraih: Also needs reformatting.

t/mojo/daemon.t
```diff
@@ -161,6 +161,16 @@ $app->routes->any(
...
+my $whatever = 'Failed!';
+
+$app->hook(
+  before_worker_start => sub {
+    my ($app) = shift;
+    my $ok = ref $app && $app->isa('Mojolicious');
+    $whatever = $ok ? 'Whatever!' : 'Wrong type';
+  }
+);
+
```
kraih, line 173: Where is the test?

t/mojo/morbo.t
```diff
@@ -71,6 +71,7 @@ app->log->level('fatal');
 Mojo::IOLoop->next_tick(sub { path('$started')->touch });
 
 get '/hello' => {text => 'Hello World!'};
+get '/start' => {text => 'Started Morbo!'};
```
kraih, line 74: There is nothing special about morbo, these tests don't need to exist if the hook was implemented propoerly in `Mojo::Server::Daemon`.

kraih: Instead what does need tests is prefork.

t/mojo/psgi.t
```diff
@@ -41,12 +41,19 @@ subtest 'Reverse proxy' => sub {
 
 subtest 'Binding' => sub {
   my @server;
+  my @started;
```
kraih, line 44: Combine the two lines.

lib/Mojolicious/Guides/Cookbook.pod
```diff
@@ -178,19 +178,29 @@ worker is allowed to handle (often as low as C<1>).
...
 
-During startup your application is preloaded in the manager process, which does not run an event loop, so you can use
-L<Mojo::IOLoop/"next_tick"> to run code whenever a new worker process has been forked and its event loop gets started.
+During startup your application is preloaded in the manager process
+but it will not start any processing there.
+
+You can use
+the hook L<Mojolicious/"before_worker_start"> to run code in each child
+process (the worker processes that are being forked) before it is about
+to start processing incoming requests:
```
kraih, line 187: Reformatting.

lib/Mojolicious/Guides/Cookbook.pod
```diff
@@ -178,19 +178,29 @@ worker is allowed to handle (often as low as C<1>).
...
   });
 
   get '/' => {text => 'Hello Wor...ALL GLORY TO THE HYPNOTOAD!'};
 
   app->start;
 
+That can be handy to setup global state once.  Note that the hook
+L<Mojolicious/"before_worker_start"> is B<not> specific to pre-forked
+processes: the hook is also called when running as a single daemon
+process, or during development when the application runs via C<morbo>.
```
kraih, line 202: This section is about prefork only.

lib/Mojo/Server/Daemon.pm
kraih: For example, this would not work for `Mojo::UserAgent::Server`, which all our tests rely on. That's a blocker for this PR i'm afraid.

rmanfredi: So what's the proper place for the hook to be fired?

lib/Mojolicious.pm
rmanfredi: Hmm... adding more entries to a class which can have many heirs... we need to be careful to not choose a routine name that is likely to be conflicting with any user code already written.

As for the one-liner, agreed, it fits on 80 chars, and I can omit the trailing ';' if we make it fit on a single line,

    $ echo 'sub worker_starting { $_[0]->plugins->emit_hook(before_worker_start => $_[0]); }' | wc -c
    81 # counts the trailing hidden "\n"

lib/Mojolicious.pm
rmanfredi: Reformatting?  how so?

(not to be picky, but it's not just the *timing*, it's also the *place* where it is run)

t/mojo/daemon.t
rmanfredi: Goog catch! :1st_place_medal:

rmanfredi: That is fixed now.  New commit takes care of Mojo::UserAgent::Server as well.

t/mojo/morbo.t
rmanfredi: You are right of course.  But should not we always make sure the proper semantics we want is indeed covered through all the different high-level use cases?
Sure, morbo relies on Mojo::Server::Daemon today, but what if, one day, this uses a new implementation and we forget about the semantic we want for the hook we are introducing here?

t/mojo/psgi.t
rmanfredi: Done.

lib/Mojolicious.pm
rmanfredi: Done.

lib/Mojolicious/Guides/Cookbook.pod
rmanfredi: Killed paragraph.

lib/Mojolicious/Guides/Cookbook.pod
rmanfredi: Done, reformatted to 80 columns.

lib/Mojo/Server/Daemon.pm
rmanfredi: @kraih it does work with Mojo::UserAgent::Server now.

rmanfredi:
I believe this hook should really be named "before_app_start".

lib/Mojolicious.pm
rmanfredi: Anyway, this paragraph is now gone.

s1037989:
> I believe this hook should really be named "before_app_start".

I like this name better.  My only comment is just thinking about how this PR will ensure that the hook will get called for each worker start, even those [re]starting later; and my very possibly flawed reaction to `before_app_start` would imply that this hook code will get run once and only once (thinking #1694) -- i.e. *before* the app starts and starts forking off workers, but that is likely a distinction between the executable "app" and the Mojolicious "app", where the definition of "app" should indeed favor the latter, but possibly confusing nonetheless.

Also, a possible contradiction: is it really `before_app_start` if it is executed *after* `app->start`?

rmanfredi:
> Also, a possible contradiction: is it really `before_app_start` if it is executed _after_ `app->start`?

Good point.  But `before_app_service` would be ugly, would not it be? I would also ditch `before_app_enters_ioloop` :smile: 

Yeah, naming is really hard.

Grinnz:
The name should reference that it is run on the start of each worker process, not the application. before_app_start would be misleading for both these reasons.

rmanfredi:
> The name should reference that it is run on the start of each worker process, not the application. before_app_start would be misleading for both these reasons.

I have made also patches to make sure the new hook is run before every possible application start I could see, so that behaviour can be controlled in a standard way: regardless of whether you are going to use pre-forking, or run in a single daemon, or via `morbo`, the aim is to have a hook to initialize global state that you would not want to do for **every** request you are going to serve...

Grinnz:
But it is for *every* worker, which makes it different from "before application start", which only occurs once.

rmanfredi:
> My only comment is just thinking about how this PR will ensure that the hook will get called for each worker start, even those [re]starting later.

That's the idea.  I can see that it works for my application, which already relies on the new hook here: the code gets triggered every time a new child is starting. My logging vouches for it :smile.

The trick is... how to write good tests for it.  That's why I haven't touched `t/mojo/prefork.t` yet.

rmanfredi:
> But it is for _every_ worker, which makes it different from "before application start", which only occurs once.

Yes, it is for **every** worker.  That's why I had named it `worker_starting` initially, and then realized that when I used `morbo` it did not get called at all and that meant having another hook to do the same thing: getting the local process setup for efficiently processing incoming requests.  So it's not just for pre-forked processes.

lib/Mojo/UserAgent/Server.pm
```diff
@@ -12,11 +12,16 @@ sub app {
 
   # Singleton application
   state $singleton;
-  return $singleton = $app ? $app : $singleton unless ref $self;
+  unless (ref $self) {
+    my $starting = ref $app && (!defined $singleton || $singleton ne $app);
+    $app->worker_starting if $starting;
+    return $singleton = $app ? $app : $singleton;
+  }
```
kraih, line 19: I do not want to see any changes in this module.

kraih (changes requested):

lib/Mojo/Server/CGI.pm
```diff
@@ -3,6 +3,14 @@ use Mojo::Base 'Mojo::Server';
 
 has 'nph';
 
+sub new {
+  my $class = shift;
+  my $self  = $class->SUPER::new(@_);
+  my $app   = $self->app;
+  $app->worker_starting if ref $app;
+  return $self;
+}
```
kraih, line 12: This is now inconsistent with `before_server_start`. Also a blocker for me.

kraih (changes requested):

lib/Mojo/Server/CGI.pm
rmanfredi: How is it inconsistent exactly?

lib/Mojo/UserAgent/Server.pm
rmanfredi: OK, fixed.

lib/Mojo/Server/CGI.pm
kraih: `Mojo::Server::CGI` is excluded from deployment hooks like `before_server_start`.

kraih: And this is also a deployment hook you are working on. It does not get emitted for commands like `./myapp.pl eval -V 'app->mode'` after all.

lib/Mojo/UserAgent/Server.pm
kraih: To be more precise, i want it to work for `Mojo::UserAgent::Server`, but without any changes to the actual module. The feature needs to stay contained to `Mojo::Server::Daemon`.

kraih: If there's no elegant solution, then `Mojo::IOLoo->next_tick` is simply better in my opinion.

t/mojo/morbo.t
kraih: No, morbo will always be a full reload of `Mojo::Server::Daemon`.

t/mojo/daemon.t
```diff
@@ -109,6 +109,16 @@ subtest 'Transaction' => sub {
 my $app = Mojolicious->new;
 my $ua  = Mojo::UserAgent->new(ioloop => Mojo::IOLoop->singleton);
 
+my $whatever = 'Failed!';
+
```
kraih, line 113: Empty line.

t/mojo/cgi.t
```diff
@@ -9,7 +9,14 @@ use Mojolicious::Lite;
 # Silence
 app->log->level('fatal');
 
-get '/' => {text => 'Your Mojo is working!'};
+my $working = "NOT working";
+
```
kraih, line 13: Empty line.

lib/Mojolicious/Guides/Cookbook.pod
kraih: We use 120 columns.

lib/Mojo/Server/Daemon.pm
kraih: @jhthorsen Yes, that's bad encapsulation.

t/mojo/cgi.t
rmanfredi: Took care of it.

lib/Mojo/UserAgent/Server.pm
rmanfredi: Done, there is no need to change UserAgent::Server after all.

lib/Mojo/Server/CGI.pm
rmanfredi: @kraih you are right, something is not working as expected yet.  I only discovered Mojo 2 days ago, hence I apologize for being a slow learner.
I really wish to have a global hook that triggers possibly costly initialization for the application, only **once** and regardless how the application is actually started.

rmanfredi: @krait There, now it runs exactly once per process.
It will get run if you do `./myapp.pl get /` but not if you do `./myapp.pl eval -V ...` because then you are not going to serve a request, hence why perform global initialization only required for request serving?

t/mojo/morbo.t
rmanfredi: OK, I trust your judgment.  Removing specific morbo testing!

rmanfredi:
So, at this stage, I believe I have answered all the review comments so far.  I'm letting you review the current changes now, awaiting more feedback.

Note that the semantics of the `before_app_start` event is to be emitted **once** per process (and per application), for global initialization by the application before it starts processing web requests, regardless of how exactly it is being run.

This means we do not need to evaluate code that could lie in `./myapp.pl` before the app is actually started and is about to serve requests. The event actually becomes a global application initialization hook.  That's the intent anyway.

lib/Mojolicious/Guides/Cookbook.pod
rmanfredi: Reformatted to 120 columns since it is the standard for this project.

lib/Mojo/Server/Daemon.pm
rmanfredi: @jhthorsen @kraih I am sorry, I do not follow your logic.

Why is it bad encapsulation to send **advertised** events to an object so that it can react appropriately if it wants?  How is this violating encapsulation at all?  Calling `$app->app_starting` is no more violating encapsulation than calling `$app->server` is.  Or am I totally misunderstanding what you are talking about here?

lib/Mojo/Server/Daemon.pm
```diff
@@ -35,6 +35,7 @@ sub run {
   my $loop = $self->ioloop;
   my $int  = $loop->recurring(1 => sub { });
   local $SIG{INT} = local $SIG{TERM} = sub { $loop->stop };
+  $self->app->app_starting;
```
kraih, line 38: This should trigger twice now.

lib/Mojo/Server/CGI.pm
kraih: And that makes it inconsistent with `before_server_start`. Which does not run for `./myapp.pl cgi`.

lib/Mojolicious.pm
```diff
@@ -175,6 +175,16 @@ sub plugin {
 
 sub server { $_[0]->plugins->emit_hook(before_server_start => @_[1, 0]) }
 
+sub app_starting {
+  my $self = shift;
+
+  # Ensure we run 'before_app_start' only once per process
+  $self->{_started} //= {};
+  return if exists $self->{_started}->{$$};
+  $self->{_started}->{$$} = undef;
```
kraih, line 184: This should be completely unnecessary.

t/mojo/prefork.t
```diff
@@ -42,6 +45,62 @@ unlike $log, qr/Creating process id file/,     'right message';
...
+  }
+);
+$counter->routes->post(
+  '/server/:pid' => sub {
+    my $c   = shift;
+    my $pid = $c->stash('pid');
+    $counter_server->{$pid}++;
+    $c->rendered(201);
+  }
+);
```
kraih, line 88: Way too overcomplicated for a simple test.

kraih: Do the absolute minimum necessary to test the feature.

t/mojo/prefork.t
```diff
@@ -133,4 +223,39 @@ ok !$graceful, 'server has been stopped immediately';
...
+is all_values_one($href->{spawned}), 1, '1 spawn trigger per process';
+
+my $foreign = 0;
+foreach my $pid (keys %{$href->{events}}) {
+  $foreign++ unless exists $href->{spawned}->{$pid} || exists $href->{server}->{$pid};
+}
+
+is $foreign, 0, 'no before_app_start ran in foreign processes';
+
+kill 'TERM', $dpid;
```
kraih, line 259: This is a mess. Anything that risks making our fork tests less reliable is a blocker i'm afraid.

kraih (changes requested):

t/mojo/prefork.t
kraih: More constructively: Please make a self contained subtest (all new tests are required to use subtests), and make sure that it waits until the forked process is stopped safely.

t/mojo/daemon.t
```diff
@@ -109,6 +109,15 @@ subtest 'Transaction' => sub {
 my $app = Mojolicious->new;
 my $ua  = Mojo::UserAgent->new(ioloop => Mojo::IOLoop->singleton);
 
+my $whatever = 'Failed!';
+$app->hook(
+  before_app_start => sub {
+    my ($app) = shift;
```
kraih, line 115: We don't use shift in list context.

t/mojo/daemon.t
```diff
@@ -109,6 +109,15 @@ subtest 'Transaction' => sub {
...
+    my $ok = ref $app && $app->isa('Mojolicious');
+    $whatever = $ok ? 'Whatever!' : 'Wrong type';
```
kraih, line 117: Way too complicated again. Why have special case that is now untested?

t/mojo/prefork.t
```diff
@@ -76,6 +141,13 @@ $prefork->app->hook(
   before_server_start => sub {
     my ($server, $app) = @_;
     push @server, $server->workers, $app->mode;
+    Mojo::UserAgent->new->post("$durl/server/$$");
+  }
+);
+$prefork->app->hook(
+  before_app_start => sub {
+    my ($app) = @_;
```
kraih, line 149: It's not a list.

lib/Mojo/Server/Daemon.pm
kraih: Now you're crossing two layers instead of just one. The server knowing about a specific set of methods it can call on the app instance it manages is very different from it digging into very specific app properties that would normally only be reached via delegation by the app. It makes implementing new servers much much harder. Just because it is public doesn't make it part of the "server <-> app" protocol.

lib/Mojo/Server/CGI.pm
kraih: I think this is a blocker too. We might have to consider adding the `before_server_start` hook to cgi first.

lib/Mojolicious.pm
```diff
@@ -175,6 +175,16 @@ sub plugin {
...
+  $self->plugins->emit_hook(before_app_start => $self);
```
kraih, line 185: I find the new name misleading. It implies that it runs before `app->start`.

lib/Mojo/Server/Daemon.pm
rmanfredi: It did, hence the PID protection in `Mojolicious`.
If I remove this line, then `morbo ./myapp.pl` is not triggering the event!

lib/Mojolicious.pm
rmanfredi: It's not.  See above comment... If I do not do this, then there is a duplicate invocation (as you guessed) with having run() and start() both invoke the event...
Yet if both do not, then the event is not always emitted.

t/mojo/daemon.t
rmanfredi: Right.  I meant `($app) = @_;` but as your earlier comment indicates, you use `$app = shift` when there is just one argument expected.

t/mojo/daemon.t
rmanfredi: What do you mean "untested"? This is tested indeed... if I change the 'Whatever!' string, the tests are failing!
Or did you mean something I don't get?

lib/Mojolicious.pm
rmanfredi: I agree here.  In line with other comment from @s1037989 yesterday.
Naming **is** hard.  What would you choose? `before_app_serving`?  Or perhaps drop the `before_` and say `start`?  Or `init`?

t/mojo/prefork.t
rmanfredi: Not familiar with the `subtest` thing. `t/mojo/prefork.t` did not use these.
Are you suggesting that I put *all* the preforking tests into a single subtest?

With my little webapp, I can collect statistics about all the tests done and make sure everything is working as expected: that callbacks are invoked the expected amount of time and in the right processes.  Since we have many concurrent processes, I thought it would be nicer to cleanly abstract the statistics collection into a single tracking service that every sub-process knows how to reach and which has a single REST API.

I can add a `waitpid()` call at the end to make sure we cleanup properly, but I have added a failsafe in the webapp by instantiating a callback every 0.5 second to check whether the child has been re-parented to `init` (sorry, one should say `systemd` nowadays).  Did I do that right?

lib/Mojo/Server/CGI.pm
rmanfredi: Why do you try to link the fate of `before_server_start` with that of `before_app_start`? They have different semantics now.

We could add `before_server_start` to CGI, but that would be another PR, because, for me, the two events have different semantics, even if, at first, when I started toying with this idea, I emitted one close to the other...

t/mojo/prefork.t
rmanfredi: Aren't parameters always passed as a LIST in Perl?  Even if there is a single one?

lib/Mojolicious.pm
kraih: `after_server_start`, consistent with `before_server_start`? 😁

t/mojo/prefork.t
kraih: Yes, but we use the convention `my $app = shift` for single arguments. It became an idiom because it used to be a little faster with some versions of Perl (i don't actually remember the details, ask on IRC if you want to know more).

t/mojo/daemon.t
kraih: The `Wrong type` variant appears untested and rather pointless. You could just ref the app and use that as test value.

kraih: Or better yet, call a method in `$app` and use that, like `->moniker` or so.

t/mojo/prefork.t
kraih: Please leave the existing tests alone, and treat everything new as a separate new subtest. We will convert the rest of `prefork.t` to subtests in time. Make the test app no more complicated than absolutely necessary. Every line that looks unnecessary will be questioned and might end up as a blocker.

kraih: The conversion is an ongoing process independent of this PR. https://github.com/mojolicious/mojo/issues/1520

lib/Mojolicious.pm
rmanfredi: Sure, but that would mean it has comparable semantics to some "server starting" event and... remember... there is no such thing for CGI yet by your own admission... :grinning: 

That being said, I have absolutely no stake in this naming.  I just want to have this new functionality in **standard** Mojolicious and not just in my own forked framework.  Forking *Open-Source* projects is always... bad.

t/mojo/prefork.t
rmanfredi: Yeah, I remember that discussion on p5p at the time.  It was a loooong time ago.  In my own code, I just don't care and prefer to say `($x) = @_;` even if there is just one argument because I favor readability over hypothetical performance.  My only exception is for `my $self = shift;` which I always use on a line by itself, even if there are following arguments.

But `Mojolicious` is not _my_ code and I will follow your standards by amending this code once again.

lib/Mojo/Server/Daemon.pm
kraih: Right, that's a different problem though. This is newly designed feature, so it really shouldn't use workarounds like that.

lib/Mojolicious.pm
kraih: It's a workaround, not a solution for the actual problem (which is that morbo initialises the listener in the parent process before the application is ever loaded).

lib/Mojolicious.pm
kraih: Note that Mojolicious gets forked so much that we had to adopt an official policy for it. https://docs.mojolicious.org/Mojolicious/Guides/Contributing#FORK-POLICY

t/mojo/prefork.t
kraih: If it was up to me alone we would just be using signatures everywhere. 😉

lib/Mojo/Server/CGI.pm
kraih: Yes, i think `before_server_start` should be added to `Mojo::Server::CGI` after all.

kraih: And done. https://github.com/mojolicious/mojo/compare/81a2753e992e...8e1611e3682a

kraih:
I've been thinking about this for a bit, and there are more edge cases that have not been considered yet. The best solution i can think of would be something like:
```diff
diff --git a/lib/Mojo/Server/Daemon.pm b/lib/Mojo/Server/Daemon.pm
index f85982841..98ddcee24 100644
--- a/lib/Mojo/Server/Daemon.pm
+++ b/lib/Mojo/Server/Daemon.pm
@@ -54,6 +54,7 @@ sub start {
   elsif (!@{$self->acceptors}) {
     $self->app->server($self);
     $self->_listen($_) for @{$self->listen};
+    $self->ioloop->next_tick(sub { $self->app->server($self, 1) });
   }
 
   return $self;
```

kraih:
PR seems abandoned. Given our growing review backlog, i'm closing this to make room for other PRs.

PR #1768, opened 2021-04-17
Mojolicious::Guides::Cookbook - Include User settings in example services files

Grinnz:
### Summary
Add example User settings in the service file examples for deployment.

### Motivation
Mojolicious servers default to deploying to ports 3000 and 8080, which don't require elevated privileges, so starting the service as a regular user is a better default suggestion.

### References
https://www.freedesktop.org/software/systemd/man/systemd.exec.html#User/Group%20Identity

### DIFF
--- a/lib/Mojolicious/Guides/Cookbook.pod
+++ b/lib/Mojolicious/Guides/Cookbook.pod
@@ -147,6 +147,7 @@ To manage the web server with systemd, you can use a unit configuration file lik
 
   [Service]
   Type=simple
+  User=sri
   ExecStart=/home/sri/myapp/script/my_app daemon -m production -l http://*:8080
 
   [Install]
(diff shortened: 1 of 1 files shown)


lib/Mojolicious/Guides/Cookbook.pod
```diff
@@ -147,6 +147,8 @@ To manage the web server with systemd, you can use a unit configuration file lik
 
   [Service]
   Type=simple
+  User=sri
+  Group=sri
```
kraih, line 151: I would just leave out the group, or at least call it something like `users`.

Grinnz: I usually want to ensure a process does not remain in the `root` group, but the docs do say: "If no group is set, the default group of the user is used." So leaving out the group is reasonable.

kraih (approved):

tianon (approved):
(I'm not core, but I seem to recall non-core reviews are appreciated 😅👍)

kraih:
@tianon Very much appreciated. 😃

jhthorsen (approved):
LGTM 👍

PR #1773, opened 2021-04-27
Mojolicious::Guides::Testing - mention how to test Mojolicious::Lite applications

Grinnz:
### Summary
Include the example from Test::Mojo/new to show how to load a Mojolicious::Lite application with Test::Mojo.

### Motivation
The guide does not currently specify how to test a Mojolicious::Lite application which doesn't have an application class.


jberger:
I wonder if it should mention how it disambiguates between path and class? It kinda makes sense intuitively but later it makes you say "hmmmm".

jhthorsen (approved):
I like it 👍

jberger (approved):
Yeah, I do too. I thought about what specific change I'd be asking for and honestly I couldn't come up with one. Yes I had to look in the code to see how it disambiguated, but in practice it won't matter.

PR #1775, opened 2021-04-30
Promise map improved

mjaix:
### Summary
Generalized Mojo::Promise::Map to also support 'any' and 'all_settled' aggregations plus an optional delay

### Motivation
Since the deprecation/removal of Mojo::IOLoop::delay, the only concurrency-limiting elegant way is to use Mojo::Promise::map.
Since I need the limiting as part of a general purpose functionality, I augmented Mojo::Promise::map to support various kinds of aggregation (all[default], any, all_settled) and an optional delay between processing the promises.
Test cases as well as a cookbook entry for Mojo::Promise::map is supplied.

### References
LIST RELEVANT ISSUES, PULL REQUESTS AND FORUM DISCUSSIONS HERE

### DIFF
--- a/lib/Mojo/Promise.pm
+++ b/lib/Mojo/Promise.pm
@@ -75,18 +75,27 @@ sub map {
   my $start_next = sub {
     return () unless my $item = shift @items;
     my ($start_next, $chain) = (__SUB__, shift @wait);
-    $_->$cb->then(
-      sub {
-        $chain->resolve(@_);
-        if ($akey == 1) { @items = () }
-        else            { $start_next->() }
-      },
-      sub {
-        $chain->reject(@_);
-        if ($akey == 0) { @items = () }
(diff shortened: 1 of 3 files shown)


mjaix:
Ruminating a little bit about not support 'race', I think that should be also done. Albeit 'race' could be simply achieved by applying race() directly to the first n items to be started concurrently, it should be supported in map() to get that function "complete" regarding the support of all possible aggregations.

jhthorsen (commented):
I know this change is blocked, since it's not rebased, but I wanted to chime in that I'm leaning toward 👎 . I think this is a bit too much to include in core Mojolicious, and I think it could be added as a role on CPAN instead.

s1037989:
Is this PR still needed?
@mjaix What do you think about making this a role?

PR #1779, opened 2021-05-05
Add Try/Catch syntax feature

doojonio:
### Summary
In Perl 5.34 try-catch syntax [was added](https://metacpan.org/pod/release/XSAWYERX/perl-5.34.0-RC1/pod/perldelta.pod). So I think that it's a good idea to add flag to Mojo::Base import subroutine for this feature
### Motivation
Try/Catch syntax is using to avoid [some problems](https://metacpan.org/pod/Try::Tiny#Clobbering-$@) with eval and $@
### References
I created discussion [here](https://github.com/mojolicious/mojo/discussions/1778)

### DIFF
--- a/lib/Mojo/Base.pm
+++ b/lib/Mojo/Base.pm
@@ -128,6 +128,13 @@ sub import {
       experimental->import($_) for qw(signatures postderef);
     }
 
+    # Try/Catch syntax (Perl 5.34+)
+    elsif ($flag eq '-try_catch') {
+      Carp::croak 'Try/Catch syntax requires Perl 5.34+' if $] < 5.034;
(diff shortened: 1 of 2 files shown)


Grinnz:
This needs documentation in Mojo::Base and Mojolicious::Lite.

Grinnz:
As for suitability, I think it's a good idea but I'm not sure if we should do it yet. The only reason the -signatures flag was added for an experimental feature is because it's been experimental for very long and stable in practical terms.

kraih:
Agreed, this seems a bit early.

doojonio:
Thanks for feedback. Now I think too that it's too early

PR #1780, opened 2021-05-09
Allow inline templates to use layouts.

mkende:
### Summary
Allow inline templates to use layouts. This requires the layout to be specified directly in the call to `render` or in the template itself, same as when using the `render_to_string` method.

**Note**:  this is edited from the initial version where a layout set in the stash was also applied. The current version of this PR starts at [this comment](https://github.com/mojolicious/mojo/pull/1780#issuecomment-859111008).

### Motivation
There is no reason to *disallow* inline templates from using layouts if they can work and if doing so won’t break existing users.

In my particular case I am building a site where users can provide custom themes that are mostly implemented as templates. These templates can’ be guaranteed to have unique names so i can’t add there parents directory to the templates directories of the application. Using inline template is a workaround but currently prevent using layouts. This PR is attempting to fix this.

This is my first contributing to Mojolicious, so I may be missing something and the fix may not be correct (although all current tests pass, except those that specifically tested that inline template didn’t work with layouts and that are now fixed).

### References
None.

### DIFF
--- a/lib/Mojolicious/Controller.pm
+++ b/lib/Mojolicious/Controller.pm
@@ -133,13 +133,14 @@ sub render {
 
   # Localize "extends" and "layout" to allow argument overrides
   my ($maybe, $ts) = @{$args}{'mojo.maybe', 'mojo.string'};
-  my $stash = $self->stash;
+  my $stash  = $self->stash;
+  my $inline = exists $args->{inline} || exists $stash->{inline};
   local $stash->{layout}  = $stash->{layout}  if exists $stash->{layout};
   local $stash->{extends} = $stash->{extends} if exists $stash->{extends};
 
-  # Rendering to string
+  # Rendering to string or using inline templates
   local @{$stash}{keys %$args}         if $ts || $maybe;
-  delete @{$stash}{qw(layout extends)} if $ts;
+  delete @{$stash}{qw(layout extends)} if $ts || $inline;
 
   # All other arguments just become part of the stash
   @$stash{keys %$args} = values %$args;
--- a/lib/Mojolicious/Guides/Rendering.pod
+++ b/lib/Mojolicious/Guides/Rendering.pod
@@ -525,8 +525,9 @@ To set a C<layout> stash value application-wide you can use L<Mojolicious/"defau
 
   $app->defaults(layout => 'mylayout');
 
-Layouts can also be used with L<Mojolicious::Controller/"render_to_string">, but the C<layout> value needs to be passed
-as a render argument (not a stash value).
+Layouts can also be used with L<Mojolicious::Controller/"render_to_string">, or when rendering
+L<inline templates|/"Rendering inline templates">, but the C<layout> value needs to be passed as a render argument
+(not a stash value).
 
   my $html = $c->render_to_string('reminder', layout => 'mail');
 
--- a/lib/Mojolicious/Renderer.pm
+++ b/lib/Mojolicious/Renderer.pm
@@ -108,7 +108,8 @@ sub render {
   # Inheritance
   my $content = $stash->{'mojo.content'} //= {};
   local $content->{content} = $output =~ /\S/ ? $output : undef if $stash->{extends} || $stash->{layout};
-  while ((my $next = _next($stash)) && !defined $inline) {
+  delete $options->{inline};
+  while ((my $next = _next($stash))) {
     @$options{qw(handler template)} = ($stash->{handler}, $next);
     $options->{format} = $stash->{format} || $self->default_format;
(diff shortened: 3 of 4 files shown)


CandyAngel (commented):
Done a quick bit of digging and it looks like inline templates using layouts was experimental in the past and was removed in 677fe638566ad9af204fb4f76f31c841bec2e8c0 due to causing issues.

kraih:
If i remember correctly the problem was users setting `app->defaults(layout => 'default')` and then plugins that used inline rendering for various things misbehaving because their results suddenly had a layout wrapper.

mkende:
That situation seems like a reasonable semantics and such a plugin is already broken if it uses a non-inline layout. That is, I would argue that the problem is either in setting a global layout or in the plugin not protecting against that.

If it helps, I added a test that setting "layout => undef" when rendering an inline template correctly disable a default layout.

mkende:
Another option would be to also delete the layout and extends stash value in Mojolicious::Controller::render() when rendering an inline template and not just when rendering to string (I guess that part has been implemented to fix that same problem with plugins — it seems improbable by the way that a plugin would not be rendering to a string if it wants to ignore the layout).

I think that this would solve my issue (being able to use layouts with inline template), but feels like adding a hack around the fact that using a global layout is anyway a dangerous thing (reading the documentation I already wondered why rendering to string behaved differently from the rest).

mkende:
I understand that. What I meant is that having the global default layout applies to all template rendering, independently of whether the template is rendered to a string or not, or whether the template comes from a string or not seems a reasonable API, even if it can break existing behavior. And I was claiming that possibly code relying on the special casing of inline template needs to be fixed (either by not using a global default layout or by explicitly setting "layout => undef" when rendering a template in a plugin and you want to be sure to not be impacted by a global default template).

But as this might be difficult to do for all the existing code an alternative option might be to extend the current special handling of template rendered to string, to inline template and ignore the default layout in that case

To summarize the situation and possible options :

|    | *normal* template| rendered to string | inline template |
| --- | :---: | :---: | :---: |
| current behavior |use layout | ignore global | never use |
| this PR |use layout | ignore global | use layout |
| alternative 1 |use layout | use layout | use layout |
| alternative 2 |use layout | ignore global | ignore global |

The current behavior is the most complex because each type of template has different rules. This PR and the alternative 1 are simpler and, subjectively, I think that they offer a more consistent API, but may be difficult to roll-out due to the breaking change to existing code. Alternative 2 might be offer a good balance.

Let me know what you think and I’m happy to implement alternative 1 or 2 if you would prefer one of them to this PR or the current situation.

mkende:
Following the discussion on IRC two days ago which seems to indicate that PR #1780 is blocked for now, I would be interested in feedback on this PR here, in particular on what I called "option 2" in my previous comment (and which is now the version that is implemented here).

The PR does not introduce new keywords or new API, it should not break any existing use*, it makes the existing API more consistent, and it allows new features. So, I would hope that this is not a controversial change.

*The only cases that are changed are calls that are using inline template and explicitly specifying a layout in the call to`render` (or in the template itself), as the layout will no longer be ignored.

kraih:
Kinda related, mojo.js is testing inline layouts. https://github.com/mojolicious/mojo.js/blob/main/docs/Introduction.md#layouts

mkende:
Well I saw your message the other day about mojo.js being 15 times faster than Mojolicious and I’m sure there are plenty of other similar reasons to pick up Node these days, but I just like Perl too much to switch.

kraih:
No, you seem to be misunderstanding what i meant. All the feature we are testing with mojo.js can be backported to Mojolicious if they are a success.

mkende:
Can a template declare its own layout in mojo.js as in Mojolicious? Because having a different stash keyword for the layout of inline template is reasonable but the template itself should not have to know whether it is an inline or non inline template (so, in Mojolicious, my expectation would be that the same `layout` helper should work if used from inside a layout executed inline).

mkende:
Kiwiroy, do you have other concerns with this PR? If not, do you feel like approving it if you have write access?

mkende:
Any chance this can be reviewed? I understand that Kraih does not have the time currently for that, but maybe there are other maintainer of the project who could review it? If not, what should be my expectation here about having this PR eventually reviewed?

kraih:
Because of the side effects and better alternative implemented by mojo.js, i have to give this proposal a 👎 .

mkende:
What side effect? (and what alternative is mojo.js using ? can this be implemented in Mojolicious?)

marcusramberg:
I think this can be closed, if this functionality is to be added #1887 seems a more likely candidate.

PR #1786, opened 2021-05-18
Tidy some test imports

oalders:
### Summary
Tidying up test imports using https://metacpan.org/pod/perlimports

### Motivation
I've documented the motivation for these changes at https://metacpan.org/pod/App::perlimports

### References
#1747 

I'd be interested in knowing if you find these changes useful, objectionable, a waste of time or some combination of the above.

### DIFF
--- a/t/mojo/json_xs.t
+++ b/t/mojo/json_xs.t
@@ -17,7 +17,7 @@ sub TO_JSON { shift->something }
 package main;
 
 use Mojo::ByteStream;
-use Mojo::Util qw(decode encode);
+use Mojo::Util qw(encode);
 
 subtest 'Basics' => sub {
   my $array = decode_json '[]';
--- a/t/mojo/roles.t
+++ b/t/mojo/roles.t
@@ -50,7 +50,7 @@ package main;
 use Mojo::ByteStream;
 use Mojo::Collection;
 use Mojo::DOM;
-use Mojo::File;
+use Mojo::File ();
 
 # Plain class
 my $obj = Mojo::RoleTest->new(name => 'Ted');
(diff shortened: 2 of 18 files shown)


t/mojo/json_xs.t
```diff
@@ -16,8 +16,8 @@ sub TO_JSON { shift->something }
 
 package main;
 
-use Mojo::ByteStream;
-use Mojo::Util qw(decode encode);
+use Mojo::ByteStream ();
```
kraih, line 19: This one is wrong though, there's no default export.

t/mojo/roles.t
```diff
@@ -47,10 +47,10 @@ sub hello {'hello mojo!'}
 
 package main;
 
-use Mojo::ByteStream;
-use Mojo::Collection;
+use Mojo::ByteStream ();
+use Mojo::Collection ();
```
kraih, line 51: Also wrong, no default export.

oalders: The idea here is that even though the bare parens actually don't do anything, they serve to document that this statement will not be the source of any imported symbols. For someone already familiar with those classes it's obvious, but for someone else maybe not so much.

Having said that, I can force-push some changes with these removed. What I'm trying to do is get an idea of what people prefer when it comes to imports and this is helpful feedback.

jberger: I don't think it matters in this case but doesn't calling use with bare parens prevent calling import? That would indeed change behavior in some cases

oalders: Right. The bare parens can actually change the behaviour, since `import()` won't get called at all in that case.

oalders:
Thanks for the feedback! I'm going to close this now.

PR #1791, opened 2021-06-01
pad session values to 1025 bytes by default

jberger:
Pad the session value to at least 1025 bytes. This should prevent a small value from being used as part of a brute-force attack to decode the session secret.

### DIFF
--- a/lib/Mojolicious/Sessions.pm
+++ b/lib/Mojolicious/Sessions.pm
@@ -8,9 +8,9 @@ has [qw(cookie_domain secure)];
 has cookie_name        => 'mojolicious';
 has cookie_path        => '/';
 has default_expiration => 3600;
-has deserialize        => sub { \&Mojo::JSON::j };
+has deserialize        => sub { \&_deserialize };
 has samesite           => 'Lax';
-has serialize          => sub { \&Mojo::JSON::encode_json };
+has serialize          => sub { \&_serialize };
 
 sub load {
   my ($self, $c) = @_;
@@ -59,6 +59,14 @@ sub store {
     secure   => $self->secure
   };
   $c->signed_cookie($self->cookie_name, $value, $options);
+}
+
+sub _deserialize { Mojo::JSON::decode_json($_[0] =~ s/\}\KZ*$//r) }
+
+sub _serialize {
+  no warnings 'numeric';
+  my $out = Mojo::JSON::encode_json($_[0]);
+  return $out . 'Z' x (1025 - length $out);
 }
 
 1;
--- a/t/mojolicious/lite_app.t
+++ b/t/mojolicious/lite_app.t
@@ -274,6 +274,12 @@ get '/session_cookie/2' => sub {
   $c->render(text => "Session is $value!");
 };
 
+get '/session_length' => sub {
+  my $c = shift;
+  $c->session->{q} = $c->param('q');
+  $c->rendered(204);
+};
+
 get '/foo' => sub {
   my $c = shift;
   $c->render(text => 'Yea baby!');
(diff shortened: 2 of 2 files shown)


Grinnz:
If the padding byte doesn't matter I'd prefer to use something more indicative like null bytes maybe?

jberger:
> If the padding byte doesn't matter I'd prefer to use something more indicative like null bytes maybe?

I suppose that could work since it is base64 encoded. Originally I did think about null, but I chose a text character because I was worried about how it would interact with the HTTP protocol etc, but on reflection that doesn't matter once base64 encoded. Anyway, I don't care what the padding is. In theory it could be random as long as it isn't `}` (by this implementation).

Grinnz:
The other option is to do this padding after the serialization and base64 so it happens regardless of the serializer. It would have to be some bytes that are not valid in base64 but are valid in cookie values (punctuation other than `",;\/+` would work).

kraih (approved):

jhthorsen (approved):
I think this is a very clever solution 👍

latk:
I am trying to understand the rationale of this change (prompted by [a Stack Overflow question](https://stackoverflow.com/questions/74245591/mojolicious-weird-cookies)). The stated purpose of this change is to make brute-force key recovery attacks more difficult, by padding the cookie payload until it is at least 1025 bytes long. The length seems to have been chosen arbitrarily. The padded cookie value is later used in Controller.pm as input for a HMAC:

```perl
  my $sum = Digest::SHA::hmac_sha256_hex("$name=$value", $self->app->secrets->[0]);
  return $self->cookie($name, "$value--$sum", $options);
```

I think the change in this PR fails to achieve its stated purpose:

1. The difficulty of a brute-force attack depends on the entropy of the input, not directly on its length. Adding deterministic padding does not add entropy, and therefore does not make brute-force attacks more difficult. At most, increasing the input length produces a linear slowdown.

2. As far as I understand HMACs, the difficulty of a key recovery attack depends only on the strength of the secret key, up to the limit imposed by the hash function's internal size. The difficulty for key recovery does not depend on the size of the plaintext message, which is known to the attacker. For the same reason, using bytes from a CSPRNG for padding the message as a kind of salt would not help.

If my understanding happens to be correct, it could make sense to revert this change. **The primary effect of this PR seems to be increased bandwidth usage for users, not increased security.** If this padding is still desired, it could be added just before the HMAC calculation, without carrying it around in the cookie.

Towards the goal of defending against brute-force attacks, other mechanisms might be more helpful – such as extending the documentation for Mojolicious' secrets to explain how to generate secrets with sufficiently high entropy (e.g. `head -c 32 /dev/urandom | base64` or `openssl rand -base64 32`). [NIST recommends](https://doi.org/10.6028/NIST.SP.800-131Ar2) key lengths of at least 112 bits for HMACs. I would also mention that secrets would ideally be provided by the deployment environment, and never written into the application's source code.

<sub>(Since the HMAC maintains its security properties regardless of the changes in this PR, this is not a vulnerability. I therefore commented here instead of going through the vulnerability disclosure process.)</sub>

jberger:
It was a very specific brute force attack. I don't know if we're mentioning which openly but a common tool could be used if the encrypted message was short enough. Meanwhile you have to be able to strip it out easily and not cost too much performance since this happens a lot

CarterSScott:
I ran into an issue where I was trying to access the cookie that the session generates directly through the signed_cookie method. The padded Z's were still there and I had to write some regex to remove them. Should there be some sort of check in place for the signed_cookie method to remove these padded Z's?

stigtsp:
> It was a very specific brute force attack. [..]

Hi! Would you mind sharing some details about this attack?

s1037989:
Hi! Thanks for reaching out to discuss this! I'm just a community member
myself, but appreciate the topic. Are you saying that you're warning the
Mojolicious team that their security measure is about to have a cracking
tool released to the world? And, regardless, do you have a suggestion for
what could be done to improve the framework? Is the reported issue only the
simple default password, or is there another problem that should be solved
with some superior technique?

jkramarz:
@s1037989, more notifying that strapping together parts of existing modules with duct tape and bubble gum works just fine for cracking the signatures in efficient way.

Not sure when the token signatures switched from HMAC-SHA1 to HMAC-SHA256 (it predates this merge request), but since then it were already fine. Solution implemented here, while clever in what it achieves, in my opinion does not directly increase security of the tokens. It is equal to HMAC-SHA256 signed JWT, without usual JWT-specific attacks.

If the idea of environments (e.g. like in Ruby on Rails, `RAILS_ENV=production`, with separate configuration ) is present here, so there is some determinant sufficient to decide that it is no longer in development and it would be possible to prevent running application without secret configured - I would call it a day.

Grinnz:
There is a production mode, which can be set manually or by running the application with the hypnotoad application server. It may be reasonable to refuse to start in production mode without a configured secret, but that deserves its own issue/discussion.

kraih:
Maybe also worth mentioning that the app generator shipping with Mojolicious does generate a config file with secret: https://github.com/mojolicious/mojo/blob/ecb44cf8b1ec11402e28aac1e2a3224d505e95c4/lib/Mojolicious/Command/Author/generate/app.pm#L202

Should those be longer by default too?

stigtsp:
I've done some work on a proposal to fix this that seems to be in line with what @jkramarz is proposing: 
- #2200

PR #1828, opened 2021-08-06
Mojo::IOLoop example fix (minor)

Yenya:
The example from the Mojo::IOLoop page needs '-signatures',
and even with '-signatures' it fails with

Mojo::Reactor::Poll: I/O watcher failed: Too many arguments for subroutine 'main::__ANON__' at /usr/local/share/perl5/5.32/Mojo/IOLoop.pm line 112.

because the server handler gets third argument - the $id.

### DIFF
--- a/lib/Mojo/IOLoop.pm
+++ b/lib/Mojo/IOLoop.pm
@@ -233,7 +233,7 @@ Mojo::IOLoop - Minimalistic event loop
   use Mojo::IOLoop;
 
   # Listen on port 3000
-  Mojo::IOLoop->server({port => 3000} => sub ($loop, $stream) {
+  Mojo::IOLoop->server({port => 3000} => sub ($loop, $stream, $id) {
     $stream->on(read => sub ($stream, $bytes) {
       # Process input chunk
       say $bytes;


lib/Mojo/IOLoop.pm
```diff
@@ -230,10 +230,10 @@ Mojo::IOLoop - Minimalistic event loop
 
 =head1 SYNOPSIS
 
-  use Mojo::IOLoop;
+  use Mojo::IOLoop -signatures;
```
kraih, line 233: This is incorrect.

Yenya: Thanks for the review. Would you care to elaborate? Without -signatures, the example does not work for me, as the anonymous subroutines get no arguments.

Grinnz: -signatures is a flag for Mojo::Base or Mojolicious::Lite. It is not supported by Mojo::IOLoop. Signatures are assumed in many examples without being specified everywhere; see https://metacpan.org/pod/Mojolicious::Guides#Signatures and [https://metacpan.org/pod/Mojolicious::Guides::FAQ#What-does-%22Illegal-character-in-prototype%22-mean?](https://metacpan.org/pod/Mojolicious::Guides::FAQ#What-does-%22Illegal-character-in-prototype%22-mean?).

Yenya: OK, thanks for explanation. It seems that -signatures is accepted by Mojo::IOLoop nevertheless, as it probably inherits from Mojo::Base.

Removed the -signatures part from the merge request.

PR #1877, opened 2021-11-10
doc: reword testing guide paragraph

bmeneg:
### Summary
Reword the paragraph explaining the same code but with two different approaches.

### Motivation
The way the paragraph is written now may lead the reader to think that explanation is valid only for one of the snippets instead of the two it's actually referring to. To avoid it, slightly change the first sentence.

### DIFF
--- a/lib/Mojolicious/Guides/Testing.pod
+++ b/lib/Mojolicious/Guides/Testing.pod
@@ -159,9 +159,9 @@ To test a L<Mojolicious::Lite> application, pass the file path to the applicatio
   use Mojo::File qw(curfile);
   my $t = Test::Mojo->new(curfile->dirname->sibling('myapp.pl'));
 
-This object initializes a L<Mojo::UserAgent> object, loads the Mojolicious application C<Frogs>, binds and listens on a
-free TCP port (e.g., 32114), and starts the application event loop. When the L<Test::Mojo> object (C<$t>) goes out of
-scope, the application is stopped.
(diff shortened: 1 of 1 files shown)


Grinnz:
I don't follow - the paragraph is explaining both examples, which are equivalent, one for full apps and one for lite apps.

bmeneg:
@Grinnz ooh, I see what confused me:
```
This object initializes a L<Mojo::UserAgent> object, loads the Mojolicious application C<Frogs>
```
The second snippet loads the application from the file `myapp.pl`, so the wording is odd and led me to understand that this paragraph was referring only to the snippet `my $t = Test::Mojo->new('Frogs');`.

But yes, you're right, the paragraph is valid for both snippets and is in the correct place.
What if we only remove the `C<Frogs>`? Something like:
```diff
- This object initializes a L<Mojo::UserAgent> object, loads the Mojolicious application C<Frogs>, binds and listens on a (...)
+ The object initializes a L<Mojo::UserAgent> object, loads the Mojolicious application, binds and listens on a (...)
```

bmeneg:
@Grinnz I updated the patch with the changes I mentioned in my last comment.
Tell me what you think.

(( It all might be a completely misunderstanding on my side only as a non-native/bad English speaker, so if you prefer to drop the change completely, that's fine by me 😃 no worries at all. ))

Grinnz (approved):

jberger (approved):

jhthorsen (approved):

jberger:
The blocking tidy problems have been resolved, I'm going to rebase/merge this. Thanks

PR #1878, opened 2021-11-11
use the new perltidy

jberger:
update the codebase for the new perltidy so that PRs don't fail the perltidy test.


jberger:
@mojolicious/maintainers please review this

kraih (approved):

karenetheridge:
Why not instead update the .perltidyrc so the same preferences are maintained, and to keep the diff to a minimum?

jhthorsen (approved):
I hope perltidy doesn’t revert the changes 😃

marcusramberg:
Afaict these changes are for the better.

Issue #1873, opened 2021-11-07
UA can't upload a Mojo::Asset::File having start_range & end_range


akarelas:
* Mojolicious version: 9.22
* Perl version: v5.34
* Operating system: Ubuntu Linux 20.04

### Steps to reproduce the behavior
Attempts to upload part of a file with Mojo::UserAgent, using Mojo::Asset::File and the asset's start_range/end_range properties fail after a long pause.

```perl
use Data::Dumper;

my $asset = Mojo::Asset::File->new(
    path        => 'file.txt',
    start_range => 100,
    end_range   => 105,
);

my $ua = Mojo::UserAgent->new;

my $tx = $ua->post('http://localhost:3000/upload_file', form => {
    up => { file => $asset },
});

print Dumper($tx->error);
```
### Expected behavior

I expected those 6 bytes to be uploaded quickly and no error to occur.

### Actual behavior

After 60 seconds I got this error:

```
$VAR1 = {
          'error' => {
                       'message' => 'Premature connection close'
                     }
        };
```

kiwiroy:
`Mojo::Asset::File->size()` only ever returns the size of the whole asset so the multipart message has the wrong content length, hence the *premature connection close* as the server waits for more content. It can be done with a subset of the asset by creating a new one to pass in the form `{file => Mojo::Asset::File->new->add_chunk($asset->get_chunk($start, $end))}`

kraih:
That seems like a bug.

PR #1879, opened 2021-11-14
Return size of range when is_range is true

kiwiroy:
### Summary
This PR makes `Mojo::Asset::File` and `Mojo::Asset::Memory` return the size of the _range_ if `is_range` is true for the asset.

### Motivation
It seems inconsistent to return the size of the whole asset in the case when a range has been applied to an asset and is wrong when calculating content length for uploads with such assets.

### References
* #1873

### DIFF
--- a/lib/Mojo/Asset/File.pm
+++ b/lib/Mojo/Asset/File.pm
@@ -49,7 +49,7 @@ sub contains {
   $handle->sysseek($self->start_range, SEEK_SET);
 
   # Calculate window size
-  my $end  = $self->end_range // $self->size;
+  my $end  = $self->_end_range;
   my $len  = length $str;
   my $size = $len > 131072 ? $len : 131072;
   $size = $end - $self->start_range if $size > $end - $self->start_range;
@@ -116,7 +116,12 @@ sub new {
   return $file;
 }
 
-sub size { -s shift->handle }
+sub size {
+  my $self = shift;
+  my $size = -s $self->handle;
+  return $size unless $self->is_range;
+  return $self->_end_range - $self->start_range + 1;
+}
 
 sub slurp {
   my $handle = shift->handle;
@@ -127,6 +132,11 @@ sub slurp {
 }
 
 sub to_file {shift}
+
+sub _end_range {
+  my $self = shift;
+  return (!$self->is_range && -s $self->handle) || $self->end_range || (-s $self->handle) - 1;
+}
 
 1;
 
@@ -245,7 +255,7 @@ Construct a new L<Mojo::Asset::File> object.
 
   my $size = $file->size;
 
-Size of asset data in bytes.
+Size of asset data in bytes. If <Mojo::Asset/"is_range"> is true, the size is that of the range.
 
 =head2 slurp
 
--- a/lib/Mojo/Asset/Memory.pm
+++ b/lib/Mojo/Asset/Memory.pm
@@ -24,7 +24,7 @@ sub contains {
   my $start = $self->start_range;
   my $pos   = index $self->{content} // '', $str, $start;
   $pos -= $start if $start && $pos >= 0;
-  my $end = $self->end_range;
+  my $end = $self->_end_range(0);
 
   return $end && ($pos + length $str) >= $end ? -1 : $pos;
 }
@@ -41,12 +41,21 @@ sub get_chunk {
 
 sub move_to { path($_[1])->spurt($_[0]{content} // '') and return $_[0] }
 
-sub size { length(shift->{content} // '') }
+sub size {
+  my $self = shift;
+  return length($self->{content} // '') unless $self->is_range;
+  return $self->_end_range(-1) - $self->start_range + 1;
+}
 
 sub slurp { shift->{content} // '' }
 
 sub to_file { Mojo::Asset::File->new->add_chunk(shift->slurp) }
 
+sub _end_range {
+  my $self = shift;
+  return undef unless $self->is_range;
+  return $self->end_range || shift() + length($self->{content} // '');
+}
 1;
 
 =encoding utf8
@@ -140,7 +149,7 @@ Move asset data into a specific file.
 
   my $size = $mem->size;
 
-Size of asset data in bytes.
+Size of asset data in bytes. If <Mojo::Asset/"is_range"> is true, the size is that of the range.
 
 =head2 slurp
 
--- a/t/mojolicious/upload_lite_app.t
+++ b/t/mojolicious/upload_lite_app.t
@@ -45,6 +45,25 @@ subtest 'Asset and filename' => sub {
     ->content_is('xlalalatset');
 };
 
+subtest 'Asset and filename with range' => sub {
+  my $file = Mojo::Asset::File->new->add_chunk('lalaln' x 10)->add_chunk('z')->start_range(19)->end_range(30);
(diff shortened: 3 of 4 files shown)


lib/Mojo/Asset/File.pm
```diff
@@ -128,6 +133,13 @@ sub slurp {
 
 sub to_file {shift}
 
+sub _end_range {
+  my $file = shift;
+  return undef unless $file->is_range;
+  return $file->end_range if $file->end_range;
```
kraih, line 139: Calling the same method twice is very inefficient.

kiwiroy: Now directly accessing hash element.

lib/Mojo/Asset/File.pm
```diff
@@ -128,6 +133,13 @@ sub slurp {
 
 sub to_file {shift}
 
+sub _end_range {
+  my $file = shift;
+  return unless $file->is_range;
+  return $file->{end_range} if $file->{end_range};
```
kraih, line 139: `end_range` is not provided by this class, so it should use the accessor.

kraih: as in `if (my $end_range = $self->end_range) { return $end_range }`.

lib/Mojo/Asset/File.pm
```diff
@@ -128,6 +133,13 @@ sub slurp {
 
 sub to_file {shift}
 
+sub _end_range {
+  my $file = shift;
+  return unless $file->is_range;
```
kraih, line 138: This `return` is missing a return value.

lib/Mojo/Asset/Memory.pm
```diff
@@ -41,12 +41,22 @@ sub get_chunk {
...
+  return $self->_end_range(-1) - $self->start_range + 1;
+}
 
 sub slurp { shift->{content} // '' }
 
 sub to_file { Mojo::Asset::File->new->add_chunk(shift->slurp) }
 
+sub _end_range {
+  my $self = shift;
+  return unless $self->is_range;
```
kraih, line 56: Missing `return` value.

t/mojolicious/upload_lite_app.t
```diff
@@ -45,6 +45,24 @@ subtest 'Asset and filename' => sub {
...
+    ->content_is('xlalalnlalalnlalalnlalalnlalalnltset', 'content correct with only end set');
+  $file = Mojo::Asset::File->new->add_chunk('lalala' x 10)->add_chunk('z')->start_range(51);
+  $t->post_ok('/upload' => form => {file => {file => $file, filename => 'x'}, test => 'tset'})->status_is(200)
+    ->content_is('xalalalalaztset', 'content correct with only start set');
+  $file->add_chunk('est');
+  $t->post_ok('/upload' => form => {file => {file => $file, filename => 'x'}, test => 'tset'})->status_is(200)
+    ->content_is('xalalalalazesttset', 'content correct after chunk added');
+  $file->start_range(0);
+  $t->post_ok('/upload' => form => {file => {file => $file, filename => 'x'}, test => 'tset'})->status_is(200)
+    ->content_is('xlalalalalalalalalalalalalalalalalalalalalalalalalalalalalalazesttset', 'content correct no range');
```
kraih, line 63: I'm no expert, but `content correct no range` does not sound like proper english to me.

lib/Mojo/Asset/File.pm
```diff
@@ -116,7 +116,12 @@ sub new {
   return $file;
 }
 
-sub size { -s shift->handle }
+sub size {
+  my $file = shift;
```
jhthorsen, line 120: This should be $self and not $file imo. (same multiple places)

lib/Mojo/Asset/Memory.pm
```diff
@@ -41,12 +41,22 @@ sub get_chunk {
...
 
 sub slurp { shift->{content} // '' }
 
 sub to_file { Mojo::Asset::File->new->add_chunk(shift->slurp) }
 
+sub _end_range {
+  my $self = shift;
+  return undef unless $self->is_range;
+  if (my $end = $self->end_range) { return $end }
+  return (shift // 0) + length($self->{content} // '');
```
jhthorsen, line 58: I suggest always passing in a value, since _end_range() is only used internally.

lib/Mojo/Asset/File.pm
```diff
@@ -49,7 +49,7 @@ sub contains {
   $handle->sysseek($self->start_range, SEEK_SET);
 
   # Calculate window size
-  my $end  = $self->end_range // $self->size;
+  my $end  = $self->_end_range // $self->size;
```
jhthorsen, line 52: Could _end_range() return $self->size instead of undef?

lib/Mojo/Asset/File.pm
```diff
@@ -128,6 +133,13 @@ sub slurp {
 
 sub to_file {shift}
 
+sub _end_range {
+  my $file = shift;
+  return undef unless $file->is_range;
+  if (my $end = $file->end_range) { return $end }
+  return -1 + -s $file->handle;
```
jhthorsen, line 140: I think this could be golfed down to:
```
return $file->is_range && $file->end_range || (-s $file->handle) - 1;
```

jhthorsen (dismissed):

lib/Mojo/Asset/File.pm
kiwiroy: Agreed.

lib/Mojo/Asset/File.pm
kiwiroy: Good idea and makes the golf easier

lib/Mojo/Asset/File.pm
kiwiroy: golfed.

lib/Mojo/Asset/File.pm
```diff
@@ -116,7 +116,12 @@ sub new {
   return $file;
 }
 
-sub size { -s shift->handle }
+sub size {
+  my $self = shift;
+  my $size = -s $self->handle;
+  return $size unless $self->is_range;
```
jhthorsen, line 122: Shouldn't this be `return -s $self->handle unless $self->is_range`?

lib/Mojo/Asset/File.pm
```diff
@@ -128,6 +133,11 @@ sub slurp {
 
 sub to_file {shift}
 
+sub _end_range {
+  my $self = shift;
+  return (!$self->is_range && -s $self->handle) || $self->end_range || (-s $self->handle) - 1;
```
jhthorsen, line 138: I don't think my idea was very good if the size of the file is zero (0)?

lib/Mojo/Asset/Memory.pm
```diff
@@ -41,12 +41,21 @@ sub get_chunk {
...
+}
 
 sub slurp { shift->{content} // '' }
 
 sub to_file { Mojo::Asset::File->new->add_chunk(shift->slurp) }
 
+sub _end_range {
+  my $self = shift;
+  return undef unless $self->is_range;
+  return $self->end_range || shift() + length($self->{content} // '');
```
jhthorsen, line 57: Might be a personal thing, but having shift() there made me very confused.

kraih: I think `shift()` there is bad style, unless it's absolutely necessary for performance.

kraih:
Is this PR abandoned?

kraih:
I suppose it is.

PR #1883, opened 2021-11-17
Fix potential core dump during worker process shutdown

### LINKED ISSUE #1449: Hypnotoad / Prefork dump cores on small values of accepts and disabled keep-alive
* Mojolicious version: 8.27
* Perl version: 5.30.1
* Operating system: Linux 5.3.13-arch1-1

*It's a TL;DR version, scroll down and see [attached archive](https://github.com/mojolicious/mojo/files/3951190/repro.tar.gz) for code and more details*
I run a simple synchronous server

    get '/' => sub {
        usleep 1_000_000 * rand;
        shift->render(text => "i'm ok");
    };

with a command

    ./server.pl prefork --listen 'http://*:3002' \
      --accepts 1 \
      --clients 1 \
      --requests 1 \
      --workers 10

and a client in a second terminal to generate some load.

    ./client.pl

### Expected behavior
I expect each worker process to exit normally after it served one request.

### Actual behavior
Eventually a core dump of a worker process is generated. `gdb` shows `Program terminated with signal SIGQUIT, Quit.`.

### Setting up a clean environment
I use a Linux box, [plenv](https://github.com/tokuhirom/plenv) to compile recent Perl version with debug symbols and [cpanm](https://metacpan.org/pod/distribution/App-cpanminus/bin/cpanm) to install Mojolicious locally.

    plenv install --as=5.30.1-debugging 5.30.1 -DDEBUGGING
    plenv local 5.30.1-debugging
    plenv install-cpanm
    cpanm -L local Mojolicious@8.27 --quiet

Then make sure core dumps will end up in a current working directory (you may want to back up previous value but it won't persist across system boots). It's not a necessary step but `client.pl` and `gdb` commands here assume it's done.

    echo 'core' | sudo tee /proc/sys/kernel/core_pattern

Start server

    PERL5LIB=local/lib/perl5 ./server.pl prefork --listen 'http://*:3002' \
      --accepts 1 \
      --clients 1 \
      --requests 1 \
      --workers 10

and client from another terminal.

    PERL5LIB=local/lib/perl5 ./client.pl

It usually takes up to 15 minutes to reproduce on my Intel i5-2500K box. After client terminates there should be 1 to 3 `core.<pid>` files in a current working directory.

If a core dump was generated by e.g. a process with pid 33156, it's backtrace can be inspected with the following command:

    gdb -ex bt ~/.plenv/versions/5.30.1-debugging/bin/perl core.33156

To observe signals-related behaviour I run server via [strace](https://strace.io/).
_It generates a lot of `traces/trace.<pid>` files which could take up some disk space._

    mkdir traces
    PERL5LIB=local/lib/perl5 strace -o traces/trace -ff -e signal,write -s 128 -tt -- ./server.pl prefork \
      --listen 'http://*:3002' \
      --accepts 1 \
      --clients 1 \
      --requests 1 \
      --workers 10

### Dumps and traces

According to backtraces it seems processes usually terminate inside `Perl_pp_exit` or `perl_destruct` functions. In a syscall traces of dumped processes `SIGQUIT` arrives some microseconds after `SIGQUIT` handler is restored to `SIG_DFL` (a default one).

I assume there's a race condition due to a time gap between signal handlers are reset by Perl and process terminates.

### Potential fix

See `server-fixed.pl` in [attached archive](https://github.com/mojolicious/mojo/files/3951190/repro.tar.gz). This seems to do the trick but race condition still remains (AFAIK a signal may arrive between `rt_sigaction` setting `SIGQUIT` handler to `SIG_DFL` and `rt_sigprocmask` blocking the process from recieving `SIGQUIT`).

	app->hook(before_server_start => sub {
		my ($server, $app) = @_;

		my $sigset = POSIX::SigSet->new(SIGQUIT);
		$server->ioloop->on(finish => sub {
			my $loop = shift;

			sigprocmask(SIG_BLOCK, $sigset) // die "Could not block SIGQUIT\n";
		});
	});

jixam:
### Summary

Remove redundant `stop_gracefully ` after `finish`.

### Motivation

When reaching `max_accepts`, `IOLoop` will
1. run a `finish` callback on the `Prefork` worker which will
2. send a `finished=1` heartbeat on a socket which will
3. cause the `Prefork` manager to `SIGQUIT` the worker which will
4. have the worker call `stop_gracefully` on the `IOLoop`

This seems redundant (and slightly circular) as the `IOLoop` is already in the process of stopping. It is also a small bug because the worker process could have uninstalled its `local $SIG{QUIT}` handler between step 2 and step 3 and the signal will then cause a core dump (#1449).

I am a bit wary of suggesting this but I could not find anything else happening due to that `finish` callback so I propose to break the cycle already at the first step. 

(Most of the PR is cleaning up the heartbeat protocol now that `finished` is no longer used.)

### References

This fixes #1449 which is a rare but real problem.

### DIFF
--- a/lib/Mojo/Server/Prefork.pm
+++ b/lib/Mojo/Server/Prefork.pm
@@ -79,7 +79,7 @@ sub run {
   $self->app->log->info("Manager $$ stopped");
 }
 
-sub _heartbeat { shift->{writer}->syswrite("$$:$_[0]\n") or exit 0 }
+sub _heartbeat { shift->{writer}->syswrite("$$\n") or exit 0 }
 
 sub _manage {
   my $self = shift;
@@ -133,11 +133,9 @@ sub _spawn {
   return $self->emit(spawn => $pid)->{pool}{$pid} = {time => steady_time} if $pid;
 
   # Heartbeat messages
-  my $loop     = $self->cleanup(0)->ioloop;
-  my $finished = 0;
-  $loop->on(finish => sub { $finished = 1 });
+  my $loop = $self->cleanup(0)->ioloop;
   weaken $self;
-  my $cb = sub { $self->_heartbeat($finished) };
+  my $cb = sub { $self->_heartbeat() };
   $loop->next_tick($cb);
   $loop->recurring($self->heartbeat_interval => $cb);
 
@@ -176,12 +174,11 @@ sub _wait {
   return unless Mojo::Util::_readable(1000, fileno($reader));
   return unless $reader->sysread(my $chunk, 4194304);
 
-  # Update heartbeats (and stop gracefully if necessary)
+  # Update heartbeats
   my $time = steady_time;
-  while ($chunk =~ /(\d+):(\d)\n/g) {
+  while ($chunk =~ /(\d+)\n/g) {
     next unless my $w = $self->{pool}{$1};
     @$w{qw(healthy time)} = (1, $time) and $self->emit(heartbeat => $1);
-    $w->{graceful} ||= $time if $2;
   }
 }
 


kraih:
I'm a big fan of simplifying code, but i'm also worried about losing important functionality. This PR needs to be double checked.

kraih:
This PR needs to be rebased.

jixam:
Thanks, I will just close it as nobody seems interested in reviewing.

kraih:
That's the reason i was asking for it to be rebased, if tests don't work nobody will review. 🙄

jixam:
All tests pass and there are no conflicts so I just reopened 🤷

kraih:
The PR would leave the documentation incorrect at least, since the worker does not go into graceful shutdown mode anymore.

rabbiveesh:
It sounds to me like we wouldn't want this change if it can end up with impossible-to-debug zombie processes.
As an app developer, I want to be able to assume that when I shutdown my app, it will shutdown in all instances

jixam:
I didn't consider that a manual `SIGQUIT` is on the same code path so I agree that this PR is a no-go as is. However, I am unsure how to change it.

Please consider this skeleton app:
```
perl -Mojo -E 'get "/" => sub ($c) { $c->inactivity_timeout(0); $c->render_later; }; app->start' prefork -a 100
```

It seems wrong to close all requests here just because `max_accepts` is reached. I think that is a current bug?

But for `SIGQUIT`, yes – the timeout should obviously apply.

jixam:
I am not going to update this PR as I think the fix by @brsakai-csco is better and my concern above is a separate issue.

jixam:
Closing in favor of #2073.

PR #1887, opened 2021-12-05
Add an inline_layout stash keyword, based on the mojo.js inlineLayout one.

mkende:
### Summary
This adds a way to pass an inline layout in calls to render() using a new `inline_layout` keyword.

### Motivation
Using the mojo.js semantics to allow inline layout seems to be the safest way.

### References
See the discussion in #1780.


lib/Mojolicious/Renderer.pm
```diff
@@ -214,6 +216,15 @@ sub _next {
   return join '/', 'layouts', $layout;
 }
 
+sub _render_content {
+  my ($self, $c, $next, $options, $output) = @_;
+  my $stash = $c->stash;
```
kraih, line 221: This call seems a bit wasteful.

lib/Mojolicious/Guides/Rendering.pod
```diff
@@ -530,6 +530,10 @@ as a render argument (not a stash value).
 
   my $html = $c->render_to_string('reminder', layout => 'mail');
 
+Layouts can also be passed inline, using the C<inline_layout> value.
+
+  my $text = $c->render(inline => 'text', inline_layout => '<%== content %>');
+
```
kraih, line 536:
```
For renderers that support C<inline> templates, such as C<ep>, you can also pass a layout with the C<inline_layout> stash value.

  $c->render(inline => 'World', inline_layout => 'Hello <%= content %>!');
```

lib/Mojolicious/Renderer.pm
mkende: Removed.

lib/Mojolicious/Guides/Rendering.pod
mkende: Updated as per your comment.

mkende:
This shoud be ready for another round of review at your convenience (I’m not sure what notification is sent by Github).

marcusramberg (dismissed):
This looks like an useful addition to me 👍🏻

lib/Mojolicious/Renderer.pm
```diff
@@ -214,6 +220,15 @@ sub _next {
   return join '/', 'layouts', $layout;
 }
 
+sub _render_content {
```
kraih, line 223: Is this method just dead code? Seems like it's never called.

kraih (changes requested):

lib/Mojolicious/Renderer.pm
```diff
@@ -108,6 +108,12 @@ sub render {
   # Inheritance
   my $content = $stash->{'mojo.content'} //= {};
   local $content->{content} = $output =~ /\S/ ? $output : undef if $stash->{extends} || $stash->{layout};
+  if ($stash->{inline_layout}) {
+    @$options{qw(template inline)} = (undef, delete $stash->{inline_layout});
+    $options->{format} = $stash->{format} || $self->default_format;
```
kraih, line 113: I don't think this line has any test coverage.

lib/Mojolicious/Renderer.pm
```diff
@@ -108,6 +108,12 @@ sub render {
...
+    if ($self->_render_template($c, \my $tmp, $options)) { $output = $tmp }
+    $content->{content} //= $output if $output =~ /\S/;
```
kraih, line 115: Don't think this line has test coverage either.

lib/Mojolicious/Renderer.pm
```diff
@@ -108,6 +108,12 @@ sub render {
   # Inheritance
   my $content = $stash->{'mojo.content'} //= {};
   local $content->{content} = $output =~ /\S/ ? $output : undef if $stash->{extends} || $stash->{layout};
+  if ($stash->{inline_layout}) {
+    @$options{qw(template inline)} = (undef, delete $stash->{inline_layout});
```
kraih, line 112: Setting the `template` value to `undef` seems not to have test coverage.

lib/Mojolicious/Renderer.pm
mkende: Right, it’s no longer used.

lib/Mojolicious/Renderer.pm
mkende: I’ve removed that line as there is no way to specify the layout format. I suppose that we could introduce a new keyword for that, but it’s probably not worth it.

lib/Mojolicious/Renderer.pm
mkende: Removed too. If I’m not mistaken this is useful only in the event that the processing continues after that. Which can’t happen for now.

mkende:
Thanks, please take another look.

(I’m not sure why the CI is failing, the messages seem unrelated to the change as they are about some docker initialisation error)

mkende:
Any chance to get a new review? I believe that I have addressed all your comment and that this might be ready to be merged.

lib/Mojolicious/Renderer.pm
```diff
@@ -108,6 +108,10 @@ sub render {
   # Inheritance
   my $content = $stash->{'mojo.content'} //= {};
   local $content->{content} = $output =~ /\S/ ? $output : undef if $stash->{extends} || $stash->{layout};
+  if ($stash->{inline_layout}) {
+    @$options{inline} = delete $stash->{inline_layout};
```
kraih, line 112: `@$options` doesn't make sense here.

kraih: I can't quite say what yet, but something about the logic seems wrong. 🤔

lib/Mojolicious/Renderer.pm
```diff
@@ -108,6 +108,10 @@ sub render {
   # Inheritance
   my $content = $stash->{'mojo.content'} //= {};
   local $content->{content} = $output =~ /\S/ ? $output : undef if $stash->{extends} || $stash->{layout};
+  if ($stash->{inline_layout}) {
```
kraih, line 111: `if (defined...`.

lib/Mojolicious/Renderer.pm
kraih: Maybe that `inline` value needs to be `local`-ised to avoid side effects?

mkende: You’re right that it should be possible to execute another layout defined in the inline layout (if we started from a non-inline template).

I localized the inline value and fixed another bug (a layout and inline_layout defined initially could have both been executed), and I observed that it prevented entering the second loop. However, I was not able to get it to work in the correct case (template => inline_layout => other layout defined in the inline_layout), so I was not able either to write a good test case for the fixed part.

mkende:
Let me know if there are still issues to address here.

mkende:
If there are no more comments on this PR, could it be merged?

lib/Mojolicious/Renderer.pm
```diff
@@ -107,7 +107,14 @@ sub render {
 
   # Inheritance
   my $content = $stash->{'mojo.content'} //= {};
-  local $content->{content} = $output =~ /\S/ ? $output : undef if $stash->{extends} || $stash->{layout};
+  local $content->{content} = $output =~ /\S/ ? $output : undef
+    if $stash->{extends} || $stash->{layout} || $stash->{inline_layout};
+  if (defined $stash->{inline_layout}) {
+    local $options->{inline} = delete $stash->{inline_layout};
+    delete $stash->{layout};
```
kraih, line 114: Do the tests still pass if this line is removed?

mkende: They do. I tried to write something to test this line but did not succeed (setting `layout` and `inline_layout` together in a call to render results in an error 500 but I’m not sure why). I still have the impression that this line should stay.

Do you have an idea of how to test it? Or do you think that it should be removed?

kraih: Code that does nothing should not be present.

kraih:
Too bad this PR is not ready, i would have liked to ship it in the next release.

PR #1897, opened 2021-12-21
Finish rendering early in Mojolicious::dispatch

andrii-suse:
### Summary
If rendering happened in before_dispatch (e.g. dynamic paths were processed), no need to try to render with static or find a route.

### Motivation
If before_dispatch sets 'rendered' flag, further actions around static and routes not only waste of resources, but may also finish rendering, which will suppress delayed rendering from before_dispatch

### References
#1534

### DIFF
--- a/lib/Mojolicious.pm
+++ b/lib/Mojolicious.pm
@@ -105,12 +105,14 @@ sub dispatch {
 
   my $plugins = $self->plugins->emit_hook(before_dispatch => $c);
 
+  my $stash = $c->stash;
+  return if $stash->{'mojo.rendered'};
+
   # Try to find a static file
   my $tx = $c->tx;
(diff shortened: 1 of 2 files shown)


jhthorsen (dismissed):
This change makes sense to me, but it won't get accepted without any tests.

andrii-suse:
> This change makes sense to me, but it won't get accepted without any tests.

Makes sense. I've added a static file with the same name as in the delayed handler from one of before_dispatch hooks and that test fails without the changes proposed for Mojolicious.pm .

jhthorsen (commented):
Shouldn't there be a test that checks that "before_dispatch" renders something and therefor dispatch returns? I'm not sure if I'm missing out on something here...

andrii-suse:
> Shouldn't there be a test that checks that "before_dispatch" renders something and therefor dispatch returns?

dispatch doesn't return immediately in such case - it still shall call `$c->helpers->log->trace`.

PR #1901, opened 2022-01-07
Complex form multipart example

PaulGWebster:
### Summary
A minor addition to the examples/synopsis

### Motivation
I found it really difficult to get multipart working

### References
No references

### Problems
There is no way to make a working example of this that will just run, unless anyone knows a site anywhere that would accept such a payload, By the way I actually expect this pull to be denied, but I want to raise that a form multipart example should be included!

### Request
Also I quite literally guessed how to do the 'file1' that is that it would be 'filename' for what to name the file and what not, I could not find the documentation for it, so if anyone knows where the documentation for that section is, please let me know :)

### DIFF
--- a/lib/Mojo/UserAgent.pm
+++ b/lib/Mojo/UserAgent.pm
@@ -392,6 +392,24 @@ Mojo::UserAgent - Non-blocking I/O HTTP and WebSocket user agent
   # Form POST (application/x-www-form-urlencoded)
   my $tx = $ua->post('https://metacpan.org/search' => form => {q => 'mojo'});
 
+  # Form multipart POST with both a local file and document within a string, accepting a JSON return.
+  my $tx = $ua->post(
+    'https://example.com/'
+    => {'Accept' => 'application/json'}
+    => form 
+    => {
+      'file1' => {
+        'content'   =>  '<?xml version="1.0"?><P ID="greeting">Hi!</P>',
+        'type'      =>  'text/xml',
+        'filename'  =>  'submit.xml'
+      }
+    }
+    => {
+     'file2' => {
+        'file'      =>  '/some/file/somewhere.xml'
+      }
+  );
+
   # Search DuckDuckGo anonymously through Tor
   $ua->proxy->http('socks://127.0.0.1:9050');
   say $ua->get('api.3g2upl4pq6kufc4m.onion/?q=mojolicious&format=json')->result->json('/Abstract');


kraih:
Always squash your commits.

lib/Mojo/UserAgent.pm
```diff
@@ -392,6 +392,24 @@ Mojo::UserAgent - Non-blocking I/O HTTP and WebSocket user agent
   # Form POST (application/x-www-form-urlencoded)
   my $tx = $ua->post('https://metacpan.org/search' => form => {q => 'mojo'});
 
+  # Form multipart post with both a local file and document within a string, accepting a json return
```
kraih, line 395: Lower case `post` and `json`? 🤨

kraih: `...accepting a JSON response`.

lib/Mojo/UserAgent.pm
```diff
@@ -392,6 +392,24 @@ Mojo::UserAgent - Non-blocking I/O HTTP and WebSocket user agent
...
+  my $tx = $ua->post(
+    "https://example.com/"
```
kraih, line 397: Inconsistent quotes.

kraih:
👎 from me i'm afraid. Not opposed to having more forma examples, but this doesn't look clean.

lib/Mojo/UserAgent.pm
```diff
@@ -392,6 +392,24 @@ Mojo::UserAgent - Non-blocking I/O HTTP and WebSocket user agent
...
+        'content'   =>  '<?xml version="1.0"?><P ID="greeting">Hi!</P>',
+        'type'      =>  'text/xml',
+        'filename'  =>  'submit.xml'
+      }
+    }
+    => {
+     'file2' => {
+        'file'      =>  '/some/file/somewhere.xml'
+      }
+  );
```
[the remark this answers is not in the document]
PaulGWebster: What would the standard formatting look like so I may update it

lib/Mojo/UserAgent.pm
PaulGWebster: Updated

lib/Mojo/UserAgent.pm
PaulGWebster: Updated

PaulGWebster:
> 👎 from me i'm afraid. Not opposed to having more forma examples, but this doesn't look clean.

I updated all but one of your exceptions, the one I have not updated required your input

kathackeray (changes requested):
@PaulGWebster

Your example code has improperly nested brackets (so throws an error when run).  I've slightly reformatted and commented your example to demonstrate:

```perl
my $tx = $ua->post(
  'https://example.com/' => {'Accept' => 'application/json'} => form =>
  {
    'file1' => {
        'content'   =>  '<?xml version="1.0"?><P ID="greeting">Hi!</P>',
        'type'      =>  'text/xml',
        'filename'  =>  'submit.xml'
    }
  } =>
  {
    'file2' => { # <-------------------------------------- whoops!!
      'file'      =>  '/some/file/somewhere.xml'
  }
);
```

As for the formatting, I've compared it to the other examples in the synopsis and the docs for [Mojo::UserAgent::Transactor](https://docs.mojolicious.org/Mojo/UserAgent/Transactor#tx), and I've had a go at reformatting your example below to see if it is more in keeping with expectations.  Note that I have removed the `Accept` header, as this seems to distract from the purpose of the example.  Also I have combined both parts into a single hashref.

```perl
# Form multipart POST with both a local file upload and document within a string
my $tx = $ua->post('http://localhost:3000/' => form => {
    file_upload => {
      file      => 'test-file.txt',
    },
    file_string => {
      content   => '<?xml version="1.0"?><p id="greeting">How do!</p>',
      type      => 'text/xml',
      filename  => 'submit.xml'
    }
  }
);
```

This doesn't have the project's perltidy rules applied, but I've noticed that other examples in the synopsis appear to have prioritised clarity over adherence to those rules.  If perltidy conformance is required for the example, it would look like this:

```perl
# Form multipart POST with both a local file upload and document within a string
my $tx = $ua->post(
  'http://localhost:3000/' => form => {
    file_upload => {file => 'test-file.txt',},
    file_string =>
      {content => '<?xml version="1.0"?><p id="greeting">How do!</p>', type => 'text/xml', filename => 'submit.xml'}
  }
);
```

I tested it against a local `netcat` server:

```shell
$ nc -kdl localhost 3000
POST / HTTP/1.1
Content-Type: multipart/form-data; boundary=FAnFX
Accept-Encoding: gzip
User-Agent: Mojolicious (Perl)
Host: localhost:3000
Content-Length: 268

--FAnFX
type: text/xml
Content-Disposition: form-data; name="file_string"; filename="submit.xml"

<?xml version="1.0"?><p id="greeting">How do!</p>
--FAnFX
Content-Disposition: form-data; name="file_upload"; filename="test-file.txt"

Hello there!

--FAnFX--
```

I hope this helps.

PR #1914, opened 2022-02-11
Use subtests t/mojo/template.t; issue #1520

tschaefer:
### Summary
Convert t/mojo/template.t to using subtests.

### Motivation
See issue [#1520](https://github.com/mojolicious/mojo/issues/1520)


kathackeray (approved):
Tests pass. Same number of total tests/subtests after as before.  Tests remain the same.

The `$capture` string variable is duplicated within 13 subtests with identical content for all.

```
$ ack 'my \$capture' t/mojo/template.t
  my $capture = 'no warnings "redefine"; sub capture { shift->(@_) }';
  my $capture = 'no warnings "redefine"; sub capture { shift->(@_) }';
  my $capture = 'no warnings "redefine"; sub capture { shift->(@_) }';
  ...
```

Arguably it could be declared once and reused.  It doesn't seem to be subtest-specific and isn't modified.  I'm approving the PR because the code is correct and functional.  If somebody else fails it on the repetition, I'll know to be stricter next time.

kraih:
> Arguably it could be declared once and reused. It doesn't seem to be subtest-specific and isn't modified.

In tests it is more important to keep the scenarios isolated than to reduce duplicated code.

tschaefer:
Sorry for the confusion. The close was not on purpose.

I just wanted to ask, if I should fit the indentation manually blamed in the perltidy check? The local run of perltidy with the given .perltidyrc doesn't bring the "wanted" result.

kathackeray:
I have version `v20211029` of `perltidy` locally, and running it on that file does produce the indentation changes.  might your version be older?

tschaefer:
I'm using `v20220217` same as in the related check. It's working from the command line not from within vim, seems some setup stuff from my side. Anyways, I'll force-push the commit.

t/mojo/template.t
```diff
@@ -20,170 +20,202 @@ sub exception { die 'ohoh' }
...
+  is $output, "lalala <%= 1 +\n 1 %> 12\n34\n", 'expression tag has been replaced';
+};
 
-# Replace comment tag
-$mt     = Mojo::Template->new;
-$output = $mt->render('<%%# 1 + 1 %>');
-is $output, "<%# 1 + 1 %>\n", 'comment tag has been replaced';
+subtest 'Replace comment tag' => sub {
+  my $mt     = Mojo::Template->new;
+  my $output = $mt->render('<%%# \'1 + 1 %>');
```
kathackeray, line 152: Addition of `'` to a test input and expected response.  Test still passes, but unsure of the reason.

t/mojo/template.t
```diff
@@ -20,170 +20,202 @@ sub exception { die 'ohoh' }
...
+  is $output, "%= 1 + 1\n", 'expression line has been replaced';
+};
 
-# Replace comment line
-$mt     = Mojo::Template->new;
-$output = $mt->render('  %%# 1 + 1');
-is $output, "  %# 1 + 1\n", 'comment line has been replaced';
+subtest 'Replace comment line' => sub {
+  my $mt     = Mojo::Template->new;
+  my $output = $mt->render('  %%# \'1 + 1');
```
kathackeray, line 176: Addition of `'` to a test input and expected response. Test still passes, but unsure of the reason.

kathackeray (approved):
Tests pass. Same number of total tests / subtests after as before.  Perltidy passes.

Tests remain _largely_ the same.  However, I have highlighted a couple of small changes to the test input and expected responses, the reason for which is not explained in the issue as far as I can tell.  I doubt this constitutes a fail.

t/mojo/template.t
tschaefer: I can undo this unintentional *adds*. To be honest, to do this restructure of the tests is a kind of monkey job. So I tried to do it with some *clever* regex substitutions, seems that went not that well ...

t/mojo/template.t
```diff
@@ -793,293 +871,325 @@ bar
...
+  my $mt = Mojo::Template->new;
+  $mt->append('$_O = "FOO!"');
+  my $output = $mt->render('23');
+  is $output, "FOO!", 'appending code';
+};
+
+subtest 'Multiline comment' => sub {
+  my $mt     = Mojo::Template->new;
+  my $output = $mt->render(<<'EOF');
+<html><%# 'this is
```
kathackeray, line 1038: Another stray single quote

kathackeray (changes requested):

Issue #1905, opened 2022-01-18
Generating Dockerfile refers to wrong script name


toreau:
If you do this:

    mojo generate app MyApp::Something

The following script is created:

    script/my_app_something

But if you generate a Dockerfile using this:

    script/my_app_something generate dockerfile

The Dockerfile will refer to a wrongly named `CMD` script (and also `WORKDIR`), ref. `my_app_something` vs. `my_app-something`:

    $ cat Dockerfile 
    FROM perl
    WORKDIR /opt/my_app-something
    COPY . .
    RUN cpanm --installdeps -n .
    EXPOSE 3000
    CMD ./script/my_app-something prefork

tianon:
Hmm, that's strange; here's the code responsible for that value:

https://github.com/mojolicious/mojo/blob/4093223cae00eb516e38f2226749d2963597cca3/lib/Mojolicious/Command/Author/generate/dockerfile.pm#L11-L12

It's generated either from `MOJO_EXE` or `$self->app->moniker`, and I'm a little surprised by this. :grimacing:

I guess it maybe needs to use `class_to_file` instead somehow, like `mojo generate app` does?
(I'm grasping at straws here :innocent:)

https://github.com/mojolicious/mojo/blob/4093223cae00eb516e38f2226749d2963597cca3/lib/Mojolicious/Command/Author/generate/app.pm#L13

Or perhaps this is technically a bug in `moniker`?

kathackeray:
The name `my_app_something` for `MyApp::Something` is ambiguous (it would be the same for `My::AppSomething`).  Generating app names in the form `my_app-something` (as the dockerfile generation assumes) would seem to be the superior solution.

Is such a solution acceptable, or is the naming for app generation constrained by some need for backward compability?

kathackeray:
The `decamelize()` function in `Mojo::Util` handles transforming to `-` and `_` appropriately.

However, the `class_to_file()` function in `Mojo::Util` removes occurrences of `::` before handing off to `decamelize()`.  

```
sub class_to_file {
  my $class = shift;
  $class =~ s/::|'//g;
  $class =~ s/([A-Z])([A-Z]*)/$1 . lc $2/ge;
  return decamelize($class);
}
```

This ensures that `decamelize()` never receives any instances of `::` to convert into `-`, resulting in `_` only.  There may be a reason for this that I haven't anticipated;  no comment in the code.

[...]

PR #1918, opened 2022-02-21
Ensure naming of files during app and dockerfile generation is consistent

kathackeray:
### Summary

* Use `Mojo::Util::class_to_file()` for file names generated from class names
* Update documentation

Currently, if generating an app named `MyApp::HTMLBuilder` and its dockerfile, the following inconsistent naming is formed:

```perl
$ mojo generate app MyApp::HTMLBuilder
$ cd my_app_htmlbuilder && ./script/my_app_htmlbuilder generate dockerfile
```

Name                   | Translation                  |Uses
 ---                   | ---                          | ---
App directory          | my_app_htmlbuilder           | `class_to_file()`
App script name        | my_app_htmlbuilder           | `class_to_file()`
App config name        | my_app-h_t_m_l_builder.yml   | `decamelize()`
Dockerfile content refs | my_app-h_t_m_l_builder       | `decamelize()`

This PR ensures only that `class_to_file()` is used consistently.

#### Important - This change will break existing apps!

For an app generated before this change, the former, inconsistent naming of the files will be in place.

```shell
$ ./script/my_app_original_app
Configuration file "/home/kat/tmp/~/mojo-app-gen-test/my_app_original_app/my_app_original_app.yml" missing, maybe you need to create it?
```

Steps for fixing:

* Rename the config file from `my_app-original_app.yml` to `my_app_original_app.yml`
* Regenerate or edit existing Dockerfiles to the changed naming for config

### Motivation

The goal of this PR is to ensure that naming files based on class names is predictable, by utilising only the `class_to_file()` function for this purpose - not mixing between that and `decamelize()`.

Whilst there are other improvements to be made, I have left these to future PRs to avoid including too much change here.  Once usage is standardised, the following further improvements might be considered.

* Ensure generated dockerfiles work out-of-the-box
* Consider adopting generic naming within the app dir (suggested by @jberger in #1905)
* Modify `class_to_file()` to produce unambiguous and reversible file names
* Modify `decamelize()` to produce output that is reversible with `camelize()`
* Validate that class names passed to `class_to_file()` are valid

### References

#1905 - Generating Dockerfile refers to wrong script name

### DIFF
--- a/lib/Mojolicious.pm
+++ b/lib/Mojolicious.pm
@@ -35,7 +35,7 @@ has log              => sub {
 };
 has 'max_request_size';
 has mode               => sub { $ENV{MOJO_MODE} || $ENV{PLACK_ENV} || 'development' };
-has moniker            => sub { Mojo::Util::decamelize ref shift };
+has moniker            => sub { Mojo::Util::class_to_file ref shift };
 has plugins            => sub { Mojolicious::Plugins->new };
 has preload_namespaces => sub { [] };
 has renderer           => sub { Mojolicious::Renderer->new };
@@ -425,7 +425,7 @@ variables or C<development>.
   $app        = $app->moniker('foo_bar');
 
 Moniker of this application, often used as default filename for configuration files and the like, defaults to
-decamelizing the application class with L<Mojo::Util/"decamelize">.
+transforming the application class with L<Mojo::Util/"class_to_file">.
 
 =head2 plugins
 
--- a/lib/Mojolicious/Command/Author/generate/app.pm
+++ b/lib/Mojolicious/Command/Author/generate/app.pm
@@ -19,7 +19,7 @@ sub run {
   $self->render_to_rel_file('appclass', "$name/lib/$app", {class => $class});
 
   # Config file (using the default moniker)
-  $self->render_to_rel_file('config', "$name/@{[decamelize $class]}.yml");
+  $self->render_to_rel_file('config', "$name/$name.yml");
 
   # Controller
   my $controller = "${class}::Controller::Example";
--- a/t/mojo/daemon.t
+++ b/t/mojo/daemon.t
@@ -270,7 +270,7 @@ subtest 'Timeout' => sub {
 subtest 'Pipelined' => sub {
   my $daemon = Mojo::Server::Daemon->new({listen => ['http://127.0.0.1'], silent => 1});
   my $port   = $daemon->start->ports->[0];
-  is $daemon->app->moniker, 'mojo-hello_world', 'right moniker';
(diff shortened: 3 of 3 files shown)


lib/Mojolicious.pm
```diff
@@ -35,7 +35,7 @@ has log              => sub {
 };
 has 'max_request_size';
 has mode               => sub { $ENV{MOJO_MODE} || $ENV{PLACK_ENV} || 'development' };
-has moniker            => sub { Mojo::Util::decamelize ref shift };
+has moniker            => sub { Mojo::Util::class_to_file ref shift };
```
jberger, line 38: This is a breaking change unfortunately. This has implications about how configuration files are loaded by default. I'd be fine changing this at a major version or at least in line with our breaking changes rules (though I would personally rather this be at a major version)

kathackeray: Is there a process I should follow to put this forward for a major release?

kathackeray:
Sorry, trying to squash. Bear with me.

kathackeray:
This proposed first step to resolution of #1905 has failed review on account of a breaking change.  As any clean fix for this issue would necessitate a breaking change, I can't see a way to push this PR through.  As such I'm closing.

kraih:
This PR has not failed review. There was just one objection.

kathackeray:
Noted on the closure: I didn't know how to proceed and wanted to avoid clogging the PR list.  I'll be more patient.

tianon:
I'm not a maintainer, but IMO this is the cleanest solution to the problem - having two different ways to lowercase the app class name that both end up in the filesystem almost next to each other seems like it's worth making a breaking change next release (especially since I imagine the number of users affected is not huge, but even if it is, the discrepancy is kind of an eyesore).

lib/Mojolicious.pm
christopherraa: While I support your stance @jberger that this is a breaking change and thus should be planned for a major release, it is also a question of definition: is the proposed change a _change_ or a _bugfix_?

When looking over a couple of projects here I see that we usually end up overriding `moniker` because of how the default moniker is generated (see example table from the PR description). So from that perspective if this change was to be implemented we no longer would have to set `moniker` in our apps, because the problem with the auto-generated moniker would go away.

From that perspective I think it is fair to view this as a bugfix, and thus be considered for a minor-release.

christopherraa (approved):
This looks good to me. Though I would like to see others weigh with their opinions on this is considered a bug or an ordinary change. See my reply to the comment made by @jberger .

marcusramberg:
I'm really unsure what to think about this. I'm leaning towards jberger's view that we should fix this in a major release.

Issue #1790, opened 2021-06-01
Switch away from HMAC SHA1 for signed cookies


kraih:
It appears brute forcing HMAC SHA1 has become way too easy. So it is time for us to switch to something more expensive. HMAC SHA256 would seem reasonable. Thoughts?

kraih:
The SHA1 functions and methods should be deprecated too.

Grinnz:
+1 to both. Deprecating the sha1 functions will be a little disruptive, but those functions are literally just Digest::SHA wrappers, people can switch to those if it's not security-sensitive.

Grinnz:
I'll note that Mojo::Util still has md5 wrappers too though.

kraih:
Ok, i'll leave the deprecation out for now. But we still have an issue with session logouts sometimes being too short. Which makes brute force attacks too easy.

kraih:
The change to SHA256 is merged. https://github.com/mojolicious/mojo/compare/f577670a1e4c...3f10b6af0271

PR #1977, opened 2022-09-06
Include SHA-256 functions in Mojo::Util

duffee:
### Summary
Import SHA-256 functions from Digest::SHA into Mojo::Util

### Motivation
As of v9.19, signed cookies now use HMAC-SHA-256 instead of HMAC-SHA1.
Import the equivalent SHA-256 functions into **Mojo::Util** in line with the current available SHA-1 functions.

I tripped over this last week while migrating a module from Mojo 8 to 9.  Our test module was mocking the
signed cookie and when I reached for the updated `hmac_sha256_hex` function, it wasn't in **Mojo::Util**
and I was sad.  Was just going to raise an issue and thought that a PR was just as quick and explains the change
better than I could.

### References
 Switch away from HMAC SHA1 for signed cookies #1790

### DIFF
--- a/lib/Mojo/Util.pm
+++ b/lib/Mojo/Util.pm
@@ -4,7 +4,7 @@ use Mojo::Base -strict;
 use Carp           qw(carp croak);
 use Data::Dumper   ();
 use Digest::MD5    qw(md5 md5_hex);
-use Digest::SHA    qw(hmac_sha1_hex sha1 sha1_hex);
+use Digest::SHA    qw(hmac_sha1_hex sha1 sha1_hex hmac_sha256_hex sha256 sha256_hex);
 use Encode         qw(find_encoding);
 use Exporter       qw(import);
 use File::Basename qw(dirname);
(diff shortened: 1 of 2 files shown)


kraih:
We don't usually add functions to `Mojo::Util` that are not used inside the framework for something.

duffee:
I see that **Mojolicious::Controller** uses `Digest::SHA::hmac_sha256_hex` directly in the `every_signed_cookie` and `signed_cookie` subs and that you're planning to deprecate the old SHA1 functions.  Is that a design decision to remove digest code out of Mojo core or would you find it useful to migrate the current SHA1 functions to SHA-256?

kraih:
I think we'd rather remove those than add more. Their only use anymore is basically as methods of `Mojo::ByteStream`, in some limited situations. Happy to listen to counter arguments though.

Grinnz:
I think it would be a good idea to add these from the point of view of promoting good practices by easy availability, and SHA-256 is a rather commonly used hash at this point. And if so, they should be added to Mojo::ByteStream as well.

duffee:
I'd hate to argue against my own PR, but thinking some more, I'd actually prefer performance over convenience.  The business has recently moved into the cloud where memory and runtime now have a price tag on them.

I'm not a profiling guru.  If someone can determine whether monkey patching the sha-256 functions into Mojo::Util improves or hurts performance for code that uses them (as well as for code that doesn't), I would go with that.  If it's better, I'll volunteer to add them to Mojo::ByteStream as well.

kraih:
I'm afraid it looks like this PR failed to reach the required number of votes. Thank you.

PR #1978, opened 2022-09-08
Add support for static assets

kraih:
This PR adds the long discussed support for static assets generated by bundlers like Webpack. See the included documentation for details.

### DIFF
--- a/lib/Mojolicious/Controller.pm
+++ b/lib/Mojolicious/Controller.pm
@@ -272,6 +272,11 @@ sub url_for {
   return $url;
 }
 
+sub url_for_asset {
+  my ($self, $asset) = @_;
+  return $self->url_for($self->app->static->asset_path($asset));
+}
+
 sub write {
   my ($self, $chunk, $cb) = @_;
   $self->res->content->write($chunk, $cb ? sub { shift; $self->$cb(@_) } : ());
@@ -767,6 +772,13 @@ current request.
 
   # "/list?q=mojo&page=2" if current request was for "/list?q=mojo&page=1"
   $c->url_with->query({page => 2});
+
+
+=head2 url_for_asset
+
+  my $url = $c->url_for_asset('/app.js');
+
+Generate a portable L<Mojo::URL> object with base for a static asset.
 
 =head2 write
 
--- a/lib/Mojolicious/Guides/Rendering.pod
+++ b/lib/Mojolicious/Guides/Rendering.pod
@@ -332,6 +332,76 @@ ensure that they don't conflict with existing methods you may already have.
 A list of all built-in helpers can be found in L<Mojolicious::Plugin::DefaultHelpers> and
 L<Mojolicious::Plugin::TagHelpers>.
 
+=head2 Static files
+
+Static files are automatically served from the C<public> directories of the application, which can be customized with
+L<Mojolicious::Static/"paths">, or one of the C<DATA> sections from L<Mojolicious::Static/"classes">. And if that's not
+enough you can also serve them manually with L<Mojolicious::Plugin::DefaultHelpers/"reply-E<gt>static"> and
+L<Mojolicious::Plugin::DefaultHelpers/"reply-E<gt>file">.
+
+  use Mojolicious::Lite -signatures;
+
+  get '/' => sub ($c) {
+    $c->reply->static('index.html');
+  };
+
+  get '/some_download' => sub ($c) {
+    $c->res->headers->content_disposition('attachment; filename=bar.png;');
+    $c->reply->static('foo/bar.png');
+  };
+
+  get '/leak' => sub ($c) {
+    $c->reply->file('/etc/passwd');
+  };
+
+  app->start;
+
+=head2 Static assets
+
+While Mojolicious does not have any special support for frontend frameworks like L<Vue.js|https://vuejs.org> and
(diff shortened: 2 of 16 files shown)


kraih:
Same feature for mojo.js: https://github.com/mojolicious/mojo.js/pull/77

lib/Mojolicious/Controller.pm
```diff
@@ -272,6 +272,11 @@ sub url_for {
   return $url;
 }
 
+sub url_for_asset {
+  my ($self, $asset) = @_;
+  return $self->url_for($self->app->static->asset_path($asset));
+}
```
jhthorsen, line 278: Why is this a method and not a helper?

lib/Mojolicious/Controller.pm
```diff
@@ -768,6 +773,13 @@ current request.
   # "/list?q=mojo&page=2" if current request was for "/list?q=mojo&page=1"
   $c->url_with->query({page => 2});
 
+
+=head2 url_for_asset
+
+  my $url = $c->url_for_asset('/app.js');
```
jhthorsen, line 779: I think it would make sense to just use "app.js" and not "/app.js" in the example here, since it's relative to the assets directory.

lib/Mojolicious/Guides/Rendering.pod
```diff
@@ -332,6 +332,76 @@ ensure that they don't conflict with existing methods you may already have.
...
+  get '/leak' => sub ($c) {
+    $c->reply->file('/etc/passwd');
+  };
+
+  app->start;
+
+=head2 Static assets
+
+While Mojolicious does not have any special support for frontend frameworks like L<Vue.js|https://vuejs.org> and
+L<React|https://reactjs.org>, the C<public/assets> directory is reserved for static assets created by bundlers like
```
jhthorsen, line 362: Is the directory "reserved"..? I think the wording makes it sound like you can't use this for anything else.

lib/Mojolicious/Guides/Rendering.pod
```diff
@@ -332,6 +332,76 @@ ensure that they don't conflict with existing methods you may already have.
...
+L<Webpack|https://webpack.js.org> and L<Rollup.js|https://rollupjs.org> ahead of time. Asset files can be of any type,
+they just have to follow the C<[name].[checksum].[extensions]> naming scheme, like C<myapp.ab1234cd5678ef.js>. You can
+then use L<Mojolicious::Controller/"url_for_asset"> or L<Mojolicious::Plugin::TagHelpers/"assset_tag"> to generate URLs
```
jhthorsen, line 365: s/assset_tag/asset_tag/

lib/Mojolicious/Plugin/TagHelpers.pm
```diff
@@ -34,6 +34,16 @@ sub register {
   $app->helper($_ => sub { shift; _tag(@_) }) for qw(t tag);
 }
 
+sub _asset_tag {
+  my ($c, $target) = (shift, shift);
+
```
jhthorsen, line 39: Do you need an empty line here?

lib/Mojolicious/Controller.pm
jhthorsen, line 278: I'm surprised this is a method and not a helper.

lib/Mojolicious/Static.pm
```diff
@@ -34,12 +42,19 @@ sub dispatch {
...
+  $c->res->headers->cache_control('no-cache')
+    if $c->app->mode eq 'development' && index($canon_path, $self->asset_dir) == 0;
+
   return !!$c->rendered;
 }
 
 sub file {
   my ($self, $rel) = @_;
 
+  $self->warmup unless $self->{index};
```
jhthorsen, line 56: It would be nice if there was an environment variable or something that allows you to reload the assets while developing, in case there's a new file added by the bundler while the server is running. Example: A background image in a CSS-file that gets processed by the bundler.

jhthorsen (commented):
I was on vacation so didn't get to comment in time, but here is my review.

lib/Mojolicious/Controller.pm
kraih: Same reason `url_for` is a method, it's too essential to encourage overloading. Too many things will depend on it working a certain way.

lib/Mojolicious/Guides/Rendering.pod
kraih: It is "soft" reserved now. The app generator creates it and files in it have some special handling by the static file server.

lib/Mojolicious/Static.pm
kraih: I welcome improvements in this area.

kraih: A very easy way to do this would be to turn running `warmup` more than once into a feature.

kraih: As it happens, this already works for both, `$app->static->warmup` and `await app.static.warmup()`. There's just no tests yet. 😉

lib/Mojolicious/Controller.pm
kraih: One good example is plugins adding their own public directory to the static file server and then depending on the default behaviour of `url_for_asset`.

PR #1982, opened 2022-09-23
Auto set secure like true if samesite is like None

lucas1:
### Summary
Auto definition of **secure** when **samesite** equals None

### Motivation
Simplify using **samesite** as None

### References
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite#none


Grinnz:
I'm not sure whether it's the responsibility of the framework to do this without the user's request.

lucas1:
When **secure** is used will work only with HTTPS, **samesite** like None doesn't work without **secure**, so I don't see the problem with it.

PR #1985, opened 2022-10-06
Avoid warnings on questionable multipart/form-data

### LINKED ISSUE #1787: Warning on multipart/form-data request
* Mojolicious version: 9.17
* Perl version: 5.32.1
* Operating system: Ubuntu 21.04

While investigating some warning in our logs, I've noticed that Mojolicious will warn on questionable `multipart/form-data` that we have seen in the wild. For instance, given this script:

```perl

post '/echo' => sub ($c) {
  $c->param('does not matter');
  $c->render(data => $c->req->body);
};

app->start;
```
and this request:

```sh
curl -v -F '=' 'http://localhost:3000/echo'
```

the following warning will be generated:

```
Use of uninitialized value in string eq at Mojo/Parameters.pm line 42.
```

A similar but different warning can be generated from:

```sh
touch tmp
curl -v -F '=@tmp'  'http://localhost:3000/echo'
```

This causes the following:

```
Use of uninitialized value in hash element at /opt/perl5.32.1/lib/site_perl/5.32.1/Mojo/Message.pm line 240.
Mojo::Message::_cache at /opt/perl5.32.1/lib/site_perl/5.32.1/Mojo/Message.pm line 80
Mojo::Message::every_upload at /opt/perl5.32.1/lib/site_perl/5.32.1/Mojolicious/Controller.pm line 63
Mojolicious::Controller::every_param at /opt/perl5.32.1/lib/site_perl/5.32.1/Mojolicious/Controller.pm line 119
Mojolicious::Controller::param at /opt/perl5.32.1/lib/site_perl/5.32.1/Mojolicious/Plugin/MM/QuickParams.pm line 168
```

### Expected behavior

No warning.

### Actual behavior

The warnings mentioned above:

```
Use of uninitialized value in string eq at Mojo/Parameters.pm line 42.
```

and

```
Use of uninitialized value in hash element at Mojo/Message.pm line 240
```

jixam:
### Summary

Avoid storing form-data with malformed headers under an `undef` name. This removes a warning when trying to access the parameters:
```
Use of uninitialized value in string eq at Mojo/Parameters.pm line 42.
```

### Motivation

To remove monitoring alerts from the above warning.

### References

Fixes #1787

### DIFF
--- a/lib/Mojo/Message.pm
+++ b/lib/Mojo/Message.pm
@@ -267,6 +267,7 @@ sub _parse_formdata {
     my ($filename) = $disposition =~ /[; ]filename="((?:\\"|[^"])*)"/;
     next if $upload && !defined $filename || !$upload && defined $filename;
     my ($name) = $disposition =~ /[; ]name="((?:\\"|[^;"])*)"/;
+    next if !defined $name;
     $part = $part->asset->slurp unless $upload;
 
     if ($charset) {
--- a/t/mojo/request.t
+++ b/t/mojo/request.t
@@ -1003,6 +1003,35 @@ subtest 'Parse HTTP 1.1 multipart request with "0" filename' => sub {
   is $req->upload('upload')->asset->size, 69, 'right size';
 };
 
+subtest 'Parse HTTP 1.1 multipart request with missing name' => sub {
+  my $req = Mojo::Message::Request->new;
+  is $req->content->progress, 0, 'right progress';
+  $req->parse("GET /foo/bar.html HTTP/1.1\x0d\x0a");
+  is $req->content->progress, 0, 'right progress';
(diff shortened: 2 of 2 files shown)


kraih:
Looks fine. Not sure though if the value should really be discarded completely, or if the name should maybe default to something like the empty string.

jberger (approved):

jberger:
> Looks fine. Not sure though if the value should really be discarded completely, or if the name should maybe default to something like the empty string.

Depends on why it has no name I suppose. If it is just (slightly) malformed then probably best to continue as much as possible. If it thinks it is assuming that with only one value it doesn't need a name, i.e. "a default", then yeah, using an "empty" name might be better.

PR #1987, opened 2022-10-11
Avoid "Unhandled rejected promise warning" when using async/await

jhthorsen:
### Summary
This PR will avoid warning "Unhandled rejected promise warning" when using async/await and a promise gets rejected.

I think this change is correct after reading https://metacpan.org/pod/Future::AsyncAwait::Awaitable#AWAIT_DONE where it says "...result of an existing still-pending future", which means there's no reason to create new promise objects.

### Motivation
Less warnings.

### DIFF
--- a/lib/Mojo/Promise.pm
+++ b/lib/Mojo/Promise.pm
@@ -11,11 +11,9 @@ use constant DEBUG => $ENV{MOJO_PROMISE_DEBUG} || 0;
 has ioloop => sub { Mojo::IOLoop->singleton }, weak => 1;
 
 sub AWAIT_CHAIN_CANCEL { }
-
-sub AWAIT_CLONE { _await('clone', @_) }
-
-sub AWAIT_DONE { shift->resolve(@_) }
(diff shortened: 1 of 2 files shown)


jberger:
Without commenting on the actual change, the test confused me because in the absence of the `catch` being attached, `reject_without_warning` would actually emit a warning. It took me a minute to understand in the test. Could that be just `reject_after_timeout` and the comment about with or without warning could be in the test? I would have made it much easier for me to understand

jberger:
That's great @jhthorsen , thanks!

jberger:
Given the reading of Paul's documentation and that the test agrees with how I would expect the behavior to function, I generally approve this PR. However, I feel I should ask, is there any useful way to test the change made to AWAIT_DONE?

jhthorsen:
AWAIT_DONE() is tested implicitly, since all the other tests pass and it cannot result in a warning.

jberger (approved):

marcusramberg (approved):
Nice improvement! 🌮

PR #1991, opened 2022-10-15
Fix handling returned $promise from async method, closes #1990

### LINKED ISSUE #1990: Unexpected async/await promise behavior in 9.28
* Mojolicious version: 9.28
* Perl version: NA
* Operating system: NA

As explained on IRC, there has at least been a change of behavior, I think it qualifies as a bug as I think what I'm doing is allowed. It seems that in certain circumstances, when returning a promise from an async function, and that async function is awaited elsewhere, that promise can be returned from await, rather than the result of the promise.

My small scale replication is

```bash
perl -Mojo -MMojo::Base=-async_await -E 'async sub main { print await doit() } async sub doit { await Mojo::Promise->timer(1); return Mojo::Promise->resolve("hello") } main()->wait'
```

It should print "hello" but instead in Mojo 9.28 it prints the Mojo::Promise=HASH ... object

note that `return Mojo::Promise->resolve("hello")` does work, this is why I suspect that the optimization in 9.28 is possibly not handling thenables as it should, however, if `doit` doesn't have some other async calls in it then it does work too, which is why I wonder if there might be some race condition in Future::AsyncAwait and this optimization in mojo might have just revealed it by getting rid of an intermediate promise that masked it, therefore I'm tagging in @leonerd too

jhthorsen:
### Summary
Fix issue introduced in 166cedb244009e956bd8c3de16d86292bb0f59ee when an async method returns a promise.

### Motivation
@jberger reported the issue.

### References
166cedb244009e956bd8c3de16d86292bb0f59ee

### DIFF
--- a/lib/Mojo/Promise.pm
+++ b/lib/Mojo/Promise.pm
@@ -230,8 +230,10 @@ sub _settle {
 }
 
 sub _settle_await {
-  my ($status, $self) = (shift, shift);
-  @{$self}{qw(results status)} = ([@_], $status) if !$self->{results};
+  my ($status, $self, @results) = @_;
+  return $results[0]->then(sub { $self->resolve(@_); () }, sub { $self->reject(@_); () })
+    if blessed $results[0] && $results[0]->can('then');
+  @{$self}{qw(results status)} = ([@results], $status) if !$self->{results};
   $self->_defer;
 }
 
--- a/t/mojo/promise_async_await.t
+++ b/t/mojo/promise_async_await.t
@@ -79,6 +79,11 @@ get '/six' => sub {
 
 my $ua = Mojo::UserAgent->new(ioloop => Mojo::IOLoop->singleton);
 
+async sub await_then_return_promise {
+  my $method = shift;
+  return Mojo::Promise->resolve(await Mojo::Promise->$method(0.01, $method));
+}
+
 async sub reject_p {
   await Mojo::Promise->reject('Rejected promise');
 }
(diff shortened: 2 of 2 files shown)


t/mojo/promise_async_await.t
```diff
@@ -157,6 +162,15 @@ subtest 'Async function body returning a promise' => sub {
   is $text, 'value', 'right content';
 };
 
+subtest 'Async nested function returning a promise' => sub {
+  my $res;
+  await_then_return_promise('timer')->then(sub { $res = shift })->catch(sub { $res = shift })->wait;
+  is $res, 'timer', 'right content';
+
+  await_then_return_promise('timeout')->then(sub { $res = shift })->catch(sub { $res = shift })->wait;
+  like $res, qr{^timeout at}, 'right content';
```
kraih, line 171: Why not the standard `qr/.../`?

marcusramberg (dismissed):
I *think* i understand what this fixes. Head slightly hurts now tho. 🍻  needed to remedy this situation.

lib/Mojo/Promise.pm
```diff
@@ -231,6 +231,8 @@ sub _settle {
 
 sub _settle_await {
   my ($status, $self) = (shift, shift);
+  return $_[0]->then(sub { $self->resolve(@_); () }, sub { $self->reject(@_); () })
+    if blessed $_[0] && $_[0]->can('then');
```
kraih, line 235: At this point, just make it more readable, just like `_settle`.

t/mojo/promise_async_await.t
jhthorsen: No good reason. I personally have changed to `qr{...}` since I more often have "/" inside the regexp than "{", so it's just a matter of habit.

marcusramberg (approved):

jberger:
Bad news, running my example from #1990 on this branch the example hangs.

jberger (approved):
Yes, that fixes my example, thanks.

jberger:
I also agree about that change to the github test runner, please open a PR for that too.

marcusramberg (approved):
third time is the charm, clearly ;-)

PR #1992, opened 2022-10-15
Improved "at ... line ..." reporting for rejected async method

jhthorsen:
### Summary
Before this commit:
> Rejected promise at Mojo/Promise.pm line 23.

After this commit:
> Rejected promise at t/mojo/promise_async_await.t line 88.

### Motivation
I think it would be useful to see where the promise was rejected.

### DIFF
--- a/lib/Mojo/Promise.pm
+++ b/lib/Mojo/Promise.pm
@@ -1,7 +1,7 @@
 package Mojo::Promise;
 use Mojo::Base -base;
 
-use Carp qw(carp);
+use Carp qw(carp croak);
 use Mojo::Exception;
 use Mojo::IOLoop;
 use Scalar::Util qw(blessed);
@@ -18,8 +18,9 @@ sub AWAIT_FAIL         { _settle_await(reject  => @_) }
 sub AWAIT_GET {
   my $self    = shift;
   my @results = @{$self->{results} // []};
-  die $results[0] unless $self->{status} eq 'resolve';
-  return wantarray ? @results : $results[0];
+  return wantarray ? @results : $results[0] if $self->{status} eq 'resolve';
+  die $results[0]                           if ref $results[0] || $results[0] =~ m!\n!;
+  croak $results[0];
 }
 
 sub AWAIT_IS_CANCELLED {undef}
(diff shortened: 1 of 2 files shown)


t/mojo/promise_async_await.t
```diff
@@ -132,6 +132,12 @@ subtest 'Exception handling and async/await' => sub {
   $t->get_ok('/four')->status_is(500)->content_like(qr/this went perfectly/);
 };
 
+subtest 'Exception handling with file and line reporting' => sub {
+  my $error;
+  reject_p()->catch(sub { $error = shift })->wait;
+  like $error, qr{^Rejected promise at .*promise_async_await\.t line \d+\.$}, 'right content';
```
kraih, line 138: Just use the standard `qr/.../`.

lib/Mojo/Promise.pm
```diff
@@ -18,8 +18,9 @@ sub AWAIT_FAIL         { _settle_await(reject  => @_) }
 sub AWAIT_GET {
   my $self    = shift;
   my @results = @{$self->{results} // []};
-  die $results[0] unless $self->{status} eq 'resolve';
-  return wantarray ? @results : $results[0];
+  return wantarray ? @results : $results[0] if $self->{status} eq 'resolve';
+  die $results[0]                           if ref $results[0] || $results[0] =~ m!\n!;
+  croak $results[0];
```
kraih, line 23: Does this really need `die` and `croak`? Don't like so much logic just to throw an exception. Ternary would probably also look better here.

jhthorsen: Test::Future::AsyncAwait::Awaitable fails if I use croak and there's a newline in the message:

```
# Subtest: Mojo::Promise conforms to Awaitable API immediate fail
    not ok 5 - AWAIT_GET throws exception

    #   Failed test 'AWAIT_GET throws exception'
    #   at Test/Future/AsyncAwait/Awaitable.pm line 129.
    #          got: 'Oopsie
    #  at Test/Future/AsyncAwait/Awaitable.pm line 128.
    # '
    #     expected: 'Oopsie
    # '
    1..5
```

marcusramberg (approved):

kraih (approved):

PR #2000, opened 2022-11-01
add a possible controller in method 'to' if it's not an app

lucas1:
### Summary
This change to add a possible controller in method 'to' when it's not an app

### Motivation
When the value set is not an app, the mojo tests if it is a controller, but normally returns an error for not having a namespace defined.

When you write the code:

```
my $foo = $r->any('/foo')->to('foo');
$foo->get('/test')->to('#test');
```

Return the error:
Controller "foo" does not exist at Mojolicious.pm line 127.

This change solves the problem above.

### DIFF
--- a/lib/Mojolicious/Routes.pm
+++ b/lib/Mojolicious/Routes.pm
@@ -103,7 +103,12 @@ sub _class {
   # Application class
   my @classes;
   my $class = $field->{controller} ? camelize $field->{controller} : '';
-  if ($field->{app}) { push @classes, $field->{app} }
+  if ($field->{app}) { 
+    push @classes, $field->{app};
+    
+    # Maybe add a possible controller
+    push @classes, "${_}::" . camelize $field->{app} for @{$self->namespaces};
+  }
 
   # Specific namespace
   elsif (defined(my $ns = $field->{namespace})) {
--- a/t/mojolicious/routes.t
+++ b/t/mojolicious/routes.t
@@ -1023,4 +1023,22 @@ subtest 'Unknown placeholder type (matches nothing)' => sub {
   is_deeply $m->stack, [], 'empty stack';
 };
 
+subtest 'Intuitive controller (when it is not app)' => sub {
+  my $c = Mojolicious::Controller->new;
+
+  $r = Mojolicious::Routes->new;
+  my $panel = $r->any('/panel')->to('panel'); # not added # at the end
+  $panel->get('/users')->to('#users'); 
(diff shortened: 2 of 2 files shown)


jhthorsen:
What is wrong with `$r->any('/foo')->to('foo#');` ?

jberger:
Without a hash the shortcut form specifies a controller, the hash is implied at the end (like a url). If you don't like the bare fragment form that @jhthorsen has pointed out then you can use key/value as well.

As this PR would cause massive breakage to an understood api in going to assume core will vote this down and close it preemptively.

In the future a quick question on expected behavior could clarify before preparing a PR.

lucas1:
> What is wrong with `$r->any('/foo')->to('foo#');` ?

Hi, @jhthorsen.
No problem using it like this, only I forgot to put the # at the end. I thought it was a mistake.

lucas1:
> Without a hash the shortcut form specifies a controller, the hash is implied at the end (like a url). If you don't like the bare fragment form that @jhthorsen has pointed out then you can use key/value as well.
> 
> As this PR would cause massive breakage to an understood api I'm going to assume core will vote this down and close it preemptively.
> 
> In the future a quick question on expected behavior could clarify before preparing a PR.

Hi, @jberger.
I believe this change would not break anything, see the first value added to the array is:
`push @classes, $field->{app}`

My change continues adding this value first. So you will realize that the loop started on line 119 will return the value of **$field->{app}** if the same is true:
https://github.com/mojolicious/mojo/blob/main/lib/Mojolicious/Routes.pm#L119-L127

I added only a simple solution to a problem that I thought existed.

I will continue using 
`to('foo#');`
or I can use so:
`to('MyApp::Controller::Foo');`

Thank you and sorry to take your time.

kraih:
I think the tests might be too limited to really understand the intention here.

lucas1:
Hi @kraih.

I created some tests using and not using #:
https://github.com/lucas1/mojo/commit/82b2442ab1c8c184a0a7af0d9bd88010571a518c

After my changes, the code `$v1->any('/bar')->to('Vn::Bar');` has the same result as code `$v1->any('/bar')->to('Vn::Bar#');`.
I believe it is a good solution because the method _class tests if the value of  $field->{app} is an app, if false will test if is a controller, but normally the controllers are in the folder Controller, I only added the namespaces to try to find the controller in the folder Controller.

I added tests using another app, to show that the changes do not affect the current functioning.

Thanks.

Issue #1980, opened 2022-09-18
MOJO_SERVER_DEBUG=1 fails with UNIX domain sockets


s1037989:
```
$ . /etc/os-release && echo $VERSION
20.04.2 LTS (Focal Fossa)
$ mojo version
CORE
  Perl        (v5.34.0, linux)
  Mojolicious (9.27, Waffle)

OPTIONAL
  Cpanel::JSON::XS 4.09+   (4.26)
  EV 4.32+                 (4.33)
  IO::Socket::Socks 0.64+  (0.74)
  IO::Socket::SSL 2.009+   (2.075)
  Net::DNS::Native 0.15+   (n/a)
  Role::Tiny 2.000001+     (2.002004)
  Future::AsyncAwait 0.52+ (0.58)

This version is up to date, have fun!
```

### Steps to reproduce the behavior
Start `mojo daemon` with and without `MOJO_SERVER_DEBUG` on a UNIX domain socket and `mojo get` it.

### Expected behavior
Web server handles request and logs server debugs, as with `MOJO_SERVER_DEBUG` disabled.

#### Server
```
$ MOJO_SERVER_DEBUG= mojo daemon -l 'http+unix://my-app'
Web application available at my-app
```

#### Client
```
$ mojo get 'http+unix://my-app'
Your Mojo is working!
```

### Actual behavior
Web server doesn't handle request and doesn't log server debugs, as with `MOJO_SERVER_DEBUG` enabled.  Gives this error:

#### Server
```
$ MOJO_SERVER_DEBUG=1 mojo daemon -l 'http+unix://my-app'
Web application available at my-app
Mojo::Reactor::EV: I/O watcher failed: Can't locate object method "peerhost" via package "IO::Socket::UNIX" at $HOME/perl5/perlbrew/perls/perl-5.34.0/lib/site_perl/5.34.0/Mojo/Server/Daemon.pm line 197.
```

#### Client
```
$ mojo get 'http+unix://my-app'
Premature connection close at $HOME/perl5/perlbrew/perls/perl-5.34.0/lib/site_perl/5.34.0/Mojolicious/Command/get.pm line 77.
```

s1037989:
Is this an acceptable patch to submit in a PR?  If this needs a test, can you suggest how I could go about testing this?

```
$ git diff lib/Mojo/Server/Daemon.pm
diff --git a/lib/Mojo/Server/Daemon.pm b/lib/Mojo/Server/Daemon.pm
[...]

PR #2011, opened 2022-11-26
fix UNIX domain socket debugging reported by #1980

s1037989:
### Summary
This patch logs the handle's `peerpath` if the handle is a `IO::Socket::UNIX` object, otherwise logs the handle's `peeraddr`.

### Motivation
The server fails when debugging a server listening on a unix domain socket because a `IO::Socket::UNIX` object does not have a `peerhost` method.  This patch retains existing functionality except when the handle is an `IO::Socket::UNIX` object (i.e. listening on a unix domain socket).

### References
MOJO_SERVER_DEBUG=1 fails with UNIX domain sockets #1980.

### DIFF
--- a/lib/Mojo/Server/Daemon.pm
+++ b/lib/Mojo/Server/Daemon.pm
@@ -194,7 +194,7 @@ sub _listen {
       my ($loop, $stream, $id) = @_;
 
       $self->{connections}{$id} = {tls => $tls};
-      warn "-- Accept $id (@{[$stream->handle->peerhost]})\n" if DEBUG;
+      warn "-- Accept $id (@{[_peer($stream->handle)]})\n" if DEBUG;
       $stream->timeout($self->inactivity_timeout);
 
       $stream->on(close   => sub { $self && $self->_close($id) });
@@ -211,6 +211,8 @@ sub _listen {
   $url->port($self->ports->[-1]) if !$options->{path} && !$url->port;
   say 'Web application available at ', $options->{path} // $url;
 }
+
+sub _peer { $_[0]->isa('IO::Socket::UNIX') ? $_[0]->peerpath : $_[0]->peerhost }
 
 sub _read {
   my ($self, $id, $chunk) = @_;


lib/Mojo/Server/Daemon.pm
```diff
@@ -194,7 +194,8 @@ sub _listen {
       my ($loop, $stream, $id) = @_;
 
       $self->{connections}{$id} = {tls => $tls};
-      warn "-- Accept $id (@{[$stream->handle->peerhost]})\n" if DEBUG;
+      my $peer = $stream->handle->isa('IO::Socket::UNIX') ? 'peerpath' : 'peeraddr' if DEBUG;
+      warn "-- Accept $id (@{[$stream->handle->$peer]})\n" if DEBUG;
```
kraih, line 198: This code is rather ugly. Duplicated calls of`$stream->handle` and uses of the `DEBUG` constant.

lib/Mojo/Server/Daemon.pm
```diff
@@ -194,7 +194,8 @@ sub _listen {
       my ($loop, $stream, $id) = @_;
 
       $self->{connections}{$id} = {tls => $tls};
-      warn "-- Accept $id (@{[$stream->handle->peerhost]})\n" if DEBUG;
+      my $peer = $stream->handle->isa('IO::Socket::UNIX') ? 'peerpath' : 'peeraddr' if DEBUG;
```
Grinnz, line 197: Shouldn't this still be using peerhost rather than peeraddr? peeraddr according to the documentation is a binary octet string which would not be suitable for printing

s1037989: Stupid of me!  I didn't meant to change it from `peerhost` to `peeraddr`.  I'll change that when I improve the rest of the patch.  Thanks for catching that.

lib/Mojo/Server/Daemon.pm
s1037989: Force pushed an update to address your comments.

lib/Mojo/Server/Daemon.pm
```diff
@@ -194,7 +194,11 @@ sub _listen {
       my ($loop, $stream, $id) = @_;
 
       $self->{connections}{$id} = {tls => $tls};
-      warn "-- Accept $id (@{[$stream->handle->peerhost]})\n" if DEBUG;
+      if (DEBUG) {
+        my $handle = $stream->handle;
+        my $peer = $handle->isa('IO::Socket::UNIX') ? 'peerpath' : 'peerhost';
```
Grinnz, line 199: Needs perltidy

lib/Mojo/Server/Daemon.pm
s1037989: @kraih After sleeping on it, a function is much nicer.  I have force-pushed a better solution to this PR.

lib/Mojo/Server/Daemon.pm
```diff
@@ -5,9 +5,9 @@ use Carp qw(croak);
 use Mojo::IOLoop;
 use Mojo::Transaction::WebSocket;
 use Mojo::URL;
-use Mojo::Util      qw(term_escape);
+use Mojo::Util qw(term_escape);
```
Grinnz, line 8: Make sure your perltidy is up to date, you seem to have untidied these lines.

lib/Mojo/Server/Daemon.pm
```diff
@@ -212,6 +212,12 @@ sub _listen {
   say 'Web application available at ', $options->{path} // $url;
 }
 
+sub _peer {
```
Grinnz, line 215: Not in alphabetical order

lib/Mojo/Server/Daemon.pm
s1037989: I wondered about that!  Thanks, I updated my Perl::Tidy from 20210717 to 20221112, re-tidied, and force-pushed.

lib/Mojo/Server/Daemon.pm
```diff
@@ -212,6 +212,12 @@ sub _listen {
...
+  my $handle = shift;
+  my $peer   = $handle->isa('IO::Socket::UNIX') ? 'peerpath' : 'peerhost';
+  return $handle->$peer;
```
kraih, line 218: Why do you insist on making it so verbose?
```perl
sub _peer { $_[0]->->isa('IO::Socket::UNIX') ? $_[0]->peerpath : $_[0]->peerhost }
```

s1037989: :+1:  Much better.  Updated.

jhthorsen (approved):
LGTM 👍

marcusramberg (approved):
lgtm

Issue #2008, opened 2022-11-19
proxy->start_p helper doesn't forward empty body responses


s1037989:
```
$ mojo version
CORE
  Perl        (v5.34.1, linux)
  Mojolicious (9.29, Waffle)

OPTIONAL
  Cpanel::JSON::XS 4.09+   (4.28)
  EV 4.32+                 (4.33)
  IO::Socket::Socks 0.64+  (0.74)
  IO::Socket::SSL 2.009+   (2.074)
  Net::DNS::Native 0.15+   (0.22)
  Role::Tiny 2.000001+     (2.002004)
  Future::AsyncAwait 0.52+ (0.58)

This version is up to date, have fun!
$ (. /etc/os-release && echo $PRETTY_NAME)
Pop!_OS 22.04 LTS
```
### Steps to reproduce the behavior
`t/mojolicious/proxy_app.t` patch:
```
$ git diff t/mojolicious/proxy_app.t 
diff --git a/t/mojolicious/proxy_app.t b/t/mojolicious/proxy_app.t
index 1eadb6dbd..5f9474ade 100644
--- a/t/mojolicious/proxy_app.t
+++ b/t/mojolicious/proxy_app.t
@@ -16,6 +16,13 @@ my $daemon = Mojo::Server::Daemon->new(listen => ['http://127.0.0.1'], silent =>
 my $port   = $daemon->start->ports->[0];
 my $url    = Mojo::URL->new("http://127.0.0.1:$port");
 my $r      = $app->routes;
+$r->get(
+  '/empty' => sub {
+    my $c = shift;
+    $c->res->headers->header('X-Mojo-App' => 'Empty');
+    $c->render(data => '', status => 200);
+  }
+);
 $r->get(
   '/res1' => sub {
     my $c = shift;
@@ -102,6 +109,8 @@ get '/proxy3/:method/*target' => sub {
 my $t = Test::Mojo->new;
 
 subtest 'Various response variants' => sub {
+  $t->get_ok('/proxy1/empty')->status_is(200)->header_is('X-Mojo-App' => 'Empty')->content_is('');
   $t->get_ok('/proxy1/res1')->status_is(200)->header_is('X-Mojo-App' => 'One')->content_is('One!');
   $t->get_ok('/proxy1/res2')->status_is(200)->header_is('X-Mojo-App' => 'Two')->content_is('Two!');
[...]

PR #2012, opened 2022-11-26
forward empty body responses as reported by #2008

s1037989:
### Summary
This patch allows proxying empty body responses such as those with code 200 and length 0, and, more importantly, 301 and 302 redirects with length 0.

While this patch still passes all tests -- as well as the 12 tests added -- I don't know if this solution is acceptable and in the spirit of the comment that was there.  It seems the conditional that is currently used is insufficient, and I don't know what would be better; moreover, no tests fail with this change.

### Motivation
The gateway (proxy) server using `proxy->start_p` is unable to successfully handle redirects received from the origin (backend) server.

### References
proxy->start_p helper doesn't forward empty body responses #2008.

### DIFF
--- a/lib/Mojolicious/Plugin/DefaultHelpers.pm
+++ b/lib/Mojolicious/Plugin/DefaultHelpers.pm
@@ -242,9 +242,7 @@ sub _proxy_start_p {
         }
       );
 
-      # Unknown length (fall back to connection close)
-      $source_res->once(finish => sub { $content->$write('') and $tx->resume })
-        unless length($headers->content_length // '');
+      $source_res->once(finish => sub { $content->$write('') and $tx->resume });
     }
   );
   weaken $source_tx;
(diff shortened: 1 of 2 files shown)


jhthorsen (dismissed):
Looks like the perltidy check fails. I can't say I find anything wrong with the PR other than that, but I'm afraid I have to vote neutral, since I'm not sure what is really correct or not regarding specifications.

kraih:
> Looks like the perltidy check fails.

Yes, the PR needs to be rebased.

s1037989:
Re-based and tidied.

kraih (commented):
Looks like the worst case here would be a performance regression. Given that it fixes an actual bug, it's probably worth it.

jhthorsen (approved):

marcusramberg (approved):
Looks like a good fix to me.

PR #2016, opened 2022-12-09
echo.websocket.org service no longer available

s1037989:
### Summary
`wss://ws.postman-echo.com/raw` is a drop-in replacement for this example that works.

### Motivation
`ws://echo.websocket.org` service no longer available  ([http://echo.websocket.org](http://echo.websocket.org))

### References
None.


jhthorsen (approved):
LGTM 👍

marcusramberg (approved):
👍🏻

Issue #1733, opened 2021-03-10
How to pass tls options (force TLS 1.2) to Mojo::UserAgent?


knutov:
What is the correct way to pass tls options to Mojo::UserAgent now?

I want to force use only TLS 1.2.

`IO::Socket::SSL::set_client_defaults(SSL_version => 'TLSv12:!SSLv2:!SSLv3:!TLSv1:!TLSv11');` works for me, but this setting works only as global option.

Perl        (v5.18.2, linux)
Mojolicious (9.03, Waffle)

Grinnz:
Please use the [support channels](https://docs.mojolicious.org/#SUPPORT) for discussion and questions.

Mojo::UserAgent only has passthrough for the `ca`, `cert`, `key`, and `insecure` TLS options currently - setting global defaults is an alternative.

jberger:
Can't you use https://docs.mojolicious.org/Mojo/UserAgent#socket_options ?

Grinnz:
That accepts IO::Socket::IP options not TLS options.

kraih:
As a feature request this would have been reasonable.

Issue #1230, opened 2018-06-13
Add SSL_check_crl and SSL_crl_file to Mojo::IOLoop::TLS (feature request)


fskale:
Hi,
my PKI generates a CRL on daily bases (revoking clients), so
it would be great to have the feature included in the next version of Mojolicious.
I think, that can be done with minimal effort by extending the _expand function to add both **IO::Socket::SSL** options if e.g. **tls_crl=/etc/crl.pem** have been supplied at the commandline.
Like:
<pre>
if ($args->{tls_crl}) {
      $tls->{SSL_crl_file} = $args->{tls_crl};
      $tls->{SSL_check_crl} = 'true';
}
</pre>
Right now i do:
<pre>
IO::Socket::SSL::set_server_defaults(
     SSL_check_crl => 'true', 
    SSL_crl_file => '/etc/crl.pem'
);
</pre>

Btw., you're doing a great job !

Best regards
Franz

kraih:
Looks like this feature request went stale.

shadowcat-mst:
@fskale this seems like a reasonable idea - could you try expanding it into a patch w/tests so we can discuss it in more concrete terms?

fskale:
Hi,
will try to draft a patch with tests. Right now i'm on vacation, so i'll report back next week.

Best regards
Franz

kraih:
Unfortunately i don't think a patch for this very specific feature has much of a chance of getting accepted, considering how little interest there is currently. Only way i could see support for this getting accepted was if someone found a nice generic way to pass along arbitrary `IO::Socket::SSL` options from `Mojo::Server::Daemon` and `Mojo::UserAgent`.

fskale:
@kraih 
Hope, that you get my answer althought closed already.
Mojo::IOLoop::TLS already pass args directly to IO::Socket::SSL.
Mojolicious also supports both the <code> $tls->{SSL_ca_file} = $args->{tls_ca} </code>
as well as the <code>tls_verify </code> whereas the crl is a feature to it.
Anyone who uses client certificates is also aware of the ability to use a CRL.
Nothing special though.
Client Certificate authentication along with password auth are quite common in high security environments.
Anyway, i can live with my solution.

kraih:
Yes, we do receive comments for closed issues, and regularly reopen them when there is new activity. Unfortunately you appear to be the only one who cares about this feature at the moment. So if you don't make a patch, i don't think there will be an alternative any time soon.

[...]

PR #2017, opened 2022-12-12
Allow setting TLS options for connections in Mojo::UserAgent

heikojansen:
My use case is that I need to communicate with a server which uses a too small DH key (OpenSSL error "tls_process_ske_dhe:dh key too small") but I do not want to manipulate the cipher list globally (e.g. via `IO::Socket::SSL::set_defaults( SSL_cipher_list => 'DEFAULT:!DH@SECLEVEL=1' )`). This probably lacks a test but I'm not really sure how a test would look which actually confirmed that TLS behaviour was successfully manipulated via an option.

### DIFF
--- a/t/mojo/user_agent_tls.t
+++ b/t/mojo/user_agent_tls.t
@@ -98,4 +98,28 @@ subtest 'Web server with valid certificates and no verification' => sub {
   is $ua->ioloop->stream($tx->connection)->handle->get_sslversion, 'TLSv1',      'TLSv1 has been negotiatied';
 };
 
+subtest 'Client side TLS options' => sub {
+  my $daemon = Mojo::Server::Daemon->new(app => app, ioloop => Mojo::IOLoop->singleton, silent => 1);
+  my $listen = 'https://127.0.0.1/?version=TLSv1_1';
+  my $port   = $daemon->listen([$listen])->start->ports->[0];
+
+  subtest '(Not) setting verification mode' => sub {
+    my $ua = Mojo::UserAgent->new(ioloop => Mojo::IOLoop->singleton);
+    my $tx = $ua->get("https://127.0.0.1:$port");
+    like $tx->error->{message}, qr/certificate verify failed/, 'has error';
+
+    $ua = Mojo::UserAgent->new(ioloop => Mojo::IOLoop->singleton);
+    $ua->tls_options({SSL_verify_mode => 0x00});
+    $tx = $ua->get("https://127.0.0.1:$port");
+    ok !$tx->error, 'no error';
+  };
+
+  subtest 'Setting acceptable protocol version' => sub {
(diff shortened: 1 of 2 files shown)


kraih:
I'm afraid without tests this PR cannot be considered.

kraih:
Always squash your commits.

t/mojo/user_agent_tls.t
```diff
@@ -98,4 +98,29 @@ subtest 'Web server with valid certificates and no verification' => sub {
   is $ua->ioloop->stream($tx->connection)->handle->get_sslversion, 'TLSv1',      'TLSv1 has been negotiatied';
 };
 
+subtest 'Client side TLS options' => sub {
+  my $daemon = Mojo::Server::Daemon->new(app => app, ioloop => Mojo::IOLoop->singleton, silent => 1);
+  my $listen = 'https://127.0.0.1/?version=TLSv1_1';
+  my $port   = $daemon->listen([$listen])->start->ports->[0];
+
+  subtest '(Not) setting verification mode' => sub {
+    my ( $ua, $tx );
```
kraih, line 107: This shouldn't be a separate line.

t/mojo/user_agent_tls.t
```diff
@@ -98,4 +98,29 @@ subtest 'Web server with valid certificates and no verification' => sub {
...
+    $ua = Mojo::UserAgent->new(ioloop => Mojo::IOLoop->singleton);
+    $tx = $ua->get("https://127.0.0.1:$port");
+    ok index($tx->error->{message}, 'certificate verify failed'), 'has error';
```
kraih, line 110: This would probably be better as a `like` for error reporting.

t/mojo/user_agent_tls.t
```diff
@@ -98,4 +98,29 @@ subtest 'Web server with valid certificates and no verification' => sub {
...
+    $ua->tls_options({SSL_verify_mode => 0x00});
+    $tx = $ua->get("https://127.0.0.1:$port");
+    ok !$tx->error, 'no error';
+  };
+
+  subtest 'Setting acceptable protocol version' => sub {
+    my $ua = Mojo::UserAgent->new(ioloop => Mojo::IOLoop->singleton);
+    $ua->tls_options({SSL_version => 'TLSv1_2'});
+    my $tx = $ua->get("https://127.0.0.1:$port");
+    ok index($tx->error->{message}, 'wrong ssl version'), 'has error';
```
kraih, line 122: Same here.

heikojansen:
I believe that I've addressed the remaining issues so the code should be ready for another review

heikojansen:
Apparently the perltidy formatting was still missing but yesterdays commit should have fixed that.

kraih (approved):

Grinnz (approved):
Previously requested in https://github.com/mojolicious/mojo/issues/1230, https://github.com/mojolicious/mojo/issues/1733, etc. simple enough and useful feature

PR #2021, opened 2023-01-16
tutorial typo: "lite_app" instead of "lite-app" with "mojo generate"

ngaro:
### Summary
Tutorial mentions `mojo generate lite-app` instead of `mojo generate lite_app` which is the correct syntax _(at least in my version of mojolicious, maybe this has been changed over time)_

### Motivation
Docs are rather useless if they are wrong...


Grinnz:
hyphen syntax is supported and preferred since Mojolicious 8.41. you should use the documentation of the version of Mojolicious you are using.

Issue #2023, opened 2023-01-23
Mojo::DOM doesn't handle :nth-child correctly


mauke:
* Mojolicious version: 9.31
* Perl version: v5.36.0
* Operating system: Linux (Ubuntu 22.04)

### Steps to reproduce the behavior
```perl
#!/usr/bin/env perl
use v5.12.0;
use warnings;
use Test::More;
use Mojo::DOM;

my $dom = Mojo::DOM->new(<<'_EOT_');
<!DOCTYPE html>
<ul> <li>Ax1</li> </ul>
<ul> <li>Bx1</li> <li>Bx2</li> </ul>
<ul> <li>Cx1</li> <li>Cx2</li> <li>Cx3</li> </ul>
<ul> <li>Dx1</li> <li>Dx2</li> <li>Dx3</li> <li>Dx4</li> </ul>
<ul> <li>Ex1</li> <li>Ex2</li> <li>Ex3</li> <li>Ex4</li> <li>Ex5</li> </ul>
<ul> <li>Fx1</li> <li>Fx2</li> <li>Fx3</li> <li>Fx4</li> <li>Fx5</li> <li>Fx6</li> </ul>
_EOT_

is_deeply
    $dom->find('li:nth-child(-n+3)')->map(sub { $_->to_string })->to_array,
    [qw(
        <li>Ax1</li>
        <li>Bx1</li> <li>Bx2</li>
        <li>Cx1</li> <li>Cx2</li> <li>Cx3</li>
        <li>Dx1</li> <li>Dx2</li> <li>Dx3</li>
        <li>Ex1</li> <li>Ex2</li> <li>Ex3</li>
        <li>Fx1</li> <li>Fx2</li> <li>Fx3</li>
    )],
    'all first three children';

done_testing;
```

### Expected behavior
Test passes.

<https://www.w3.org/TR/selectors-3/#nth-child-pseudo>:
> The ***:nth-child(*a*`n`+*b*)*** pseudo-class notation represents an element that has *a*`n`+*b*-1 siblings [before](https://www.w3.org/TR/selectors-3/#sel-before) it in the document tree, for any positive integer or zero value of `n`.

[...]

> The value *a* can be negative, but only the positive values of *a*`n`+*b*, for `n`≥0, may represent an element in the document tree.

> > Example:

> >     html|tr:nth-child(-n+6)  /* represents the 6 first rows of XHTML tables */

### Actual behavior
```
not ok 1 - all first three children
[...]

PR #2032, opened 2023-02-01
Fix :nth-child selector

mauke:
### Summary
Handle `:nth-child(An+B)` algebraically instead of iterating over possible values of n.

### Motivation
Previous implementation was broken (but so were the tests, so nobody noticed).

### References
See issue #2023.


t/mojo/dom.t
```diff
@@ -1040,28 +1040,28 @@ EOF
   is_deeply \@li, ['F'], 'found third last li element';
   @li = ();
   $dom->find('li:nth-child(1n+0)')->each(sub { push @li, shift->text });
-  is_deeply \@li, [qw(A B C D E F G)], 'found all li elements';
+  is_deeply \@li, [qw(A B C D E F G H)], 'found all li elements';
```
kraih, line 1043: This looks incorrect too. In Chrome i get `A C E F H I` for `li:nth-child(1n+0)`.

mauke: Can't reproduce, at least with Firefox. I put
```html
<style>
    li:nth-child(1n+0) {
        outline: 1px solid green;
    }
</style>
<ul>
    <li>A</li>
    <li>B</li>
    <li>C</li>
    <li>D</li>
    <li>E</li>
    <li>F</li>
    <li>G</li>
    <li>H</li>
</ul>
```
in a local file and opened it in the browser. All list items got a green border. Also, running `document.querySelectorAll('li:nth-child(1n+0)').length` in the developer console gave `8`.

mauke: Oh, I see what you did. You accidentally took the HTML code from line 1077, which is part of a different test. The HTML code you used has `<p>B</p>`, `<p>D</p>`, and `<p>G</p>`, which is why they don't show up with a `li` selector, and it has an additional `<li>I</li>` element. The correct HTML code for this test is in line 920 and doesn't contain any `I`.

kraih: Oops, that's exactly what i did.

kraih (approved):

kraih:
[@mojojs/dom](https://github.com/mojolicious/dom.js) will have the same bug if you want to apply the fix there too.

kraih:
Since you didn't say anything, i've ported the fix over to `@mojojs/dom`. https://github.com/mojolicious/dom.js/commit/a7ac95d131f2b4c612bdbe5db902a788154c1b80

PR #2052, opened 2023-03-16
ensure eval is always called in scalar context (#2051)

Grinnz:
### Summary
eval in list context returns empty list when it fails, not undef. List context includes "use constant" because list constants unfortunately exist.

### Motivation
We want to ensure no danger of bugs due to context sensitivity.

### References
#2051


jhthorsen (approved):
LGTM 👍

kraih (approved):

PR #2053, opened 2023-03-20
Call POSIX::setsid() as a function

atoomic:
This is avoiding a warning when load with strict sub enabled:

    Bareword "POSIX::setsid" not allowed while "strict subs" in use

### Summary

### Motivation
EXPLAIN WHY YOU BELIEVE THESE CHANGES ARE NECESSARY HERE

### References
LIST RELEVANT ISSUES, PULL REQUESTS AND FORUM DISCUSSIONS HERE

### DIFF
--- a/lib/Mojo/Server.pm
+++ b/lib/Mojo/Server.pm
@@ -34,7 +34,7 @@ sub daemonize {
   # Fork and kill parent
   die "Can't fork: $!" unless defined(my $pid = fork);
   exit 0 if $pid;
-  POSIX::setsid == -1 and die "Can't start a new session: $!";
+  POSIX::setsid() == -1 and die "Can't start a new session: $!";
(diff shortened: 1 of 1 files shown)


atoomic:
The problem is not POSIX::setid being defined or not. This is a warning when load with `strict subs` that I can fix on my side when applying this patch. This probably indicates another issue but was considering using `parens` to call the function a good thing. 
If you disagree let's close this PR.

Grinnz:
This is certainly an indication of another issue. There should be no case in which the POSIX::setsid symbol is undefined in the parser.

toddr:
> This is certainly an indication of another issue. There should be no case in which the POSIX::setsid symbol is undefined in the parser.

Can't we just merge it as "more legible code?"

PR #2061, opened 2023-04-23
Add exceptions to stash on all exception formats

### LINKED ISSUE #2048: Exceptions are not available in the after_dispatch hook when exception_format is set to json
* Mojolicious version: 9.31
* Perl version: 5.36.0

According to the Routing guide, the recommended way to forward exceptions to a web service is in the ````after_dispatch```` hook:

````perl
# Forward exceptions to a web service
$app->hook(after_dispatch => sub ($c) {
  return unless my $e = $c->stash('exception');
  $c->ua->post('https://example.com/bugs' => form => {exception => $e});
});
````

https://docs.mojolicious.org/Mojolicious/Guides/Routing#Hooks

However, when the `exception_format` is set to `json`, the exception is not available in the stash, so it is not possible to forward exceptions to a web service using the same technique. I'm not sure if this would be considered a bug or just a limitation of using json format exceptions.

### PR DESCRIPTION
**Pull request by rawleyfowler**
### Summary
Make it easier to use exceptions, when exception format is not HTML

### Motivation
It doesn't make sense for exceptions to be stored in the stash only if the format is HTML, instead it makes more sense, and is more consistent to store them in the stash for all formats

### References
fixes #2048

### DIFF
--- a/lib/Mojo/IOLoop/Client.pm
+++ b/lib/Mojo/IOLoop/Client.pm
@@ -10,13 +10,13 @@ use Scalar::Util qw(weaken);
 use Socket       qw(IPPROTO_TCP SOCK_STREAM TCP_NODELAY);
 
 # Non-blocking name resolution requires Net::DNS::Native
-use constant NNR => $ENV{MOJO_NO_NNR} ? 0 : eval { require Net::DNS::Native; Net::DNS::Native->VERSION('0.15'); 1 };
+use constant NNR => $ENV{MOJO_NO_NNR} ? 0 : !!eval { require Net::DNS::Native; Net::DNS::Native->VERSION('0.15'); 1 };
 my $NDN;
 
 # SOCKS support requires IO::Socket::Socks
 use constant SOCKS => $ENV{MOJO_NO_SOCKS}
   ? 0
-  : eval { require IO::Socket::Socks; IO::Socket::Socks->VERSION('0.64'); 1 };
+  : !!eval { require IO::Socket::Socks; IO::Socket::Socks->VERSION('0.64'); 1 };
 use constant READ  => SOCKS ? IO::Socket::Socks::SOCKS_WANT_READ()  : 0;
 use constant WRITE => SOCKS ? IO::Socket::Socks::SOCKS_WANT_WRITE() : 0;
 
--- a/lib/Mojo/IOLoop/TLS.pm
+++ b/lib/Mojo/IOLoop/TLS.pm
@@ -6,7 +6,7 @@ use Mojo::IOLoop;
 use Scalar::Util qw(weaken);
 
 # TLS support requires IO::Socket::SSL
-use constant TLS   => $ENV{MOJO_NO_TLS} ? 0 : eval { require IO::Socket::SSL; IO::Socket::SSL->VERSION('2.009'); 1 };
+use constant TLS   => $ENV{MOJO_NO_TLS} ? 0 : !!eval { require IO::Socket::SSL; IO::Socket::SSL->VERSION('2.009'); 1 };
 use constant READ  => TLS               ? IO::Socket::SSL::SSL_WANT_READ()  : 0;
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

**Review by kraih — CHANGES_REQUESTED** [review]
The tests do nothing.

**Comment by rawleyfowler** [comment]
@kraih Can you elaborate a little bit? The way I see it is the exception will become undef if there is no exception in the stash via the after dispatch hook. Should I be using `is` or `ok` instead?

**Comment by kraih** [comment]
> @kraih Can you elaborate a little bit? The way I see it is the exception will become undef if there is no exception in the stash via the after dispatch hook. Should I be using `is` or `ok` instead?

You can literally run those tests without the proposed change and everything passes.

**Comment by rawleyfowler** [comment]
@kraih Okay, I think I got it, sorry, I see now.

File: lib/Mojolicious/Plugin/DefaultHelpers.pm [hunk 1]
```diff
@@ -196,6 +196,7 @@ sub _is_fresh {
 
 sub _json_exception {
   my ($c, $e) = @_;
+  $c->stash->{exception} = _is_e($e) ? $e : Mojo::Exception->new($e);
```
kraih, line 199: `_is_e($e) ? $e : Mojo::Exception->new($e)` should probably be abstracted in a way so it gets free test coverage from somewhere else.

rawleyfowler: Just pushed up, would you like me to write tests for it as well?

lib/Mojolicious/Plugin/DefaultHelpers.pm
```diff
@@ -189,13 +189,19 @@ sub _inactivity_timeout {
 
 sub _is_e { blessed $_[0] && $_[0]->isa('Mojo::Exception') }
 
+sub _convert_to_exception {
+  my $e = shift;
+  return _is_e($e) ? $e : Mojo::Exception->new($e);
+}
```
kraih, line 195: Not ordered alphabetically.

lib/Mojolicious/Plugin/DefaultHelpers.pm
kraih: By abstracting it out it should now be covered by existing tests automatically.

kraih:
Don't forget to squash commits.

lib/Mojolicious/Plugin/DefaultHelpers.pm
```diff
@@ -90,6 +90,11 @@ sub _content {
   return Mojo::ByteStream->new($hash->{$name} // '');
 }
 
+sub _convert_to_exception {
+  my $e = shift;
+  return _is_e($e) ? $e : Mojo::Exception->new($e);
```
jhthorsen, line 95: Not sure if this is very important, but could `_is_e()` be inlined in this function? Looks like it's only being used one place now.

jhthorsen (approved):
I'm positive to this change. I've often overridden the exception helper to achieve the same, which I think I can stop doing after this 👍 

Wouldn't mind if the _is_e() method was inlined though.

lib/Mojolicious/Plugin/DefaultHelpers.pm
rawleyfowler: I'll make that change right away, thanks for the feedback!

jhthorsen (approved):
LGTM 👍

marcusramberg (approved):
LGTM :)

lib/Mojolicious/Plugin/DefaultHelpers.pm
jhthorsen: Did _is_e() come back by accident?

rawleyfowler: Must have screwed up my commit whoops

rawleyfowler:
Umm, I'm not sure what happened but this was closed?

rawleyfowler:
Okay I'm an idiot, re-opening this in a different PR, I guess I accidentally changed a commit signature in my rebase :L

PR #2065, opened 2023-05-05
Add exceptions to stash regardless of format

### LINKED ISSUE #2048: Exceptions are not available in the after_dispatch hook when exception_format is set to json

rawleyfowler:
**Summary**

Make it easier to use exceptions, when exception format is not HTML

**Motivation**

It doesn't make sense for exceptions to be stored in the stash only if the format is HTML, instead it makes more sense, and is more consistent to store them in the stash for all formats

**References**

fixes #2048 

I annihilated my branch on the other PR like a goof #2061 and github auto closed the PR because of it.

### DIFF
--- a/lib/Mojolicious/Plugin/DefaultHelpers.pm
+++ b/lib/Mojolicious/Plugin/DefaultHelpers.pm
@@ -90,6 +90,11 @@ sub _content {
   return Mojo::ByteStream->new($hash->{$name} // '');
 }
 
+sub _convert_to_exception {
+  my $e = shift;
+  return (blessed $e && $e->isa('Mojo::Exception')) ? $e : Mojo::Exception->new($e);
+}
+
 sub _csrf_token { $_[0]->session->{csrf_token} ||= hmac_sha1_sum($$ . steady_time . rand, $_[0]->app->secrets->[0]) }
 
 sub _current_route {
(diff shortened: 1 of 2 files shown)


marcusramberg (dismissed):
Still lgtm :)

marcusramberg:
> @marcusramberg Had to make a change `$e->inspect` could break

Its failing test now?

rawleyfowler:
@marcusramberg The tests pass, there is a bug with github workflow where it can't find the module if you look at the logs. Not sure.

rawleyfowler:
@kraih Is this failing test my doing or is the `@INC` broken?

jhthorsen (dismissed):

marcusramberg (approved):

jhthorsen (approved):

PR #2067, opened 2023-05-11
Add new method to `$dom`, `$dom->each_text`.

tcheukueppo:
Hello,

### Summary
1. New `$dom` method `$dom->each_text`: which returns text content of every descendant node in a list.
2. Tests on `$dom->each_text`.

### Motivation
I'd like to have this feature to be table to perform context sensitive parsing on the text content of a given tag(with all it descendant in the most frequent cases) without interference with contents of other `tags`(which is something that `$dom->all_text` does not provide), I'm totally aware of `$dom->text` but that do require me to walk through the $dom data structure and use css selectors which is quite tedious.

### DIFF
--- a/lib/Mojo/DOM.pm
+++ b/lib/Mojo/DOM.pm
@@ -168,6 +168,8 @@ sub tap { shift->Mojo::Base::tap(@_) }
 
 sub text { _text(_nodes(shift->tree), 0, 0) }
 
+sub each_text { _each_text(shift->tree, 0) }
+
 sub to_string { ${shift()}->render }
 
 sub tree { @_ > 1 ? (${$_[0]}->tree($_[1]) and return $_[0]) : ${$_[0]}->tree }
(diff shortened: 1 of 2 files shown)


tcheukueppo:
Sorry, I think it is the perltidy, let me undo that.

kraih:
I specifically meant that adding `each_text` and changing the parser regex are unrelated changes.

tcheukueppo:
After performing some tests I think modifying parser regex isn't worth it and thus no PR for that.

tcheukueppo:
done

lib/Mojo/DOM.pm
```diff
@@ -468,6 +486,16 @@ excluded.
   # "foo\nbarbaz\n"
   $dom->parse("<div>foo\n<p>bar</p>baz\n</div>")->at('div')->all_text;
 
+=head2 each_text
+
+  my @text = $dom->each_text;
```
jhthorsen, line 491: Mojo::Headers's to_hash() can take "1" as an argument, so maybe we could add that to `all_text()` instead of adding a new method? Also, how about returning a Mojo::Collection?

kraih:
I don't really see the use of this feature i'm afraid. Maybe some more specific examples for use cases would help.

tcheukueppo:
Really sorry for bothering you guys, I was working on a project and I came across this, I thought that do be useful but turns out that my approach was not the right approach, it is obviously not useful to me now and probably be never useful to anyone out there, I found another technique that did not even required me to monkey_patch anything, thanks for your attention,

tcheukueppo:
Closing this PR.

Issue #2046, opened 2023-03-09
Manager can send SIGQUIT to worker while exiting, causing coredump


brsakai-csco:
* Mojolicious version: 9.22, Waffle
* Perl version: v5.34.0
* Operating system: Ubuntu 22.04.2 LTS

### Steps to reproduce the behavior
1. Create and start simple Hypnotoad app
2. Send a `SIGQUIT` to a worker process to end it gracefully
3. The worker's manager will send a second `SIGQUIT`
4. Depending on timing, the worker will reset its signal handlers to `SIG_DFL` before catching the second `SIGQUIT`, causing it to dump its core

In more detail:

Unpack the attached tarfile which has my app and update `mojo_server_wrapper.pl` to have the correct `$ENV{REPRO_PATH}` (you could most likely use any Hypnotoad/Prefork app, but I figured a working example would be best)

[hypno.tar.gz](https://github.com/mojolicious/mojo/files/10936274/hypno.tar.gz)

Start the app from terminal (A)
```
brsakai@brsakai-1151:~/hypno$ ./mojo_server_wrapper.pl
[2023-03-09 19:03:32.09625] [29198] [info] Listening at "http://127.0.0.1:8080"
Web application available at http://127.0.0.1:8080
[2023-03-09 19:03:32.09659] [29198] [info] Manager 29198 started
[2023-03-09 19:03:32.09910] [29199] [info] Worker 29199 started
[2023-03-09 19:03:32.10025] [29200] [info] Worker 29200 started
[2023-03-09 19:03:32.10118] [29201] [info] Worker 29201 started
[2023-03-09 19:03:32.10202] [29202] [info] Worker 29202 started
[2023-03-09 19:03:32.10239] [29198] [info] Creating process id file "/tmp/mojo_server.pid"
[2023-03-09 19:03:32.10248] [29203] [info] Worker 29203 started
```

`strace` one of the workers from another terminal (B)
```
brsakai@brsakai-1151:~$ sudo strace -p 29199
strace: Process 29199 attached
epoll_wait(8,
```

Issue a `kill -QUIT` from a third terminal (C)
```
brsakai@brsakai-1151:~$ kill -QUIT 29199
```

Observe that the kill signal is received in the strace output (note that two SIGQUIT are received, one from terminal (C) and one from the manager)
```
brsakai@brsakai-1151:~$ sudo strace -p 29199
strace: Process 29199 attached
[...]

PR #2073, opened 2023-06-09
Do not send redundant SIGQUITs

brsakai-csco:
If the worker reports itself as shutting down, do not second another SIGQUIT to the worker. It already knows it is shutting down, and this can cause a race with signal handlers that triggers coredumps.

### Summary
Full writeup in #2046
The short version:
- When a worker receives a SIGQUIT, it begins to gracefully shut down
- During graceful shutdown, the worker notifies the manager thread of its shutdown
- If the manager sent the initial SIGQUIT, then it already knows the worker is shutting down and takes no action
- If the manager did not send the initial SIGQUIT, then it marks the worker for graceful shutdown, causing it to send a second SIGQUIT
- This second SIGQUIT can arrive once Perl has already de-registered the QUIT signal handler, which causes it to perform the default QUIT action (coredump)

The patch here increments the worker's `quit` state in the managing thread so that it does not send a second SIGQUIT to the worker if the manager finds out about the graceful shutdown from the worker

### Motivation
This behavior was causing coredumps in our application, since we were using SIGQUIT to gracefully end workers.
I understand from #1883/#1449 that there are also other ways for the worker to initiate graceful shutdown, which can trigger the same race condition

### References
Tested via the test app uploaded to #2046
Attempts to fix the same issue seen in #1883

### DIFF
--- a/lib/Mojo/Server/Prefork.pm
+++ b/lib/Mojo/Server/Prefork.pm
@@ -181,7 +181,10 @@ sub _wait {
   while ($chunk =~ /(\d+):(\d)\n/g) {
     next unless my $w = $self->{pool}{$1};
     @$w{qw(healthy time)} = (1, $time) and $self->emit(heartbeat => $1);
-    $w->{graceful} ||= $time if $2;
+    if ($2) {
+      $w->{graceful} ||= $time;
+      $w->{quit}++;
+    }
   }
 }
 


lib/Mojo/Server/Prefork.pm
```diff
@@ -182,6 +182,7 @@ sub _wait {
     next unless my $w = $self->{pool}{$1};
     @$w{qw(healthy time)} = (1, $time) and $self->emit(heartbeat => $1);
     $w->{graceful} ||= $time if $2;
+    $w->{quit}++ if $2;
```
kraih, line 185: That's two redundant `if $2` checks.

marcusramberg:
I like this fix better, except for the if duplication/perltidy.

lib/Mojo/Server/Prefork.pm
brsakai-csco: Ah, yeah, was attempting to match the existing style. I'll put them both in the same block

kraih:
Please squash your commits.

brsakai-csco:
@kraih Done, I think. Let me know if that doesn't look how you expect

kraih:
Is `do not second another SIGQUIT to the worker` (from the commit message) correct english?

brsakai-csco:
I didn't even see that. Must've switched the words in my head without realizing. Something like `do not send a second` > `do not send another` > `do not second another`.

Updated the commit message 👍

kraih (approved):

jixam:
I like this 👍

For completeness, just a note that this does not remove the race condition. It is still possible to hit it by sending two `SIGQUIT` signals close to each other or by sending a single signal when a worker is already shutting down.

To fully fix the race, `$SIG{QUIT} = 'IGNORE';` is needed before `exit 0;` in the worker. However, this is not trivial to do because the handler is `local`ized.

marcusramberg (approved):

PR #2084, opened 2023-07-06
Fix Promise->map hang on false values after concurrency limit

ilmari:
The check for whether to call the callback returned early if the item was falsy.  Instead, it needs to only return early if there are no items left.

### DIFF
--- a/lib/Mojo/Promise.pm
+++ b/lib/Mojo/Promise.pm
@@ -72,7 +72,8 @@ sub map {
   my @wait  = map { $start[0]->clone } 0 .. $#items;
 
   my $start_next = sub {
-    return () unless my $item = shift @items;
+    return () unless @items;
+    my $item = shift @items;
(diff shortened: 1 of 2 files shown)


jberger:
In theory this can be done without adding another line by making the later line https://github.com/mojolicious/mojo/pull/2084/files#diff-d48a9dbb018a19b463cfae82ffb2a5fd0cfe0cf4a553c09f8aba1048aaaaba0aR78 instead say `for shift @$items` but I don't know if that makes the line too long or too inscrutable. @kraih?

jberger:
Either way, I approve

jberger (approved):

ilmari:
> `for shift @$items`

Another option would be `return () unless my ($item) = splice(@items, 0, 1);`, since splice returns an empty list if the input is empty, and list assignment returns the number of items assigned, regardless of their truthiness.

kraih (approved):

PR #2089, opened 2023-07-29
Allow selection of an encoding on slurp/spew in Mojo::File

genio:
### Summary
Alter Mojo::File to:
* Allow slurp and decode with ```$file->slurp('UTF-8');``` etc.
* Allow spew and encode with ```$file->spew('contents', 'UTF-8');``` etc.

### Motivation
As we discussed on IRC, allowing these options not only makes it easier for users as well as bring things more inline with what is allowed in the Mojo.js framework.

### DIFF
--- a/lib/Mojo/File.pm
+++ b/lib/Mojo/File.pm
@@ -14,6 +14,7 @@ use File::stat            ();
 use File::Temp            ();
 use IO::File              ();
 use Mojo::Collection;
+use Mojo::Util qw(decode encode);
 
 our @EXPORT_OK = ('curfile', 'path', 'tempdir', 'tempfile');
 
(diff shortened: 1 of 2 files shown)


t/mojo/file.t
```diff
@@ -342,4 +342,25 @@ subtest 'I/O' => sub {
...
+subtest 'I/O with Encoding' => sub {
+  my $dir  = tempdir;
+  my $file = $dir->child('test.txt')->spew('♥1', 'UTF-8');
+  is $file->slurp('UTF-8'),            '♥1',         'right content';
+  is decode('UTF-8', $file->slurp()),  '♥1',         'right content';
+  is $file->spew('works too!')->slurp, 'works too!', 'right content';
+  $file->spew(encode('UTF-8', '♥1'));
+  is $file->slurp('UTF-8'),           '♥1', 'right content';
+  is decode('UTF-8', $file->slurp()), '♥1', 'right content';
+  {
```
kraih, line 354: These blocks would look better as subtests.

t/mojo/file.t
```diff
@@ -342,4 +342,25 @@ subtest 'I/O' => sub {
   }
 };
 
+subtest 'I/O with Encoding' => sub {
```
kraih, line 345: `Encoding` is upper case.

Grinnz (approved):
Looks quite useful to me. I assume spew is being added because spurt's API can't handle another argument as it would just use it as data.

genio:
@Grinnz Correct. Since spurt can take an array of content, we can't easily check the last argument. Discussion was that spurt would likely be done away with in time

t/mojo/file.t
```diff
@@ -342,4 +342,23 @@ subtest 'I/O' => sub {
...
+  is $file->slurp('UTF-8'),            '♥1',         'right content';
+  is decode('UTF-8', $file->slurp()),  '♥1',         'right content';
+  is $file->spew('works too!')->slurp, 'works too!', 'right content';
+  {
+    eval { $file->spew('♥1') };
+    like $@, qr/Wide character/, 'right error';
+  }
+};
+
+subtest 'I/O with manual encoding' => sub {
```
kraih, line 357: I would move this subtest into the previous one.

marcusramberg (approved):
I'm in favor of this change.

lib/Mojo/File.pm
```diff
@@ -127,23 +128,26 @@ sub sibling {
...
   while ($ret = $file->sysread(my $buffer, 131072, 0)) { $content .= $buffer }
   croak qq{Can't read from file "$$self": $!} unless defined $ret;
 
-  return $content;
+  return $encoding ? decode($encoding, $content) : $content;
 }
 
-sub spurt {
-  my ($self, $content) = (shift, join '', @_);
+sub spew {
```
jhthorsen, line 141: 👍

jhthorsen (approved):
Too bad we have both spew and spurt, but I like spew better, so I'm giving this change 👍

karenetheridge:
This is great; thank you!

Issue #1428, opened 2019-10-30
Mojo::Asset::File::move_to can't move file for successful requests


srchulo:
* Mojolicious version: 8.25
* Perl version: 5.30
* Operating system: CentOS Linux release 7.6.1810 (Core)

I'm not sure if this is actually a Mojolicious issue, or a memory issue. Feel free to close if this is related to my app/hardware and not Mojolicious itself!

I have a [Mojolicious::Lite](https://metacpan.org/pod/Mojolicious::Lite) app that accepts requests to download images and save them to disk, and then resizes them in a subprocess. I make about 10 requests at a time, and I am seeing this intermittent error (about a few out of every 100 requests):

```
Can't move file "/tmp/mojo.tmp.mQfXSsZHXkaHtSmT" to "/tmp/vM1f6_VXjM/original/4112782": No such file or directory at /home/user/perl5/perlbrew/perls/perl-5.30.0/lib/site_perl/5.30.0/Mojo/Asset/File.pm line 109.
```

Here is my app:

```
use Mojolicious::Lite -signatures;
use Mojo::File 'tempdir';

helper ua => sub { state $ua = Mojo::UserAgent->new };

get '/download_image' => sub ($c) {
    my ($url, $id) = map $c->param($_), 'url', 'id';

    $c->ua->get_p($url)->then(sub {
        my ($tx) = @_;

        if (not $tx->res->is_success) {
            die {reason => $tx->error->{message}};
        }

        my $tempdir = tempdir();
        my $original_dir = $tempdir->child('original')->make_path;
        my $image = $original_dir->child($id);

        # ERROR THROWN HERE
        $tx->result->save_to($image);

        # resize into multiple images in subprocess and render from there
        # ...
    })->catch(sub {
        my $err = shift;
        $c->render(json => { success => 0, %$err });
    });

    $c->render_later;
};

app->start;
```

The image urls that give this error are valid and usually work when tried again (and they load in my browser fine). Also, `$tx` says that the request was successful, which makes me think that the image should have been downloaded successfully.

At first I thought that this may be a memory issue (maybe Mojo::Asset::Memory wasn't working because I was using a lot of the memory resizing?), but I saw this line [here](https://mojolicious.org/perldoc/Mojolicious/Guides/Cookbook#Large-file-downloads):

[...]

PR #2093, opened 2023-08-13
Allow decoding of JSON messages in non-standard character sets

codehead:
### Summary
This PR modifies Mojo::Message to allow JSON decoding when a charset is specified. Currently all messages are decoded as UTF-8, so JSON messages with other encodings may fail silently.

### Motivation
Mojo::UserAgent always tries to decode messages in UTF-8 ignoring the message character set. When a message does not meet the UTF-8  specification, `Mojo::JSON::json_decode()` fails silently and `Mojo::UserAgent::json()` returns `undef`. In character sets with a large overlap with UTF-8 such as ISO-8859-1, message decoding fails only when accented characters are present, so JSON messages migh seem empty at random.
This is critical to interface with legacy systems that expose JSON messages in charsets different from UTF-8. 

### References
No public issues or PRs for Mojolicious as far as I'm aware. 
Other Perl frameworks allow non-UTF-8 charsets for JSON messages from very early on -- e.g. [Catalyst::View::JSON](https://github.com/perl-catalyst/Catalyst-View-JSON/commit/49f7babdfc49d781b83094caa0f64f6613079331).
Some Java frameworks defaulted to ISO-8859-1 encoding as recently as 2020: [Content Type being append with charset=ISO-8859-1 #1428](https://github.com/spring-cloud/spring-cloud-contract/issues/1428).
Even though [RFC4627](https://www.rfc-editor.org/rfc/rfc4627) tried to stardardize UTF-8 encoding for JSON messages as far back as 2006 (at around the same time some frameworks implemented charset specification to allow non-ascii characters in JSON) [RFC8259](https://www.rfc-editor.org/rfc/rfc8259#page-8)'s language leaves enough slack for systems that are part of a closed ecosystem.

### DIFF
--- a/lib/Mojo/Message.pm
+++ b/lib/Mojo/Message.pm
@@ -9,7 +9,7 @@ use Mojo::JSON qw(j);
 use Mojo::JSON::Pointer;
 use Mojo::Parameters;
 use Mojo::Upload;
-use Mojo::Util qw(decode);
+use Mojo::Util qw(encode decode);
 
 has content          => sub { Mojo::Content::Single->new };
 has default_charset  => 'UTF-8';
@@ -132,7 +132,8 @@ sub is_limit_exceeded { !!shift->{limit} }
 sub json {
   my ($self, $pointer) = @_;
   return undef if $self->content->is_multipart;
-  my $data = $self->{json} //= j($self->body);
+  my $charset = $self->body_params->charset || $self->default_charset;
+  my $data    = $self->{json} //= j($charset eq 'UTF-8' ? $self->body : encode('UTF-8', decode($charset, $self->body)));
   return $pointer ? Mojo::JSON::Pointer->new($data)->get($pointer) : $data;
 }
 
--- a/t/mojo/response.t
+++ b/t/mojo/response.t
@@ -5,9 +5,10 @@ use Mojo::Asset::File;
 use Mojo::Content::Single;
 use Mojo::Content::MultiPart;
 use Mojo::File qw(tempdir);
-use Mojo::JSON qw(encode_json);
+use Mojo::JSON qw(encode_json to_json);
 use Mojo::Message::Response;
-use Mojo::Util qw(encode gzip);
+use Mojo::Util       qw(encode gzip);
+use Mojo::ByteStream qw(b);
 
 subtest 'Defaults' => sub {
   my $res = Mojo::Message::Response->new;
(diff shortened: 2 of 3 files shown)


lib/Mojo/Message.pm
```diff
@@ -132,7 +132,8 @@ sub is_limit_exceeded { !!shift->{limit} }
 sub json {
   my ($self, $pointer) = @_;
   return undef if $self->content->is_multipart;
-  my $data = $self->{json} //= j($self->body);
+  my $charset = $self->body_params->charset || $self->default_charset;
```
kraih, line 135: I wonder what the performance cost of this is.

lib/Mojo/Message.pm
```diff
@@ -132,7 +132,8 @@ sub is_limit_exceeded { !!shift->{limit} }
...
+  my $data    = $self->{json} //= j($charset eq 'UTF-8' ? $self->body : encode('UTF-8', decode($charset, $self->body)));
```
kraih, line 136: `j(encode(decode()))` seems super inefficient.

jberger:
I'm out of town and reading on my phone so I'm not going to do a full review, but I think I agree with @kraih there are things in there that don't pass the smell test

Grinnz:
Use of ->json is optional. In my opinion if you have a legacy system with nonstandard JSON encoding, you should just use `from_json(decode($charset, $msg->body))`

lib/Mojo/Message.pm
codehead: You are right, `$self->content->charset` is the way to go.

lib/Mojo/Message.pm
codehead: Agreed. Tried this before:
```perl
my $data = $charset eq 'UTF-8' ? j($self->body) : from_json(decode($charset,$self->body));
```
which seems more natural but breaks the last test in t/mojo/response.t subtest "Parse response and extract JSON data"
```
t/mojo/response.t .......................... 1/?
    #   Failed test 'right result'
    #   at t/mojo/response.t line 1075.
    #     Structures begin differing at:
    #          $got->[1] = '2'
    #     $expected->[1] = '4'
    # Looks like you failed 1 test of 9.

#   Failed test 'Parse response and extract JSON data'
#   at t/mojo/response.t line 1076.
# Looks like you failed 1 test of 56.
```
i.e. this one:
```perl
  is_deeply $res->json('/baz'), [1, 2, 3], 'right result';
  $res->json->{baz}[1] = 4;
  is_deeply $res->json('/baz'), [1, 4, 3], 'right result';
```

`j(encode(decode()))` is awkward but avoids that subtle breakage.

codehead:
> Use of ->json is optional. In my opinion if you have a legacy system with nonstandard JSON encoding, you should just use `from_json(decode($charset, $msg->body))`

Thank you @Grinnz . With all due respect, use of ->text is also optional, and it does take into account the charset specified in the message and falls back to a reasonable value should decoding fail. And that syntactic sugar is very, very welcome. I hope the same can be achieved with ->json in the name of cross-framework, cross-language, cross-charset interoperability.

The thing is, you don't know you are dealing with this issue (I still refuse it to call it a bug) until it bites you. On a random message. That might be irreproducible. From a system you don't control. IMHO it would be more useful to have json() die loudly upon trying to decode a non-UTF-8 message than a silent fail.

Grinnz:
> > Use of ->json is optional. In my opinion if you have a legacy system with nonstandard JSON encoding, you should just use `from_json(decode($charset, $msg->body))`
> 
> Thank you @Grinnz . With all due respect, use of ->text is also optional, and it does take into account the charset specified in the message and falls back to a reasonable value should decoding fail. And that syntactic sugar is very, very welcome. I hope the same can be achieved with ->json in the name of cross-framework, cross-language, cross-charset interoperability.

I don't see the comparison. The only purpose of ->text is to decode from the specified charset. The only purpose of ->json is to decode from spec-compliant JSON.

> The thing is, you don't know you are dealing with this issue (I still refuse it to call it a bug) until it bites you. On a random message. That might be irreproducible. From a system you don't control. IMHO it would be more useful to have json() die loudly upon trying to decode a non-UTF-8 message than a silent fail.

There are plenty of issues one may run into in real world use cases, the framework has to balance predicting these with making the common and compliant case simple and efficient, in this case calling two functions seems like a simple option compared to complicating and slowing down the ->json method for all users in the common case where you receive JSON.

jberger:
> > > Use of ->json is optional. In my opinion if you have a legacy system with nonstandard JSON encoding, you should just use `from_json(decode($charset, $msg->body))`
> > 
> > 
> > Thank you @Grinnz . With all due respect, use of ->text is also optional, and it does take into account the charset specified in the message and falls back to a reasonable value should decoding fail. And that syntactic sugar is very, very welcome. I hope the same can be achieved with ->json in the name of cross-framework, cross-language, cross-charset interoperability.
> 
> I don't see the comparison. The only purpose of ->text is to decode from the specified charset. The only purpose of ->json is to decode from spec-compliant JSON.
> 
> > The thing is, you don't know you are dealing with this issue (I still refuse it to call it a bug) until it bites you. On a random message. That might be irreproducible. From a system you don't control. IMHO it would be more useful to have json() die loudly upon trying to decode a non-UTF-8 message than a silent fail.
> 
> There are plenty of issues one may run into in real world use cases, the framework has to balance predicting these with making the common and compliant case simple and efficient, in this case calling two functions seems like a simple option compared to complicating and slowing down the ->json method for all users in the common case where you receive JSON.

I concur with @Grinnz, there is a difference in that "text" is specified to be in a known (configured) charset, but JSON is by definition utf-8 encoded. As such, bytes that are json-like but encoded as some other charset are technically not actually JSON. I know that's not a very satisfactory answer, and it sounds dismissive, but given the difference there it seems not unreasonable to make the spec-compliant mechanism as simple and fast as possible while still making your use-case possible, if not quite as convenient. It seems that that is already the case.

kraih:
I'm afraid it looks like this PR will not pass the vote.

Issue #2094, opened 2023-08-16
v5.38 and module_true: File "/app/x.pl" did not return an application object


melo:
* Mojolicious version: 9.33
* Perl version: 5.38.0
* Operating system: Debian 12.1 (inside a container)

### Steps to reproduce the behavior

This is not a Mojolicious bug but a side effect of using v5.38 feature bundle. I'm creating the issue just to leave a note on this, for others that might see this error message.

Morbo fails to load app with error `File "/app/x.pl" did not return an application object.`

The minimal script:

```perl
use v5.38;
use Mojolicious::Lite -signatures; 
app->start;
```

Run with `morbo`:

```shell
$ morbo x.pl 
Web application available at http://127.0.0.1:3000
File "/app/x.pl" did not return an application object.
```

This works:

```perl
use v5.38;
no feature 'module_true';
use Mojolicious::Lite -signatures; 
app->start;
```

The feature bundle `v5.38` activates the `module_true` feature. This adds an implicit `return 1;` at the end of the script file. To make it work we just need to `unimport` the feature.

Bye,

leonerd:
Does it help if you wrote `return app->start` instead?

kraih:
That sucks. We probably need an FAQ entry for this and recommend against the use of the `v5.38` bundle with Mojolicious.

melo:
> Does it help if you wrote `return app->start` instead?

No, same error.

melo:
> That sucks. We probably need an FAQ entry for this and recommend against the use of the `v5.38` bundle with Mojolicious.

I'll search for the FAQ and send a PR, glad to help.

leonerd:
> That sucks. We probably need an FAQ entry for this and recommend against the use of the `v5.38` bundle with Mojolicious.

You can just

```perl
feature->unimport( 'module_true' ) if $^V ge v5.38;
```

jberger:
> > That sucks. We probably need an FAQ entry for this and recommend against the use of the `v5.38` bundle with Mojolicious.
> 
> You can just
> 
> ```perl
> feature->unimport( 'module_true' ) if $^V ge v5.38;
> ```
> 
> in your main `sub import`

That's a little unfortunate too as then class files will still have to end in `1;` though the users might not expect that (some of them at least). Is there any way we can detect that we're in a ::Lite context (and even as I ask it I'm almost certain the answer is no).

melo:
Using @leonerd suggestion and adding the following

```perl
[...]

PR #2097, opened 2023-08-24
use do rather than require to load a mojo application

haarg:
### Summary
Use `do` rather than `require` to load a mojo application to fix compatibility with new `module_true` core feature.

### Motivation
The return value of `require` is usually not a reasonable thing to rely on, aside from it being true. The first time requiring a file, it will return the value of the last statement in the file. The second time requiring a file, it will return a simple true value.

Mojo was bypassing this problem by deleting the `%INC` entry for the file, forcing it to always be loaded again. But the new `module_true` core feature will cause `require` to always return a simple true value rather than the last statement in the file. This does not apply to `do` though.

### References

#2094

### DIFF
--- a/lib/Mojo/Server.pm
+++ b/lib/Mojo/Server.pm
@@ -54,9 +54,9 @@ sub load_app {
     local @ARGS_OVERRIDE = @args;
 
     # Try to load application from script into sandbox
-    delete $INC{$path};
-    my $app = eval "package Mojo::Server::Sandbox::@{[md5_sum $path]}; require \$path";
-    die qq{Can't load application from file "$path": $@} if $@;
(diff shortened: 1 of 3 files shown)


jberger (approved):
Generally I approve, however I wonder if in the test it should explicitly load the `module_true` pragma rather than the feature bundle, and if it should do so after Mojolicious::Lite. That way if some "fixup" is ever attempted in either of those places we'd still be explicitly testing the pragma not the generic scenario? Maybe both?

jhthorsen (approved):
Cool! Bringing Mojo into the future 👍

kraih:
There is a report that this might have caused a regression: #2109

PR #2111, opened 2023-09-19
fix error reporting when loading applications with compile errors

### LINKED ISSUE #2110: Morbo issues with Mojolicious 9.34 (change in Mojo/Server.pm)
* Mojolicious version: 9.34
* Perl version: 5.34

Inspired by the example in #2094 

```
use v5.34;
$test;
app->start;
```

Then run with `morbo`

### Expected behavior

Something like

```
Web application available at http://127.0.0.1:3000
Can't load application from file "path to script": Global symbol "$test" requires explicit package name (did you forget to declare "my $test"?) at "path to script" line 3.
Compilation failed in require at (eval 93) line 1.
```

### Actual behavior

Instead I get

```
Web application available at http://127.0.0.1:3000
Can't load application from file "path to script": "path to script" did not return a true value at path/lib/perl5/Mojo/Server.pm line 59.
```

Which makes debugging ones own mistakes rather hard.

Note that the problem is related to #2097 which changed 

```
    my $app = eval "package Mojo::Server::Sandbox::@{[md5_sum $path]}; require \$path";
```
in `Mojo/Server.pm` into

```
    my $app = eval "package Mojo::Server::Sandbox::@{[md5_sum $path]}; do \$path";
```
(plus some extra checks).

Manually changing `do` back into `require` makes `morbo` behave like normal.

haarg:
When switching from loading an application from using require to using do, the error was still trying to be captured after a surrounding eval. This error would always be empty, because the error from the do was not rethrown. It would still result in an undef application, so an error would be thrown, but it would not include the actual compile error.

Change the do call inside the eval to throw an error if no application is returned. The error is then recaught outside the eval. Also add a test that shows that the real compilation error is included in the output.

Fixes #2110

### DIFF
--- a/lib/Mojo/Server.pm
+++ b/lib/Mojo/Server.pm
@@ -54,9 +54,11 @@ sub load_app {
     local @ARGS_OVERRIDE = @args;
 
     # Try to load application from script into sandbox
-    my $app = eval "package Mojo::Server::Sandbox::@{[md5_sum $path]}; do \$path";
-    my $err = $app ? undef : $@ || $! || "$path did not return a true value";
-    die qq{Can't load application from file "$path": $err} if $err;
+    my $app = eval sprintf <<'END_CODE', md5_sum($path);
+      package Mojo::Server::Sandbox::%s;
+      do $path or die $@ || $! || "$path did not return a true value";
+END_CODE
+    die qq{Can't load application from file "$path": $@} if $@;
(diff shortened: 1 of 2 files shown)


daleif:
Works fine in my morbo non-lite test.

Grinnz (approved):

jberger (approved):

Issue #2107, opened 2023-09-18
The image helper now produces absolute paths


jixam:
* Mojolicious version: 9.34
* Perl version: 5.36
* Operating system: Ubuntu 22.04 LTS

### Steps to reproduce the behavior
```
perl -E 'use Mojolicious; say join ": ", $Mojolicious::VERSION, Mojolicious->new->build_controller->image("test.png")'
```

### Expected behavior
We noticed this change in behavior which does not seem to be mentioned in Changes.

Before we adjust our application we wanted to check that the change is intended?

### Actual behavior
Previous output:
```
9.33: <img src="test.png">
```

New output:
```
9.34: <img src="/test.png">
```

kraih:
Yes, this is intentional. The generated URL is supposed to be relative to the static directory of the application.

jixam:
All right, thank you for the clarification.

This is quite a breaking change for us as we have been treating the parameter as just the verbatim `src` attribute, doing things like `image("https://example.org/test.png")`. But we will cope :)

kraih:
Absolute URLs are actually supposed to be passed through unchanged. If that doesn't work currently it would be incompatible with mojo.js and therefore a bug.

rawleyfowler:
We're effected by this as well, I'll take a crack at a PR.

rawleyfowler:
Here's an instance of the behavior in one of our projects:

```
src="/https://reasonml.github.io/img/reason.svg"
```

sergiotarxz:
I am getting to work in this.

sergiotarxz:
Should I change this behavior in sub _file_url?

jixam:
I will go ahead and close this issue since it was mostly fixed by #2112.

PR #2112, opened 2023-09-20
Add absolute check to url_for_asset and url_for_file to fix appending "/" to absolute file paths

rawleyfowler:
### Summary
There is currently a bug where:

```
%= image('https://www.somesite.com/image.png')
```

produces

```
<img src="/https://www.somesite.com/image.png">
```

This PR adds an absolute check to the `file_path` and `asset_path` methods that currently breaks this.

### Motivation
This is breaking multiple of my systems in production. (And it seems its breaking others as well)

### References
#2107

### DIFF
--- a/lib/Mojolicious/Controller.pm
+++ b/lib/Mojolicious/Controller.pm
@@ -11,6 +11,8 @@ use Mojo::Util;
 use Mojolicious::Routes::Match;
 use Scalar::Util ();
 
+my $ABSOLUTE = qr!^(?:[^:/?#]+:|//|#)!;
+
 has [qw(app tx)] => undef, weak => 1;
 has match => sub { Mojolicious::Routes::Match->new(root => shift->app->routes) };
 
@@ -239,7 +241,7 @@ sub url_for {
 
   # Absolute URL
   return $target                 if Scalar::Util::blessed $target && $target->isa('Mojo::URL');
-  return Mojo::URL->new($target) if $target =~ m!^(?:[^:/?#]+:|//|#)!;
+  return Mojo::URL->new($target) if $target =~ $ABSOLUTE;
 
   # Base
   my $url  = Mojo::URL->new;
@@ -274,12 +276,12 @@ sub url_for {
 
 sub url_for_asset {
   my ($self, $asset) = @_;
-  return $self->url_for($self->app->static->asset_path($asset));
+  return $self->url_for($asset =~ $ABSOLUTE ? $asset : $self->app->static->asset_path($asset));
 }
 
 sub url_for_file {
   my ($self, $file) = @_;
-  return $self->url_for($self->app->static->file_path($file));
+  return $self->url_for($file =~ $ABSOLUTE ? $file : $self->app->static->file_path($file));
 }
 
 sub write {
--- a/t/mojolicious/static_prefix_lite_app.t
+++ b/t/mojolicious/static_prefix_lite_app.t
@@ -77,6 +77,10 @@ subtest 'File' => sub {
   my $c = $t->app->build_controller;
   is $c->url_for_file('/unknown.css')->path, '/static/unknown.css', 'right file path';
   is $c->url_for_file('/foo/bar.css')->path, '/static/foo/bar.css', 'right file path';
+  is $c->url_for_file('https://somesite.com/file.css')->to_string, 'https://somesite.com/file.css',
+    'right absolute file path?';
+  is $c->url_for_asset('https://somesite.com/file.css')->to_string, 'https://somesite.com/file.css',
+    'right absolute asset path?';
 };
 
 done_testing();


kraih (changes requested):
This duplicates code that already exists in `Mojolicious::Controller`. Just implement it the same way it is done in mojo.js.

rawleyfowler:
I haven't seen any constants in Mojolicious, wasn't sure if you'd like to pull the regex to a constant ala Mojo.js, but besides that this should be 1:1 with Mojo.js.

lib/Mojolicious/Controller.pm
```diff
@@ -274,12 +274,12 @@ sub url_for {
...
 sub url_for_asset {
   my ($self, $asset) = @_;
-  return $self->url_for($self->app->static->asset_path($asset));
+  return $asset =~ m!^(?:[^:/?#]+:|//|#)! ? $asset : $self->url_for($self->app->static->asset_path($asset));
 }
 
 sub url_for_file {
   my ($self, $file) = @_;
-  return $self->url_for($self->app->static->file_path($file));
+  return $file =~ m!^(?:[^:/?#]+:|//|#)! ? $file : $self->url_for($self->app->static->file_path($file));
```
kraih, line 282: Don't copypasta the same regex.

lib/Mojolicious/Static.pm
```diff
@@ -64,7 +64,9 @@ sub file {
   }
 
   # Search DATA
-  if (my $asset = $self->_get_data_file($rel)) { return $asset }
+  if (my $asset = $self->_get_data_file($rel)) {
+    return $asset;
+  }
```
kraih, line 69: This does nothing.

lib/Mojolicious/Static.pm
```diff
@@ -146,7 +148,9 @@ sub warmup {
 
   # DATA sections
   my $index = $self->{index} = {};
-  for my $class (reverse @{$self->classes}) { $index->{$_} = $class for keys %{data_section $class} }
+  for my $class (reverse @{$self->classes}) {
+    $index->{$_} = $class for keys %{data_section $class};
+  }
```
kraih, line 153: Same, does nothing. Why is it included?

rawleyfowler: My perltidy runs on save. It must think that is not valid given the perltidyrc, I'll revert it.

lib/Mojolicious/Controller.pm
```diff
@@ -274,12 +275,12 @@ sub url_for {
 
 sub url_for_asset {
   my ($self, $asset) = @_;
-  return $self->url_for($self->app->static->asset_path($asset));
+  return $asset =~ ABSOLUTE ? $asset : $self->url_for($self->app->static->asset_path($asset));
```
kraih, line 278: This is wrong, in Perl we need to return a `Mojo::URL` object. Just let `url_for` handle it.

lib/Mojolicious/Controller.pm
```diff
@@ -274,12 +275,12 @@ sub url_for {
...
 }
 
 sub url_for_file {
   my ($self, $file) = @_;
-  return $self->url_for($self->app->static->file_path($file));
+  return $file =~ ABSOLUTE ? $file : $self->url_for($self->app->static->file_path($file));
```
kraih, line 283: Same.

lib/Mojolicious/Controller.pm
```diff
@@ -10,6 +10,7 @@ use Mojo::URL;
 use Mojo::Util;
 use Mojolicious::Routes::Match;
 use Scalar::Util ();
+use constant ABSOLUTE => qr!^(?:[^:/?#]+:|//|#)!;
```
kraih, line 13: We separate constants from imports.

kraih (changes requested):

t/mojolicious/static_prefix_lite_app.t
```diff
@@ -75,8 +75,9 @@ subtest 'UTF-8 encoded inline file' => sub {
 
 subtest 'File' => sub {
   my $c = $t->app->build_controller;
-  is $c->url_for_file('/unknown.css')->path, '/static/unknown.css', 'right file path';
-  is $c->url_for_file('/foo/bar.css')->path, '/static/foo/bar.css', 'right file path';
+  is $c->url_for_file('/unknown.css')->path,            '/static/unknown.css',           'right file path';
+  is $c->url_for_file('/foo/bar.css')->path,            '/static/foo/bar.css',           'right file path';
+  is $c->url_for_file('https://somesite.com/file.css'), 'https://somesite.com/file.css', 'right absolute file path?';
```
kraih, line 80: You also changed `url_for_asset`, but did not add any new tests for that.

jixam:
FWIW, I added this patch to our build and our tests started passing again. So, works for me :)

lib/Mojolicious/Controller.pm
```diff
@@ -11,6 +11,8 @@ use Mojo::Util;
 use Mojolicious::Routes::Match;
 use Scalar::Util ();
 
+use constant ABSOLUTE => qr!^(?:[^:/?#]+:|//|#)!;
```
kraih, line 14: Make this an uppercase variable instead of a constant, then it is not a potential namespace pollution problem for subclasses.

kraih (approved):

jhthorsen (approved):
LGTM 👍

Issue #2105, opened 2023-09-17
One t-test fails, Perl 5.38 on Windows 10


PhilterPaper:
* Mojolicious version: 9.34
* Perl version: 5.38.0.1
* Operating system: Windows 10

### Steps to reproduce the behavior
Installation via 'cpan' to update from CPAN

### Expected behavior
Expect clean install

### Actual behavior
Failed one t-test: t/mojo/file.t 'right file' (several occurrences listed at 285, 288, 293, 299, 308)

PhilterPaper:
If it matters, this is the new Strawberry Perl build ("portable" version). I took a quick look in file.t (list/list_tree), and there were many places mentioning "right files", so I'm not sure which was failing.

kraih:
> I took a quick look in file.t (list/list_tree), and there were many places mentioning "right files", so I'm not sure which was failing.

It tells you the lines right in the fail messages. 😄 

And we test with Strawberry Perl in our [GitHub Actions](https://github.com/mojolicious/mojo/blob/main/.github/workflows/windows.yml). Everything looks fine there.

PhilterPaper:
> And we test with Strawberry Perl in our [GitHub Actions](http://github.com/mojolicious/mojo/blob/main/.github/workflows/windows.yml). Everything looks fine there.

Well, I see that CPAN Testers [failed](http://matrix.cpantesters.org/?dist=Mojolicious%209.34;os=mswin32;perl=5.38.0;reports=1) exactly the same way on the same Perl and Windows level, so perhaps it would be a good idea to look at it again. If you don't have a Strawberry Perl to install on (not the GitHub Actions version), I would be happy to try installing an instrumented version of Mojolicious, or force an install and try running file.t and other debug on it, and report back the results to you. Let me know.

PhilterPaper:
9.35 fails on the same test. CPAN testers report same problem.

Since the Mojolicious developer(s) apparently don't have a Windows/Strawberry Perl 5.38 system to test on, please feel free to contact me if I can be of any help in testing for you.

PhilterPaper:
Ditto release 9.36. CPAN testers apparently haven't gotten to it yet.

dcsjulio:
Same here. Build fails using strawberry perl.

I suspect that the reason for that is that File::Find works with slashes, while canonpath returns backslashes on windows.

So when doing the delete of the current directory on the new sub, it fails because that key does not exist.

PhilterPaper:
[...]

PR #2113, opened 2023-09-20
Fix for handling changes in File::Find 1.41 on win

mikemagowan:
### Summary
In a relatively recent update to File::Find v1.41 (which was included in the Perl 5.38 release) a change was made to how the windows directory separator (backslash) is handled and is now converted within File::Find to a forward slash to make it consistent.  This was done in the below commit:

https://github.com/Perl/perl5/commit/414f14df98cb1c9a20f92c5c54948b67c09f072d

This now causes several tests to fail in Mojo::File with Perl 5.38 on Windows.

### Motivation
Mojolicious will not past tests on Windows with Perl 5.38.

### References
https://github.com/mojolicious/mojo/issues/2105


jhthorsen (commented):
Can a test be written for this?

mikemagowan:
There are two consequences of this bug on the returned collection:

1. The requested directory is included in the returned collection as the delete fails to match the key
2. The sort order of the collection changes (backslashes vs slashes)

When I considered a test for the two things above it felt redundant as this is already tested in the tests that fail.

I'm open to suggestion though if you have a better idea for a test for this?

lib/Mojo/File.pm
```diff
@@ -63,17 +63,18 @@ sub list_tree {
 
   # The File::Find documentation lies, this is needed for CIFS
   local $File::Find::dont_use_nlink = 1 if $options->{dont_use_nlink};
-
+
```
kraih, line 66: Why does this empty line change?

mikemagowan: Not sure how that whitespace crept in - now removed

kraih:
Squash your commits.

PR #2148, opened 2024-02-09
dont overwrite tls_option

tagg:
### Summary
tls_option(SSL_verifycn_name) is unconditionally overwritten.

### Motivation
Mojo::UserAgent tls_option() is used to pass extra options to IO::Socket::SSL, but the code in Mojo::IOLoop::TLS unconditionally overwrites any SSL_verifycn_name the user is trying to pass.

### References
As discussed on IRC.

I wanted to add a test case, but I cannot run the TEST_TLS tests. Being a very simple change, maybe a test is unneccessary.

### DIFF
--- a/lib/Mojo/IOLoop/TLS.pm
+++ b/lib/Mojo/IOLoop/TLS.pm
@@ -59,7 +59,7 @@ sub _expand {
   }
   else {
     $tls->{SSL_hostname}      = IO::Socket::SSL->can_client_sni ? $args->{address} : '';
-    $tls->{SSL_verifycn_name} = $args->{address};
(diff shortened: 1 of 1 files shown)


kraih:
I'm afraid without a test this probably won't be considered.

kraih:
Don't think any TLS changes should be considered until our tests pass in github actions again. https://github.com/mojolicious/mojo/commit/5d83a17e79b71bc3bbd492adc13b5942f1a61ac6

tagg:
Fixed in f97eee9dbd619023e54839bddd01d2e286b91ee1.

PR #2150, opened 2024-02-11
Adding tests for json_* subroutines in Test::Mojo.

sergiotarxz:
### Summary
Adding tests for json_* subroutines in Test::Mojo.

### Motivation
Issue #1488

### References
 #1488


sergiotarxz:
The result for `json_message_has` in the negative case is:

```
is_deeply \@args, ['ok', undef, 'has value for JSON Pointer "/test0"'];
```

Which appears sightly inconsistent with other negative cases being !!0 including `json_has`, probably not an issue but I think it is good to be known.

PR #2151, opened 2024-02-11
Create SECURITY.md.

sergiotarxz:
Create SECURITY.md to avoid security researchers to dive into the actual documentation searching for the correct way to send a found security bug.

### Summary
Adding a file which contains where to report security issues.

### Motivation
Because security researchers could otherwise findd a hard time finding for the correct email.

### References
Github suggests it also.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,5 @@
+# Security Policy
+
+## Reporting a Vulnerability
+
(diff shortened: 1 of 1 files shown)


sergiotarxz:
Fixes #2152. Maybe Mojolicious contributors want to add something else to this draft file.

kraih:
The core team will write the content if the file is deemed necessary.

PR #2173, opened 2024-04-05
Improve include time of Mojo::Base by extracting monkey_patch

okurz:
### Motivation
Commit 7e9a2ad introduced a "require Mojo::Util" causing a significant chain of further dependencies being pulled in which IMHO should be avoided for the very base module which is in particular being advertised as useful for just enabling static code checks and common import checks.

### Summary
This commit moves out the function "monkey_patch" into its own module to break or prevent the circular dependency between Mojo::Base and Mojo::Util.

With that the import of Mojo::Base is more efficient, `time perl -e 'use Mojo::Base`
on my system reduced from 224±12.08 ms to 52.0±2.3 ms which I consider a considerable improvement for Mojo::Base which is used as a baseclass in many cases.

### DIFF
--- a/lib/Mojo/Base.pm
+++ b/lib/Mojo/Base.pm
@@ -7,11 +7,9 @@ use feature ':5.16';
 use mro;
 
 # No imports because we get subclassed, a lot!
-use Carp         ();
-use Scalar::Util ();
-
-# Defer to runtime so Mojo::Util can use "-strict"
-require Mojo::Util;
+use Carp           ();
+use Scalar::Util   ();
+use Mojo::BaseUtil ();
 
 # Role support requires Role::Tiny 2.000001+
 use constant ROLES => !!(eval { require Role::Tiny; Role::Tiny->VERSION('2.000001'); 1 });
@@ -41,7 +39,7 @@ sub attr {
         ref $self->{$_} and Scalar::Util::weaken $self->{$_} for @$names;
         return $self;
       };
-      Mojo::Util::monkey_patch(my $base = $class . '::_Base', 'new', $sub);
+      Mojo::BaseUtil::monkey_patch(my $base = $class . '::_Base', 'new', $sub);
       no strict 'refs';
       unshift @{"${class}::ISA"}, $base;
     }
@@ -90,7 +88,7 @@ sub attr {
     else {
       $sub = sub { return $_[0]{$attr} if @_ == 1; $_[0]{$attr} = $_[1]; $_[0] };
     }
-    Mojo::Util::monkey_patch($class, $attr, $sub);
+    Mojo::BaseUtil::monkey_patch($class, $attr, $sub);
   }
 }
 
@@ -110,7 +108,7 @@ sub import {
     # Role
     elsif ($flag eq '-role') {
       Carp::croak 'Role::Tiny 2.000001+ is required for roles' unless ROLES;
-      Mojo::Util::monkey_patch($caller, 'has', sub { attr($caller, @_) });
+      Mojo::BaseUtil::monkey_patch($caller, 'has', sub { attr($caller, @_) });
       eval "package $caller; use Role::Tiny; 1" or die $@;
     }
 
@@ -131,9 +129,9 @@ sub import {
     # Module
     elsif ($flag !~ /^-/) {
       no strict 'refs';
-      require(Mojo::Util::class_to_path($flag)) unless $flag->can('new');
+      require(Mojo::BaseUtil::class_to_path($flag)) unless $flag->can('new');
       push @{"${caller}::ISA"}, $flag;
-      Mojo::Util::monkey_patch($caller, 'has', sub { attr($caller, @_) });
+      Mojo::BaseUtil::monkey_patch($caller, 'has', sub { attr($caller, @_) });
     }
 
     elsif ($flag ne '-strict') { Carp::croak "Unsupported flag: $flag" }
--- a/lib/Mojo/BaseUtil.pm
+++ b/lib/Mojo/BaseUtil.pm
@@ -1,0 +1,46 @@
+package Mojo::BaseUtil;
+
+# Only using pure Perl as the only purpose of this module is to break a circular dependency involving Mojo::Base
+use strict;
+use warnings;
+use feature ':5.16';
+
+use Exporter  qw(import);
+use Sub::Util qw(set_subname);
+
+our @EXPORT_OK = (qw(class_to_path monkey_patch));
+
+sub class_to_path { join '.', join('/', split(/::|'/, shift)), 'pm' }
+
+sub monkey_patch {
+  my ($class, %patch) = @_;
+  no strict 'refs';
+  no warnings 'redefine';
+  *{"${class}::$_"} = set_subname("${class}::$_", $patch{$_}) for keys %patch;
+}
+
+1;
+
+=encoding utf8
+
+=head1 NAME
+
+Mojo::BaseUtil - Common utility functions used in Mojo::Base, re-exported in Mojo::Util
+
+=head1 SYNOPSIS
+
+  use Mojo::BaseUtil qw(class_to_patch monkey_path);
+
+  my $path = class_to_path 'Foo::Bar';
+  monkey_patch 'MyApp', foo => sub { say 'Foo!' };
+
+=head1 DESCRIPTION
+
+L<Mojo::BaseUtil> provides functions to both L<Mojo::Base> and L<Mojo::Util> so that C<Mojo::Base> does not have to load
+the rest of L<Mojo::Util> while preventing a circular dependency.
+
+=head1 SEE ALSO
+
+L<Mojolicious>, L<Mojolicious::Guides>, L<https://mojolicious.org>.
+
+=cut
--- a/lib/Mojo/Util.pm
+++ b/lib/Mojo/Util.pm
@@ -14,9 +14,9 @@ use IO::Poll qw(POLLIN POLLPRI);
 use IO::Uncompress::Gunzip;
 use List::Util         qw(min);
 use MIME::Base64       qw(decode_base64 encode_base64);
+use Mojo::BaseUtil     qw(class_to_path monkey_patch);
 use Pod::Usage         qw(pod2usage);
 use Socket             qw(inet_pton AF_INET6 AF_INET);
-use Sub::Util          qw(set_subname);
 use Symbol             qw(delete_package);
 use Time::HiRes        ();
 use Unicode::Normalize ();
@@ -104,8 +104,6 @@ sub class_to_file {
   $class =~ s/([A-Z])([A-Z]*)/$1 . lc $2/ge;
   return decamelize($class);
 }
-
-sub class_to_path { join '.', join('/', split(/::|'/, shift)), 'pm' }
 
 sub decamelize {
   my $str = shift;
@@ -196,13 +194,6 @@ sub humanize_bytes {
   return $prefix . _round($size) . 'MiB' if ($size /= 1024) < 1024;
   return $prefix . _round($size) . 'GiB' if ($size /= 1024) < 1024;
   return $prefix . _round($size /= 1024) . 'TiB';
-}
-
-sub monkey_patch {
-  my ($class, %patch) = @_;
-  no strict 'refs';
-  no warnings 'redefine';
-  *{"${class}::$_"} = set_subname("${class}::$_", $patch{$_}) for keys %patch;
 }
 
 sub network_contains {
--- a/t/mojo/base_util.t
+++ b/t/mojo/base_util.t
@@ -1,0 +1,48 @@
+use Mojo::Base -strict;
+
+use Test::More;
+use Sub::Util qw(subname);
+
+use Mojo::BaseUtil qw(class_to_path monkey_patch);
+
+subtest 'class_to_path' => sub {
+  is Mojo::BaseUtil::class_to_path('Foo::Bar'),      'Foo/Bar.pm',     'right path';
+  is Mojo::BaseUtil::class_to_path("Foo'Bar"),       'Foo/Bar.pm',     'right path';
+  is Mojo::BaseUtil::class_to_path("Foo'Bar::Baz"),  'Foo/Bar/Baz.pm', 'right path';
+  is Mojo::BaseUtil::class_to_path("Foo::Bar'Baz"),  'Foo/Bar/Baz.pm', 'right path';
+  is Mojo::BaseUtil::class_to_path("Foo::Bar::Baz"), 'Foo/Bar/Baz.pm', 'right path';
+  is Mojo::BaseUtil::class_to_path("Foo'Bar'Baz"),   'Foo/Bar/Baz.pm', 'right path';
+};
+
+subtest 'monkey_patch' => sub {
+  {
+
+    package MojoMonkeyTest;
+    sub foo {'foo'}
+  }
+  ok !!MojoMonkeyTest->can('foo'), 'function "foo" exists';
+  is MojoMonkeyTest::foo(), 'foo', 'right result';
+  ok !MojoMonkeyTest->can('bar'), 'function "bar" does not exist';
+  monkey_patch 'MojoMonkeyTest', bar => sub {'bar'};
+  ok !!MojoMonkeyTest->can('bar'), 'function "bar" exists';
+  is MojoMonkeyTest::bar(), 'bar', 'right result';
+  monkey_patch 'MojoMonkeyTest', foo => sub {'baz'};
+  ok !!MojoMonkeyTest->can('foo'), 'function "foo" exists';
+  is MojoMonkeyTest::foo(), 'baz', 'right result';
+  ok !MojoMonkeyTest->can('yin'),  'function "yin" does not exist';
+  ok !MojoMonkeyTest->can('yang'), 'function "yang" does not exist';
+  monkey_patch 'MojoMonkeyTest',
+    yin  => sub {'yin'},
+    yang => sub {'yang'};
+  ok !!MojoMonkeyTest->can('yin'), 'function "yin" exists';
+  is MojoMonkeyTest::yin(), 'yin', 'right result';
+  ok !!MojoMonkeyTest->can('yang'), 'function "yang" exists';
+  is MojoMonkeyTest::yang(), 'yang', 'right result';
+};
+
+subtest 'monkey_patch (with name)' => sub {
+  is subname(MojoMonkeyTest->can('foo')), 'MojoMonkeyTest::foo', 'right name';
+  is subname(MojoMonkeyTest->can('bar')), 'MojoMonkeyTest::bar', 'right name';
+};
+
+done_testing();
--- a/t/pod_coverage.t
+++ b/t/pod_coverage.t
@@ -11,4 +11,7 @@ my @await = (
   qw(AWAIT_NEW_FAIL AWAIT_ON_CANCEL AWAIT_ON_READY AWAIT_WAIT)
 );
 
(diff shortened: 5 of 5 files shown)


lib/Mojo/Util.pm
```diff
@@ -75,6 +75,13 @@ our @EXPORT_OK = (
   qw(unquote url_escape url_unescape xml_escape xor_encode)
 );
 
+
+# redirecting to MonkeyPatch::monkey_patch for backward compatibility
+    no warnings 'redefine';
+sub monkey_patch {
```
guest20, line 81: You already import it, why re-define it here?

perlpunk: If it is imported via `use` then it's also defined in the Mojo::Util namespace, so you can export it.

jberger: you should be able to re-export symbols, I don't think that should be a problem

okurz: will do that then :)

lib/Mojo/Base.pm
```diff
@@ -10,8 +10,7 @@ use mro;
 use Carp         ();
 use Scalar::Util ();
 
-# Defer to runtime so Mojo::Util can use "-strict"
-require Mojo::Util;
+use Mojo::MonkeyPatch qw(monkey_patch);
```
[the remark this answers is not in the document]
okurz: ok, so `require Mojo::MonkeyPatch` then?

okurz:
* Addressed CI failure "Undefined subroutine &Mojo::Util::class_to_path" by moving class_to_path to Mojo::Base and adjusting the rest of the repository accordingly, re-exporting from Mojo::Util
* Removed the re-implementation of monkey_patch in Mojo::Util, directly re-exporting
* Replaced "use" by "require" as per https://github.com/mojolicious/mojo/pull/2173#discussion_r1553828119
* Fixed tests
* Added POD in Mojo::MonkeyPatch

lib/Mojo/Base.pm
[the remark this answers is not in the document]
Grinnz: `use Mojo::MonkeyPatch ();` is sufficient, require was only used to allow the circular dependency.

Grinnz:
> * Addressed CI failure "Undefined subroutine &Mojo::Util::class_to_path" by moving class_to_path to Mojo::Base and adjusting the rest of the repository accordingly, re-exporting from Mojo::Util

This should not be done. This creates a method in all classes using Mojo::Base as a base.

Grinnz:
Additionally, if Mojo::Util re-exports class_to_path no other adjustment should be necessary.

okurz:
> > * Addressed CI failure "Undefined subroutine &Mojo::Util::class_to_path" by moving class_to_path to Mojo::Base and adjusting the rest of the repository accordingly, re-exporting from Mojo::Util
> 
> This should not be done. This creates a method in all classes using Mojo::Base as a base.

ok, do you have a preferred alternative? Move to yet another new module with a single function, combine with Mojo::MonkeyPatch in BaseUtils (???), re-implement the one-liner?

lib/Mojo/Base.pm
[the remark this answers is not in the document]
okurz: Understood, done

okurz:
> > * Addressed CI failure "Undefined subroutine &Mojo::Util::class_to_path" by moving class_to_path to Mojo::Base and adjusting the rest of the repository accordingly, re-exporting from Mojo::Util
> 
> This should not be done. This creates a method in all classes using Mojo::Base as a base.

done, updated

Grinnz:
BaseUtil seems reasonable and self-explanatory to me, but there may be other opinions. As long as Mojo::Util imports and re-exports the functions from there correctly, there should be no changes needed to other modules.

okurz:
> BaseUtil seems reasonable and self-explanatory to me, but there may be other opinions. As long as Mojo::Util imports and re-exports the functions from there correctly, there should be no changes needed to other modules.

Sure, let's see how that looks. Added a commit to combine monkey_patch and class_to_path in Mojo::BaseUtil and reverted changes in other modules to use still the functions as exported from Mojo::Util.

lib/Mojo/BaseUtil.pm
```diff
@@ -0,0 +1,71 @@
+package Mojo::BaseUtil;
+use Mojo::Base -strict;
```
Grinnz, line 2: This will either need to be replaced with the same pragmas as Mojo::Base uses, or Mojo::Base will need to require Mojo::BaseUtil as it did Mojo::Util previously, because of the circular require.

lib/Mojo/BaseUtil.pm
```diff
@@ -0,0 +1,71 @@
...
+
+use Exporter  qw(import);
+use Sub::Util qw(set_subname);
+
+our @EXPORT_OK = (qw(monkey_patch));
```
Grinnz, line 7: Missing class_to_path export

Grinnz:
IMO, t/mojo/util.t should remain unchanged, to ensure there is no change in compatibility for the re-exported functions. t/mojo/base_util.t may be unnecessary if so.

kraih:
I'd be interested in some real world benchmarks where this makes a difference.

poti1:
I'm looking forwards to seeing this hopefully soon in the library :)

poti1:
By simply copying over `monkey_patch` and the `set_subname` declarations from Mojo::Util to my cli tool (and removing use/require Mojo::Util), I saw these results in performance improvement:

BEFORE:
real    0m0.179s
user    0m0.101s
sys     0m0.052s

NOW:
real    0m0.045s
user    0m0.018s
sys     0m0.015s

When running on my phone's termux, the change of those values is really felt.

lib/Mojo/BaseUtil.pm
```diff
@@ -0,0 +1,71 @@
...
+=head1 SYNOPSIS
+
+  use Mojo::BaseUtil qw(class_to_patch monkey_path);
+
+  my $path = class_to_path 'Foo::Bar';
+  monkey_patch 'MyApp', foo => sub { say 'Foo!' };
+
+=head1 DESCRIPTION
+
+L<Mojo::BaseUtil> provides a C<class_to_path> and the C<monkey_patch> function for L<Mojo>.
```
Grinnz, line 36: It may be worth detailing here the purpose of this module: to provide functions for both Mojo::Base and Mojo::Util, so Mojo::Base does not have to load the rest of Mojo::Util.

okurz: That's a good idea. I more or less included your sentence literally as I think it's already phrased in the best understandable way.

okurz:
* Updated POD for description
* Fixed one CI failure

lib/Mojo/BaseUtil.pm
okurz: Currently I used in Base `use Mojo::BaseUtil ();` and CI tests passed. IIUC that means we handle the circular require already. Or should I change it to require in Mojo::Base still? Can you elaborate?

jberger: I think it makes sense for this to just `use strict; use warnings;` IMO. The whole point is to make it extremely clear that this package breaks circular deps.

okurz: got it, done

okurz:
* Using only `strict+warnings` in BaseUtils to address https://github.com/mojolicious/mojo/pull/2173#discussion_r1594074777

kraih:
Please squash commits.

lib/Mojo/BaseUtil.pm
Grinnz: > Currently I used in Base `use Mojo::BaseUtil ();` and CI tests passed. IIUC that means we handle the circular require already. Or should I change it to require in Mojo::Base still? Can you elaborate?

This will still fail to use strict and warnings and the other pragmas which -strict is supposed to import, because of the circular require. Specifically, Mojo::BaseUtil is loaded during the compilation phase of Mojo::Base, and attempts to load Mojo::Base which has not finished loading and so silently ignores its import method.

okurz:
> Please squash commits.

done

lib/Mojo/BaseUtil.pm
```diff
@@ -0,0 +1,78 @@
+package Mojo::BaseUtil;
+
+# only using pure Perl as the only purpose of this module is to break a
```
kraih, line 3: Comment capitalisation. And our line length is 120 chars.

t/mojo/util.t
```diff
@@ -6,9 +6,8 @@ use lib curfile->sibling('lib')->to_string;
 use Test::More;
 use Mojo::ByteStream qw(b);
 use Mojo::DeprecationTest;
-use Sub::Util qw(subname);
```
Grinnz, line 9: The changes to this test file are inconsistent, and I think this test file could actually be left alone so as to make sure that the re-exports always remain functional.

lib/Mojo/Util.pm
```diff
@@ -67,6 +67,7 @@ my $ENTITY_RE = qr/&(?:\#((?:[0-9]{1,7}|x[0-9a-fA-F]{1,6}));|(\w+[;=]?))/;
 # Encoding and pattern cache
 my (%ENCODING, %PATTERN);
 
+# re-exporting BaseUtil functions for backward compatibility
```
Grinnz, line 70: I would consider these re-exports a permanent functionality matter, not backwards compatibility - I don't see a reason to ever remove these from Mojo::Util. This comment isn't really needed

okurz:
Covered all last three open review comments

Grinnz (approved):
Looks good to me

lib/Mojo/Util.pm
```diff
@@ -14,9 +14,9 @@ use IO::Poll qw(POLLIN POLLPRI);
 use IO::Uncompress::Gunzip;
 use List::Util         qw(min);
 use MIME::Base64       qw(decode_base64 encode_base64);
+use Mojo::BaseUtil     qw(class_to_path monkey_patch);
 use Pod::Usage         qw(pod2usage);
 use Socket             qw(inet_pton AF_INET6 AF_INET);
-use Sub::Util          qw(set_subname);
```
jberger, line 19: Do you think anyone depends on Mojo::Util::set_subname? They probably shouldn't but they could

t/mojo/base_util.t
```diff
@@ -0,0 +1,49 @@
...
+  is Mojo::BaseUtil::class_to_path("Foo'Bar"),       'Foo/Bar.pm',     'right path';
+  is Mojo::BaseUtil::class_to_path("Foo'Bar::Baz"),  'Foo/Bar/Baz.pm', 'right path';
+  is Mojo::BaseUtil::class_to_path("Foo::Bar'Baz"),  'Foo/Bar/Baz.pm', 'right path';
+  is Mojo::BaseUtil::class_to_path("Foo::Bar::Baz"), 'Foo/Bar/Baz.pm', 'right path';
+  is Mojo::BaseUtil::class_to_path("Foo'Bar'Baz"),   'Foo/Bar/Baz.pm', 'right path';
+};
+
+subtest 'monkey_patch' => sub {
+  {
+
```
jberger, line 20: nitpick but blank line?

lib/Mojo/Util.pm
Grinnz: I would not consider use of any subroutines imported without a documented interface as our responsibility.

jhthorsen:
I think the name Mojo::BaseUtil is a bit awkward. How to tell what is in which package? I can't come up with a killer name, but I'm thinking about Mojo::CoreUtil or Mojo::PureUtil, which might indicate that the functions in that package is without any *significant* dependencies? I guess Mojo::BaseUtil makes sense, if we only have functions that Mojo::Base depends on, but I wonder if it's a slippery slope...

Should any of the following functions be moved over? The reason I ask, is because maybe it makes sense to move class_to_file(), when class_to_path() was moved.

* camelize()
* class_to_file()
* decamelize()

Grinnz:
While that might make sense for code level consistency, I think it's beneficial to keep the module limited in scope to its stated purpose. All of these functions should continue to be used from Mojo::Util in normal cases. It can be viewed less as "moving" the functions from the user's POV because they continue to be re-exported.

jhthorsen:
Then I think that should be stated in the documentation, meaning that functions can be added and removed in the future to keep Mojo::Base quick to load.

Grinnz:
It could make sense to me to remove the function documentation from the new module (these would need to be added as trusted in t/pod_coverage.t), and document that it provides functions for use in Mojo::Base and shouldn't be depended on to remain compatible.

okurz:
> It could make sense to me to remove the function documentation from the new module

ok, fine for me. As there was already one "+1" I followed that suggestion and included that change

Also included is a second commit removing redundant blank lines in the added code as well as where I originally copied the test code from.

t/mojo/base_util.t
okurz: ok, done. Included is a second commit removing redundant blank lines in the added code as well as where I originally copied the test code from.

poti1:
For my own cli script, I ended up doing this to avoid loading in thousands of subs right away:

```perl
sub monkey_patch {
    my ( $class, %patch ) = @_;
    require Sub::Util;
    for ( keys %patch ) {
        *{"${class}::$_"} =
          Sub::Util::set_subname( "${class}::$_", $patch{$_} );
    }
}
```

t/mojo/base_util.t
okurz: Nope, perltidy actually demands the blank line :D Reverted the last change

lib/Mojo/BaseUtil.pm
```diff
@@ -0,0 +1,48 @@
...
+# Only using pure Perl as the only purpose of this module is to break a circular dependency involving Mojo::Base
+use strict;
+use warnings;
+
+use Exporter  qw(import);
+use Sub::Util qw(set_subname);
+
+our @EXPORT_OK = (qw(class_to_path monkey_patch));
+
+
```
kraih, line 12: Nitpick: Extra newline.

lib/Mojo/BaseUtil.pm
```diff
@@ -0,0 +1,48 @@
...
+  my $path = class_to_path 'Foo::Bar';
+  monkey_patch 'MyApp', foo => sub { say 'Foo!' };
+
+=head1 DESCRIPTION
+
+L<Mojo::BaseUtil> provides a C<class_to_path> and the C<monkey_patch> function
+for L<Mojo>. The main purpose is to provide functions to both C<Mojo::Base>
+and C<Mojo::Util> so that C<Mojo::Base> does not have to load the rest of
+C<Mojo::Util>.
```
kraih, line 42: Nitpick: Not 120 char lines.

lib/Mojo/Util.pm
kraih: No documentation, no feature.

t/mojo/base_util.t
```diff
@@ -0,0 +1,49 @@
+use Mojo::Base -strict;
+
+use Test::More;
+use Sub::Util qw(subname);
+
+use Mojo::BaseUtil qw(class_to_path monkey_patch);
+
+
+subtest 'class_to_path' => sub {
```
kraih, line 9: Nitpick: Extra newline.

t/pod_coverage.t
```diff
@@ -11,4 +11,7 @@ my @await = (
   qw(AWAIT_NEW_FAIL AWAIT_ON_CANCEL AWAIT_ON_READY AWAIT_WAIT)
 );
 
-all_pod_coverage_ok({also_private => ['BUILD_DYNAMIC', @await, 'spurt']});
+# base utils only to be used in Mojo::Base and should not be used elsewhere
```
kraih, line 14: Nitpick: Comment capitalisation.

kraih: Is that correct english? Sounds a little wrong to me.

okurz: Reworded to 
> These are base utils only to be used in Mojo::Base and not elsewhere

lib/Mojo/BaseUtil.pm
```diff
@@ -0,0 +1,46 @@
...
+=head1 SYNOPSIS
+
+  use Mojo::BaseUtil qw(class_to_patch monkey_path);
+
+  my $path = class_to_path 'Foo::Bar';
+  monkey_patch 'MyApp', foo => sub { say 'Foo!' };
+
+=head1 DESCRIPTION
+
+L<Mojo::BaseUtil> provides a C<class_to_path> and the C<monkey_patch> function for L<Mojo>. The main purpose is to
```
kraih, line 38: That phrase sounds completely wrong, it's two functions and not for `Mojo`.

lib/Mojo/BaseUtil.pm
```diff
@@ -0,0 +1,46 @@
...
+  *{"${class}::$_"} = set_subname("${class}::$_", $patch{$_}) for keys %patch;
+}
+
+1;
+
+=encoding utf8
+
+=head1 NAME
+
+Mojo::BaseUtil - Common utility functions used in Mojo::Base, re-exported in Mojo::Util.
```
kraih, line 27: No punctuation for the package description.

lib/Mojo/BaseUtil.pm
okurz: Rephrased. Please check again.

lib/Mojo/BaseUtil.pm
```diff
@@ -0,0 +1,45 @@
...
+=head1 SYNOPSIS
+
+  use Mojo::BaseUtil qw(class_to_patch monkey_path);
+
+  my $path = class_to_path 'Foo::Bar';
+  monkey_patch 'MyApp', foo => sub { say 'Foo!' };
+
+=head1 DESCRIPTION
+
+L<Mojo::BaseUtil> provides functions to both C<Mojo::Base> and C<Mojo::Util> so that C<Mojo::Base> does not have to load
```
kraih, line 38: You've used `C<...>` instead of `L<...>` for the modules, otherwise it sounds good.

okurz: Done

eduardoj:
Small typo in commit message and in pull request description: `strictures` shold be `structures`.

lib/Mojo/BaseUtil.pm
```diff
@@ -0,0 +1,45 @@
+package Mojo::BaseUtil;
+
+# Only using pure Perl as the only purpose of this module is to break a circular dependency involving Mojo::Base
+use strict;
```
Grinnz, line 4: Minor detail, I think we should `use v5.16;` here to include the feature bundle as Mojo::Base -strict would. The rest of the things it imports are unlikely to be important here.

Grinnz: or alternatively, add `use feature ':5.16';` to match Mojo::Base's declarations.

okurz: So I now did the same as in lib/Mojo/Base.pm

```
use strict;
use warnings;
use feature ':5.16';
```

good like this?

okurz:
> Small typo in commit message and in pull request description: `strictures` shold be `structures`.

uh, no. I meant strictures as from https://metacpan.org/pod/strictures but maybe that word only makes sense when using that pod. So rephrased.

Grinnz (approved):
LGTM. don't know why perltidy test is failing

kraih:
> LGTM. don't know why perltidy test is failing

New perltidy release. Just rebase the PR on main and it will be fine.

okurz:
Rebased

Grinnz (approved):

Grinnz:
Thank you for persisting to tidy this up!

jberger (approved):

PR #2181, opened 2024-05-17
Avoid using $_ when loading the commands from namespaces

### LINKED ISSUE #2175: Commands error when topic variable set to undef in compilation
* Mojolicious version: 9.30
* Perl version: 5.38
* Operating system: FreeBSD 14.0

Loop at [line 69 of Mojolicious::Commands](https://metacpan.org/dist/Mojolicious/source/lib/Mojolicious/Commands.pm#L69) encounters an undefined value if the topic variable `$_` is set to undefined in the compilation phase of that command. Example (on a freshly generated app):

```
package MyApp::CLI::test;

use v5.38;
use parent 'Mojolicious::Command';

use constant description => 'blah blah';

sub run { ... }

$_ = undef;

```
Note: In the real project case I'm not setting that value explicitly, I'm using `Beam::Wire` at compilation phase which most likely does that one way or another because the observed behavior is the same.

### Expected behavior
While polluting $_ is obviously erroneous, I believe Commands system should be resistant. 

### Actual behavior
I get these errors:
```
Use of uninitialized value $_ in substr at /root/perl5/perlbrew/perls/perl-5.38.0/lib/site_perl/5.38.0/Mojolicious/Commands.pm line 69.
substr outside of string at /root/perl5/perlbrew/perls/perl-5.38.0/lib/site_perl/5.38.0/Mojolicious/Commands.pm line 69.
Use of uninitialized value in hash element at /root/perl5/perlbrew/perls/perl-5.38.0/lib/site_perl/5.38.0/Mojolicious/Commands.pm line 69.
Can't call method "new" on an undefined value at /root/perl5/perlbrew/perls/perl-5.38.0/lib/site_perl/5.38.0/Mojolicious/Commands.pm line 69.
```

Rewriting the Mojolicious::Commands loop to this form fixes the issue for me:
```
	for my $pkg (find_modules($ns), find_packages($ns)) {
		next unless _command($pkg);
		$all{substr $pkg, length "${ns}::"} //= $pkg->new->description
	}
```

bbrtj:
### Summary
Solution proposed in #2175 

### Motivation
Polluting `$_` can happen deep down in dependencies tree of the loaded command.

### References
resolves #2175


bbrtj:
Bump, why isn't this merged? Seems like a low-hanging fruit.

kraih:
Nobody will even look at it before tests passed,

bbrtj:
I'm confused. Am I required to take any action here? Did the tests fail? They look like they never ran at all.

kraih:
No idea what went wrong i'm afraid, and nothing we can do about it.

bbrtj:
I rebased to latest main. Seems like workflows await approval now. Maybe that'll fix them.

bbrtj:
Most tests passed now. Windows test seems to be failing in other PRs too, regardless of their content.

PR #2192, opened 2024-08-09
Mojo::JSON: encode core boolean values as json bools

### LINKED ISSUE #2191: Mojo::JSON, without Cpanel::JSON::XS, doesn't encode booleans correctly in Perl v5.40
* Mojolicious version: 9.37
* Perl version: v5.40.0
* Operating system: Debian Linux 12.6

When `use v5.40;` exists, Perl's builtin core booleans are used instead of Mojo::JSON's. And when Cpanel::JSON::XS is not installed, Mojo::JSON is incapable of dealing with the core booleans in the expected way, leading to mistakes in the output of encode_json.

This simple program produces unexpected output:

```perl
#!/usr/bin/env perl

use v5.40;
use FindBin '$RealBin';
use lib "$RealBin/local/lib/perl5";

use Mojo::JSON qw/ true false encode_json /;

say encode_json({a => true, b => false});
```

### Expected behavior

Output should be:

```
{"a":true,"b":false}
```

### Actual behavior

Output is:

```
{"a":1,"b":""}
```

haarg:
### Summary
Encode perl's built in boolean values on perl 5.36 and newer

### Motivation
Newer perl versions can track if a value is a boolean rather than just a string or number. Since we can identify them as being booleans, they can be encoded in JSON as booleans. This also matches the behavior of newer versions of JSON::PP and Cpanel::JSON::XS

### References
Fixes #2191

### DIFF
--- a/lib/Mojo/JSON.pm
+++ b/lib/Mojo/JSON.pm
@@ -12,6 +12,12 @@ use constant JSON_XS => $ENV{MOJO_NO_JSON_XS}
   ? 0
   : !!eval { require Cpanel::JSON::XS; Cpanel::JSON::XS->VERSION('4.09'); 1 };
 
+use constant CORE_BOOLS => defined &builtin::is_bool;
+
+BEGIN {
+  warnings->unimport('experimental::builtin') if CORE_BOOLS;
+}
+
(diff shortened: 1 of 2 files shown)


kraih (approved):

christopherraa (approved):
Looks good to me.

PR #2196, opened 2024-08-28
Add support for partitioned cookies

### LINKED ISSUE #2179: Feature request: Support session cookies with "Partitioned"
Third-party cookies will "soon" be phased out.
As seen in Firefox console for example:

Cookie "mojolicious" will soon be rejected because it is foreign and does not have the "Partitioned" attribute.

I would like Mojolicious to support "Partitioned" session cookies, so we can keep using session cookies with CORS.

Thank you for considering this.

rawleyfowler:
### Summary
Firefox is soon going to ignore third-party aka "foreign" aka `SameSite: None`

### Motivation
These changes allow Mojolicious to handle the new "Partitioned" attribute, as well as adding the ability to set `Partitioned` on `Mojolicious::Sessions` cookies.

### References

https://www.ietf.org/archive/id/draft-cutler-httpbis-partitioned-cookies-00.html

https://github.com/privacycg/CHIPS

fixes #2179

### DIFF
--- a/lib/Mojo/Cookie/Response.pm
+++ b/lib/Mojo/Cookie/Response.pm
@@ -4,9 +4,9 @@ use Mojo::Base 'Mojo::Cookie';
 use Mojo::Date;
 use Mojo::Util qw(quote split_cookie_header);
 
-has [qw(domain expires host_only httponly max_age path samesite secure)];
+has [qw(domain expires host_only httponly max_age partitioned path samesite secure)];
 
-my %ATTRS = map { $_ => 1 } qw(domain expires httponly max-age path samesite secure);
+my %ATTRS = map { $_ => 1 } qw(domain expires httponly max-age partitioned path samesite secure);
 
 sub parse {
   my ($self, $str) = @_;
@@ -21,7 +21,7 @@ sub parse {
       next unless $ATTRS{my $attr = lc $name};
       $value =~ s/^\.// if $attr eq 'domain' && defined $value;
       $value = Mojo::Date->new($value // '')->epoch if $attr eq 'expires';
-      $value = 1                                    if $attr eq 'secure' || $attr eq 'httponly';
+      $value = 1 if $attr eq 'secure' || $attr eq 'httponly' || $attr eq 'partitioned';
       $cookies[-1]{$attr eq 'max-age' ? 'max_age' : $attr} = $value;
     }
   }
@@ -46,6 +46,9 @@ sub to_string {
 
   # "path"
   if (my $path = $self->path) { $cookie .= "; path=$path" }
+
+  # "Partitioned"
+  $cookie .= "; Partitioned" if $self->partitioned;
 
   # "secure"
   $cookie .= "; secure" if $self->secure;
@@ -122,6 +125,16 @@ HttpOnly flag, which can prevent client-side scripts from accessing this cookie.
   $cookie     = $cookie->max_age(60);
 
 Max age for cookie.
+
+=head2 partitioned
+
+  my $partitioned = $cookie->partitioned;
+  $cookie         = $cookie->partitioned(1);
+
+Partitioned flag, this is to be used in accordance to the CHIPS amendment to RFC 6265.
+Note that this attribute is B<EXPERIMENTAL> because even though most commonly used browsers support the
+feature, there is no specification yet besides L<this
+draft|https://www.ietf.org/archive/id/draft-cutler-httpbis-partitioned-cookies-00.html>.
 
 =head2 path
 
--- a/lib/Mojolicious/Sessions.pm
+++ b/lib/Mojolicious/Sessions.pm
@@ -4,7 +4,7 @@ use Mojo::Base -base;
 use Mojo::JSON;
 use Mojo::Util qw(b64_decode b64_encode);
 
-has [qw(cookie_domain encrypted secure)];
+has [qw(cookie_domain encrypted partitioned secure)];
 has cookie_name        => 'mojolicious';
 has cookie_path        => '/';
 has default_expiration => 3600;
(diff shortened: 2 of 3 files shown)


kraih (dismissed):
Looks like some values are not ordered alphabetically and documentation is missing.

lib/Mojo/Cookie/Response.pm
```diff
@@ -123,6 +126,14 @@ HttpOnly flag, which can prevent client-side scripts from accessing this cookie.
 
 Max age for cookie.
 
+=head2 partitioned
+
+  my $partitioned = $cookie->partitioned;
+  $cookie         = $cookie->parititoned(1);
+
+Partitioned flag value on all session cookies, this is to be used in accordance to the L<CHIPS ammendment
+to RFC 6265|https://www.ietf.org/archive/id/draft-cutler-httpbis-partitioned-cookies-00.html>.
```
kraih, line 135: I don't quite understand why session cookies are mentioned here specifically.

lib/Mojolicious/Sessions.pm
```diff
@@ -9,6 +9,7 @@ has cookie_name        => 'mojolicious';
 has cookie_path        => '/';
 has default_expiration => 3600;
 has deserialize        => sub { \&_deserialize };
+has partitioned        => 0;
```
kraih, line 12: How is this different from the secure flag?

lib/Mojo/Cookie/Response.pm
rawleyfowler: I had the documentation backwards my bad. This should be on the sessions pod not the Cookie::Response pod.

lib/Mojolicious/Sessions.pm
rawleyfowler: Partitioned is used to tell browsers that cookies should be stored in a separate cookie-jar, hidden from other TLDs. You can see the RFC amendment explanation here: https://github.com/privacycg/CHIPS

rawleyfowler:
We could also mark this as experimental, though it is implemented in most browsers.

lib/Mojolicious/Sessions.pm
kraih: Not what i mean. Why does it have to be initialised with 0, unlike other flags?

rawleyfowler: You're correct, I had it as `1` to begin with, but then I assumed we'd probably not want to default it on. I'll change it to be in the other declaration, good catch.

rawleyfowler:
Plack recently merged this into `Plack::Middleware::Session`: https://github.com/plack/Plack-Middleware-Session/pull/51

t/mojo/cookie.t
```diff
@@ -216,6 +217,18 @@ subtest 'Parse response cookie (RFC 6265)' => sub {
   is $cookies->[1],          undef,         'no more cookies';
 };
 
+subtest 'Partitioned cookie (RFC 6265 CHIPS)' => sub {
+  my $cookies
+    = Mojo::Cookie::Response->parse(
+    'foo="bar"; Domain=example.com; Partitioned; Path=/test; Max-Age=60; Expires=Thu, 07 Aug 2008 07:07:59 GMT; Secure;'
+    );
+  is $cookies->[0]->partitioned, 1, 'partitioned set?';
```
kraih, line 225: The question mark is out of place.

kraih:
Yes, since this is only a draft spec we have to mark it experimental.

lib/Mojolicious/Sessions.pm
```diff
@@ -150,6 +151,18 @@ A callback used to deserialize sessions, defaults to L<Mojo::JSON/"j">.
...
 
+=head2 partitioned
+
+  my $bool  = $sessions->partitioned;
+  $sessions = $sessions->partitioned($bool);
+
+Set the partitioned flag on all session cookies, this is used in accordance to the L<CHIPS ammendment to
+RFC 6265|https://www.ietf.org/archive/id/draft-cutler-httpbis-partitioned-cookies-00.html>.
+Note that this attribute is B<EXPERIMENTAL> because even though most commonly used browsers support the
+feature, the specification is still a draft.
```
kraih, line 162: Please use the same format as other features do for marking them as experimental.

kraih (dismissed):

kraih (dismissed):

t/mojo/cookie.t
```diff
@@ -216,6 +217,18 @@ subtest 'Parse response cookie (RFC 6265)' => sub {
   is $cookies->[1],          undef,         'no more cookies';
 };
 
+subtest 'Partitioned cookie (RFC 6265 CHIPS)' => sub {
+  my $cookies
+    = Mojo::Cookie::Response->parse(
+    'foo="bar"; Domain=example.com; Partitioned; Path=/test; Max-Age=60; Expires=Thu, 07 Aug 2008 07:07:59 GMT; Secure;'
```
kraih: This can probably be formatted nicer as well.

kraih:
Once the tests are formatted nicer i'd like to get this merged.

rawleyfowler:
I'll clean this up later today, thanks.

rawleyfowler:
I did notice perl-tidy broken when I cleaned this up. When it was formatted originally it was using the perl-tidy in this repo... not sure if there is a better way.

rawleyfowler:
@kraih Do you have a suggestion for how to format that test? Obviously automation has to pass, but PerlTidy seems to make it ugly.

kraih:
Just split up the strings and concat them, usually works very well. You probably shoudd rebase for tests to pass.

rawleyfowler:
@kraih Just fixed the formatting, let me know if you need anything else :)

kraih (approved):

rawleyfowler:
@kraih Any chance we can get this in soon? We're targeting a release in a couple weeks and we'd like to get off our fork if possible.

kraih:
Everybody on vacation i suppose. Thanks, merged.

rawleyfowler:
Appreciate it!

PR #2200, opened 2024-09-27
Secure by default sessions

stigtsp:
### Summary

To make session secrets secure by default, generate and persist a 256 bit session secret:

* Add `urandom_bytes` and `urandom_urlsafe` to `Mojo::Util` for generating secure random bits from either `Crypt::Random` or `/dev/urandom`

* Don't use the hard coded moniker as the default secret

* Generate and store a strong secret if not exists in `$ENV{MOJO_HOME}/mojo.secrets`, overridable with `$ENV{MOJO_SECRETS_FILE}` when app->secrets is called

* Only load secrets from `mojo.secrets` that are over 22 chars

* Use `urandom_urlsafe` when generating CSRF tokens

* Use `urandom_urlsafe` in `mojo generate app`

* Add `mojo generate secret`

* Tests:
   - Add misc tests for generating and loading mojo.secrets in `t/mojolicious/secret/` and for `mojo generate secret`.
   - Add a default secret in `t/mojolicious/mojo.secrets` so other session checks work

* Install `Crypt::URandom` in GH Windows workflow so `urandom_bytes` works on that platform

### Also consider

- Disallowing insecure secrets from being passed to `app->secrets()`
- Removing padding of HMAC message to 1025 bytes introduced by #1791

### Motivation

Making HMAC protected sessions secure by default after discussing with @jberger at PTS24. 

It has been demonstrated by [Baking Mojolicious cookies](https://www.synacktiv.com/publications/baking-mojolicious-cookies) and the recent discussion in #1791 that this is not the case.

### DIFF
--- a/lib/Mojo/Util.pm
+++ b/lib/Mojo/Util.pm
@@ -13,7 +13,7 @@ use IO::Compress::Gzip;
 use IO::Poll qw(POLLIN POLLPRI);
 use IO::Uncompress::Gunzip;
 use List::Util         qw(min);
-use MIME::Base64       qw(decode_base64 encode_base64);
+use MIME::Base64       qw(decode_base64 encode_base64 encode_base64url);
 use Mojo::BaseUtil     qw(class_to_path monkey_patch);
 use Pod::Usage         qw(pod2usage);
 use Socket             qw(inet_pton AF_INET6 AF_INET);
@@ -23,6 +23,8 @@ use Unicode::Normalize ();
 
 # Check for monotonic clock support
 use constant MONOTONIC => !!eval { Time::HiRes::clock_gettime(Time::HiRes::CLOCK_MONOTONIC()) };
+
+use constant CRYPT_URANDOM => !!eval { require Crypt::URandom };
 
 # Punycode bootstring parameters
 use constant {
@@ -72,7 +74,7 @@ our @EXPORT_OK = (
   qw(extract_usage getopt gunzip gzip header_params hmac_sha1_sum html_attr_unescape html_unescape humanize_bytes),
   qw(md5_bytes md5_sum monkey_patch network_contains punycode_decode punycode_encode quote scope_guard secure_compare),
   qw(sha1_bytes sha1_sum slugify split_cookie_header split_header steady_time tablify term_escape trim unindent),
-  qw(unquote url_escape url_unescape xml_escape xor_encode)
+  qw(unquote urandom_bytes urandom_urlsafe url_escape url_unescape xml_escape xor_encode)
 );
 
 # Aliases
@@ -377,6 +379,25 @@ sub unquote {
   $str                    =~ s/\\\\/\\/g;
   $str                    =~ s/\\"/"/g;
   return $str;
+}
+
+sub urandom_bytes {
+  my $num = shift || 32;
+
+  return Crypt::URandom::urandom($num) if CRYPT_URANDOM;
+
+  croak 'Cannot find /dev/urandom, install Crypt::URandom from CPAN' unless -e '/dev/urandom';
+
+  open(my $urandom, '<', '/dev/urandom')     or croak "Cannot open /dev/urandom: $!";
+  sysread($urandom, my $bytes, $num) == $num or croak "sysread() from /dev/urandom didn't return $num bytes";
+  close($urandom);
+
+  return $bytes;
+}
+
+sub urandom_urlsafe {
+  my $num = shift;
+  return encode_base64url(urandom_bytes($num));
 }
 
 sub url_escape {
@@ -957,6 +978,22 @@ Unindent multi-line string.
   my $str = unquote $quoted;
 
 Unquote string.
+
+=head2 urandom_bytes
+
+  my $bytes = urandom_bytes;
+  my $bytes = urandom_bytes 32;
+
+Returns strong random bytes. Uses L<Crypt::URandom> if it is installed, with fallback to /dev/urandom. The default
+number of random bytes returned is 32.
+
+=head2 urandom_urlsafe
+
+  my $token = urandom_urlsafe;
+  my $token = urandom_urlsafe 32;
+
+Generates a base64url encoded string of random bytes suitable for session tokens and similar. The default number of random
+bytes encoded is 32.
 
 =head2 url_escape
 
--- a/lib/Mojolicious.pm
+++ b/lib/Mojolicious.pm
@@ -9,7 +9,8 @@ use Mojo::Home;
 use Mojo::Loader;
 use Mojo::Log;
 use Mojo::Server;
-use Mojo::Util;
+use Mojo::Util qw(urandom_urlsafe);
+use Mojo::File qw(path);
 use Mojo::UserAgent;
 use Mojolicious::Commands;
 use Mojolicious::Controller;
@@ -41,14 +42,31 @@ has plugins            => sub { Mojolicious::Plugins->new };
 has preload_namespaces => sub { [] };
 has renderer           => sub { Mojolicious::Renderer->new };
 has routes             => sub { Mojolicious::Routes->new };
+has secrets_file       => sub { $ENV{MOJO_SECRETS_FILE} || shift->home->rel_file('mojo.secrets') };
 has secrets            => sub {
   my $self = shift;
+  my $file = $self->secrets_file;
 
-  # Warn developers about insecure default
-  $self->log->trace('Your secret passphrase needs to be changed (see FAQ for more)');
+  if (-f $file) {
 
-  # Default to moniker
-  return [$self->moniker];
+    # Read secrets and filter out those who are less than 22 characters long
+    # (~128 bits), as they are not likely to be sufficiently strong.
+    my @secrets = grep { length $_ >= 22 } split /\n/, path($file)->slurp;
+
+    die qq{"Your secrets_file "$file" does not contain any acceptable secret (of 22 chars or more)} unless @secrets;
+
+    return [@secrets];
+  }
+
+  # If no secrets file exists, generate one and attempt to write it back to
+  # secrets_file, taking care that the file is only readable by the current
+  # user.
+  my $secret = urandom_urlsafe;
+  path($file)->touch->chmod(0600)->spew($secret);
+
+  $self->log->trace(qq{Your secret passphrase has been set to strong random value and stored in "$file"});
+
+  return [$secret];
 };
 has sessions  => sub { Mojolicious::Sessions->new };
 has static    => sub { Mojolicious::Static->new };
@@ -496,11 +514,13 @@ endpoints for your application.
   my $secrets = $app->secrets;
   $app        = $app->secrets([$bytes]);
 
-Secret passphrases used for signed cookies and the like, defaults to the L</"moniker"> of this application, which is
-not very secure, so you should change it!!! As long as you are using the insecure default there will be debug messages
-in the log file reminding you to change your passphrase. Only the first passphrase is used to create new signatures,
-but all of them for verification. So you can increase security without invalidating all your existing signed cookies by
-rotating passphrases, just add new ones to the front and remove old ones from the back.
+Secret passphrases used for signed cookies and the like, defaults to 256 bits of data from your systems secure
+random number generator and is stored in the file mojo.secrets in your MOJO_HOME directory. You can override
+the location of this file by setting MOJO_SECRETS_FILE in your environment.
+
+Only the first passphrase is used to create new signatures, but all of them for verification. So you can
+increase security without invalidating all your existing signed cookies by rotating passphrases, just add new
+ones to the front and remove old ones from the back.
 
   # Rotate passphrases
   $app->secrets(['new_passw0rd', 'old_passw0rd', 'very_old_passw0rd']);
--- a/lib/Mojolicious/Command/Author/generate/secret.pm
+++ b/lib/Mojolicious/Command/Author/generate/secret.pm
@@ -1,0 +1,82 @@
+package Mojolicious::Command::Author::generate::secret;
+use Mojo::Base 'Mojolicious::Command';
+use Mojo::File qw(path);
+use Mojo::Util qw(urandom_urlsafe);
+
+has description => 'Generate secret';
+has usage       => sub { shift->extract_usage };
+
+sub run {
+  my ($self, $secret_file) = (shift, shift);
+
+  $secret_file //= $self->app->secrets_file;
+
+  my $token = urandom_urlsafe();
+
+  print "Writing secret to $secret_file\n";
+
+  path($secret_file)->touch->chmod(0600)->spew($token);
+}
+
+1;
+
+=encoding utf8
+
+=head1 NAME
+
+Mojolicious::Command::Author::generate::secret - Secret generator command
+
+=head1 SYNOPSIS
+
+  Usage: APPLICATION generate secret [PATH]
+
+    mojo generate secret
+    mojo generate secret /path/to/secret
+
+  Options:
+    -h, --help   Show this summary of available options
+
+=head1 DESCRIPTION
+
+L<Mojolicious::Command::Author::generate::secret> generates a secret token for protecting session cookies
+
+This is a core command, that means it is always enabled and its code a good example for learning to build new commands,
+you're welcome to fork it.
+
+See L<Mojolicious::Commands/"COMMANDS"> for a list of commands that are available by default.
+
+=head1 ATTRIBUTES
+
+L<Mojolicious::Command::Author::generate::secret> inherits all attributes from L<Mojolicious::Command> and implements
+the following new ones.
+
+=head2 description
+
+  my $description = $app->description;
+  $app            = $app->description('Foo');
+
+Short description of this command, used for the command list.
+
+=head2 usage
+
+  my $usage = $app->usage;
+  $app      = $app->usage('Foo');
+
+Usage information for this command, used for the help screen.
+
+=head1 METHODS
+
+L<Mojolicious::Command::Author::generate::secret> inherits all methods from L<Mojolicious::Command> and implements
+the following new ones.
+
+=head2 run
+
+  $app->run(@ARGV);
+
+Run this command.
+
+=head1 SEE ALSO
+
+L<Mojolicious>, L<Mojolicious::Guides>, L<https://mojolicious.org>.
+
+=cut
--- a/t/mojo/util.t
+++ b/t/mojo/util.t
@@ -12,7 +12,7 @@ use Mojo::Util qw(b64_decode b64_encode camelize class_to_file class_to_path dec
   qw(extract_usage getopt gunzip gzip header_params hmac_sha1_sum html_unescape html_attr_unescape humanize_bytes),
   qw(md5_bytes md5_sum monkey_patch network_contains punycode_decode punycode_encode quote scope_guard secure_compare),
   qw(sha1_bytes sha1_sum slugify split_cookie_header split_header steady_time tablify term_escape trim unindent),
-  qw(unquote url_escape url_unescape xml_escape xor_encode);
+  qw(unquote urandom_bytes urandom_urlsafe url_escape url_unescape xml_escape xor_encode);
 
 subtest 'camelize' => sub {
   is camelize('foo_bar_baz'), 'FooBarBaz', 'right camelized result';
@@ -660,5 +660,19 @@ subtest 'Hide DATA usage from error messages' => sub {
   eval { die 'whatever' };
   unlike $@, qr/DATA/, 'DATA has been hidden';
 };
+
+subtest 'urandom' => sub {
+  isnt urandom_bytes,           urandom_bytes, "two urandom_bytes invocations are not the same";
+  is length(urandom_bytes),     32,            "urandom_bytes returns 32 bytes by default";
+  is length(urandom_bytes(16)), 16,            "urandom_bytes(16) returns 16 bytes";
+
+  isnt urandom_urlsafe, urandom_urlsafe, "two urandom_urlsafe invocations are not the same";
+  like urandom_urlsafe, qr/^[-A-Za-z0-9_]{43}$/, "urandom_urlsafe returns 43 chars of urlsafe encoded base64";
+  like urandom_urlsafe(128), qr/^[-A-Za-z0-9_]{171}$/,
+    "urandom_urlsafe(128) returns 171 chars of urlsafe encoded base64";
+  like urandom_urlsafe(2048), qr/^[-A-Za-z0-9_]{2731}$/,
+    "urandom_urlsafe(2048) returns 2731 chars of urlsafe encoded base64";
+};
+
 
 done_testing();
--- a/t/mojolicious/secret/lite-create.t
+++ b/t/mojolicious/secret/lite-create.t
@@ -1,0 +1,16 @@
+use Mojo::Base -strict;
+
+use Mojo::File qw(tempdir path);
+use Test::Mojo;
+use Test::More;
+use Mojolicious::Lite;
+
+
+my $tmpdir = tempdir;
+my $file   = $tmpdir->child("mojo.secrets");
+$ENV{MOJO_SECRETS_FILE} = $file;
+
+like app->secrets->[0], qr/^[-A-Za-z0-9_]{43}$/, 'secret was generated, and matches expected urandom_urlsafe format';
+is app->secrets->[0], $file->slurp, 'secret stored at $ENV{MOJO_SECRETS_FILE} is the same as app->secrets->[0]';
+
+done_testing();
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -24,6 +24,6 @@ jobs:
       - name: Install Dependencies
         run: |
           cpanm --installdeps .
-          cpanm -n TAP::Formatter::GitHubActions
+          cpanm -n TAP::Formatter::GitHubActions Crypt::URandom
       - name: Run Tests
         run: prove --merge --formatter TAP::Formatter::GitHubActions -l t t/mojo t/mojolicious
--- a/lib/Mojolicious/Command/Author/generate/app.pm
+++ b/lib/Mojolicious/Command/Author/generate/app.pm
@@ -196,7 +196,7 @@ done_testing();
 </p>
 
 @@ config
-% use Mojo::Util qw(sha1_sum steady_time);
+% use Mojo::Util qw(urandom_urlsafe);
 ---
 secrets:
-  - <%= sha1_sum $$ . steady_time . rand  %>
+  - <%= urandom_urlsafe %>
--- a/lib/Mojolicious/Plugin/DefaultHelpers.pm
+++ b/lib/Mojolicious/Plugin/DefaultHelpers.pm
@@ -8,7 +8,7 @@ use Mojo::Collection;
 use Mojo::Exception;
 use Mojo::IOLoop;
 use Mojo::Promise;
-use Mojo::Util   qw(dumper hmac_sha1_sum steady_time);
+use Mojo::Util   qw(dumper urandom_urlsafe);
 use Time::HiRes  qw(gettimeofday tv_interval);
 use Scalar::Util qw(blessed weaken);
 
@@ -95,7 +95,7 @@ sub _convert_to_exception {
   return (blessed $e && $e->isa('Mojo::Exception')) ? $e : Mojo::Exception->new($e);
 }
 
-sub _csrf_token { $_[0]->session->{csrf_token} ||= hmac_sha1_sum($$ . steady_time . rand, $_[0]->app->secrets->[0]) }
+sub _csrf_token { $_[0]->session->{csrf_token} ||= urandom_urlsafe; }
 
 sub _current_route {
   return '' unless my $route = shift->match->endpoint;
--- a/t/mojolicious/app.t
+++ b/t/mojolicious/app.t
@@ -90,9 +90,9 @@ is_deeply $t->app->commands->namespaces,
   ['Mojolicious::Command::Author', 'Mojolicious::Command', 'MojoliciousTest::Command'], 'right namespaces';
 is $t->app,                                   $t->app->commands->app,                         'applications are equal';
 is $t->app->static->file('hello.txt')->slurp, "Hello Mojo from a development static file!\n", 'right content';
-is $t->app->static->file('does_not_exist.html'), undef,              'no file';
-is $t->app->moniker,                             'mojolicious_test', 'right moniker';
-is $t->app->secrets->[0],                        $t->app->moniker,   'secret defaults to moniker';
+is $t->app->static->file('does_not_exist.html'), undef,                                       'no file';
+is $t->app->moniker,                             'mojolicious_test',                          'right moniker';
+is $t->app->secrets->[0], 'NeverGonnaGiveYouUpNeverGonnaLetYouDown', 'secret defaults to content of mojo.secrets';
 is $t->app->renderer->template_handler({template => 'foo/bar/index', format => 'html'}), 'epl',   'right handler';
 is $t->app->build_controller->req->url,                                                  '',      'no URL';
 is $t->app->build_controller->render_to_string('does_not_exist'),                        undef,   'no result';
--- a/t/mojolicious/commands.t
+++ b/t/mojolicious/commands.t
@@ -158,6 +158,13 @@ $buffer = '';
 {
   open my $handle, '>', \$buffer;
   local *STDOUT = $handle;
+  $commands->run('generate', 'secret', '--help');
+}
+like $buffer, qr/Usage: APPLICATION generate secret \[PATH\]/, 'right output';
+$buffer = '';
+{
+  open my $handle, '>', \$buffer;
+  local *STDOUT = $handle;
   local $ENV{HARNESS_ACTIVE} = 0;
   $commands->run('help');
 }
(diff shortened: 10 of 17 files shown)


lib/Mojo/Util.pm
```diff
@@ -88,6 +88,33 @@ monkey_patch(__PACKAGE__, 'sha1_sum',      \&sha1_hex);
 monkey_patch(__PACKAGE__, 'steady_time',
   MONOTONIC ? sub () { Time::HiRes::clock_gettime(Time::HiRes::CLOCK_MONOTONIC()) } : \&Time::HiRes::time);
 
+# Building random bytes function
+monkey_patch(__PACKAGE__, 'urandom_bytes', $^O eq 'MSWin32' ? _build_urandom_bytes_win32() : _build_urandom_bytes());
```
Grinnz, line 92: This seems over-optimized, this can just be an if condition within the function.

lib/Mojolicious.pm
```diff
@@ -41,14 +42,29 @@ has plugins            => sub { Mojolicious::Plugins->new };
...
+  }
+
+  # If no secret exists, generate one and attempt to write it back to
+  # secret_file, taking care that the file is only readable by the current user.
+  my $secret = urandom_token;
+  path($file)->touch->chmod(0600)->spew($secret);
+
+  $self->log->trace("Your secret passphrase has been set to strong random value and stored in $file");
+
+  return [$secret];
```
Grinnz, line 67: While I think this is generally a good and necessary step, this will invalidate session cookies in currently insecure apps so that will need to at minimum be noted in the changelog (which is not the responsibility of this PR).

lib/Mojolicious.pm
```diff
@@ -496,8 +512,7 @@ endpoints for your application.
   my $secrets = $app->secrets;
   $app        = $app->secrets([$bytes]);
 
-Secret passphrases used for signed cookies and the like, defaults to the L</"moniker"> of this application, which is
-not very secure, so you should change it!!! As long as you are using the insecure default there will be debug messages
+Secret passphrases used for signed cookies and the like, defaults to 44 random characters in the A..Za..z0..9 range. As long as you are using the insecure default there will be debug messages
```
Grinnz, line 515: With this change there would no longer be an "insecure default" nor debug messages about such.

lib/Mojolicious/Command/Author/generate/secret.pm
```diff
@@ -0,0 +1,85 @@
...
+use Mojo::File qw(path);
+use Mojo::Util qw(urandom_token);
+
+has description => 'Generate secret';
+has usage       => sub { shift->extract_usage };
+
+sub run {
+  my ($self, $secret_file) = (shift, shift);
+
+  $secret_file //= path($self)->sibling("mojo.secret");
```
Grinnz, line 12: This seems wrong? path($self) doesn't make sense, nor should it default to writing files into the Perl lib directory.

t/mojo/util.t
```diff
@@ -661,4 +661,31 @@ subtest 'Hide DATA usage from error messages' => sub {
...
+    my %values;
+    foreach my $char (split "", urandom_token($len, \@alphabet)) {
+      $values{$char}++;
+    }
+
+    my $avg = $len / @alphabet;
+    foreach my $char (@alphabet) {
+      my $cnt  = $values{$char} || 0;
+      my $diff = abs($avg - $cnt);
+      cmp_ok($diff, "<", $tolerance * $avg, "$char is within tolerance");
```
Grinnz, line 685: I don't know how reliable this test is, of course it is good to test the randomness but I am concerned about randomness tests causing potential false negatives. If appropriate it could be skipped unless an environment variable is present.

lib/Mojo/Util.pm
Grinnz: I'd additionally recommend a constant something like this so such a condition can be inlined by the parser and it can be easily managed later:

```
use constant WIN32_URANDOM => $^O eq 'MSWin32';
```

lib/Mojo/Util.pm
```diff
@@ -88,6 +88,33 @@ monkey_patch(__PACKAGE__, 'sha1_sum',      \&sha1_hex);
...
+    return $bytes;
+  }
+}
+
+sub _build_urandom_bytes {
+  open(my $urandom, '<', '/dev/urandom') or croak "Cannot open /dev/urandom: $!";
+
+  return sub {
+    my $num = shift || 64;
+    sysread($urandom, my $bytes, $num);
```
guest20, line 113: Is it worth checking the return to find out of we got all the bytes we asked for?

Grinnz: This is covered by tests, I'm not sure how necessary it is to assert at runtime but it would be a lightweight check.

lib/Mojo/Util.pm
```diff
@@ -88,6 +88,33 @@ monkey_patch(__PACKAGE__, 'sha1_sum',      \&sha1_hex);
...
+  return sub {
+    my $num   = shift || 64;
+    my $bytes = chr(0) x $num;
+    $RtlGenRandom->Call($bytes, $num) or croak "SystemFunction036 call failed: $!";
+    return $bytes;
+  }
+}
+
+sub _build_urandom_bytes {
+  open(my $urandom, '<', '/dev/urandom') or croak "Cannot open /dev/urandom: $!";
```
Grinnz, line 109: I'm not sure if it's a good idea to leave a read handle open to this for the process life. But maybe it's fine. Would there be any issues with sharing this over forks or threads?

lib/Mojolicious.pm
kraih: This also does not support our rolling secret feature, which i believe increases security if used well.

kraih:
Make sure to squash your comits when you are ready.

lib/Mojo/Util.pm
```diff
@@ -379,6 +406,27 @@ sub unquote {
...
+  my (@bytes, @token);
+  while (@token < $length) {
+    @bytes = split "", urandom_bytes(64) unless @bytes;
+    my $num = ord(shift @bytes);
+    next if $num < $bias_lim;
+    push @token, $alphabet[$num % @alphabet];
+  }
+
+  return join "", @token;
+}
```
latk, line 428: I believe that this `urandom_token()` function is correct, however:

* it may be non-obvious to users how the requested length correlates to cryptographic strength ($`= \mathrm{length} \cdot \log_2\left(|\mathrm{alphabet}|\right)`$). As a consenquence, this very pull request also makes some odd (but harmless!) choices, e.g. 160-bit tokens for CSRF protection and ~262-bit tokens as the default app secret.
* the implementation is somewhat complicated
* it "wastes" random bits due to its bias limit, and also because it reads randomness in chunks of 512 bits

All of these could be side-stepped if choosing a design like [Python's `secrets.token_urlsafe(nbytes)` function](https://docs.python.org/3/library/secrets.html#secrets.token_urlsafe):

* no user defined alphabet. The function hardcodes a particular default alphabet.
* this simplifies the problem from "choose random letters" to "encode the bytes"
* the function's `nbytes` parameter directly relates to the cryptographic strength, which is generally more important than the output string length. Python doesn't document the default value, but CPython chooses 32 bytes, which is perfect for typical applications.

Thus, alternative suggestion for this function:

```perl
use MIME::Base64 qw(encode_base64url);

sub _build_urandom_bytes {
  ...
  return sub {
    my $num = shift || 32;  # read 256 bits by default
    ...
  }
}

sub urandom_urlsafe {
  my $nbytes = shift;
  return encode_base64url(urandom_bytes($nbytes));
}
```

Could also add a `urandom_hex()` function that encodes via `unpack "h*"`, e.g. to allow the `_csrf_helper()` to continue outputting a hex-formatted token. But I don't think that's really necessary.

stigtsp:
Thanks! I'm addressing all feedback, and will update the PR tomorrow or so.

lib/Mojo/Util.pm
guest20: This is vibes based, but the de-biasing code here feels "tricky", and I think it's because this code rejects `$num` from the low end instead of the high end of the range.

For example, with an alphabet of 100, you'd reject `$num < 55`... maybe it would be clearer to reject `$num > 200` instead? 

Rejecting the low end means you can never stop ignoring the mod and you have to think about more boundaries you can be off-by-one at while reading.

When rejecting from the high end one can visualise the code as pulling out the `$num`th element from `@alphabet x int(256 / @alphabet)`, that is extending `@alphabet` until you pass a length 255 and then trimming the end off both `$num` and off the elongated to match an even multiple. 

Obviously the math is the same, but having left-over (biasing) elements at the high end for both mod and for the rejection condition might make it more understandable?

t/mojo/util.t
kraih: Yea, this seems like it would be trouble in the future.

lib/Mojo/Util.pm
```diff
@@ -88,6 +88,33 @@ monkey_patch(__PACKAGE__, 'sha1_sum',      \&sha1_hex);
...
   MONOTONIC ? sub () { Time::HiRes::clock_gettime(Time::HiRes::CLOCK_MONOTONIC()) } : \&Time::HiRes::time);
 
+# Building random bytes function
+monkey_patch(__PACKAGE__, 'urandom_bytes', $^O eq 'MSWin32' ? _build_urandom_bytes_win32() : _build_urandom_bytes());
+
+sub _build_urandom_bytes_win32 {
+  require Win32::API;
+
+  my $RtlGenRandom = Win32::API->new('advapi32', 'INT SystemFunction036 (PVOID RandomBuffer, ULONG RandomBufferLength)')
+    or croak "SystemFunction036 import failed: $!";
```
kraih, line 98: Is this actually used in our Windows CI tests? This seems like code nobody on the team can maintain long term and we will end up removing it because of that the first it causes problems.

lib/Mojo/Util.pm
```diff
@@ -379,6 +406,27 @@ sub unquote {
   return $str;
 }
 
+sub urandom_token {
+  my $length   = shift || 44;
+  my @alphabet = @{shift || ["A" .. "Z", "a" .. "z", "0" .. "9"]};
+
+  unless (@alphabet >= 2 && @alphabet <= 256) {
+    croak "alphabet size must be between 2 and 256 elements";
+  }
```
kraih, line 415: Not covered by tests as far as i can see.

kraih: Is a custom alphabet actually valuable for us?

lib/Mojolicious/Command/Author/generate/secret.pm
```diff
@@ -0,0 +1,85 @@
+package Mojolicious::Command::Author::generate::secret;
```
kraih, line 1: Looks like the command does not have tests, like other generators do. Also wondering if we really do want a separate config file for secrets, if the command should have an option to automatically generate and rotate in a new secret.

lib/Mojolicious.pm
```diff
@@ -41,14 +42,29 @@ has plugins            => sub { Mojolicious::Plugins->new };
...
+has secrets_file       => sub { $ENV{MOJO_SECRETS_FILE} || shift->home->rel_file('mojo.secret') };
 has secrets            => sub {
   my $self = shift;
+  my $file = $self->secrets_file;
 
-  # Warn developers about insecure default
-  $self->log->trace('Your secret passphrase needs to be changed (see FAQ for more)');
+  if (-f $file) {
+    my @secrets = grep { length $_ >= 22 }    # 22 random chars in base62 alphabet should be at
+                                              # best 128 bits of entropy
```
kraih, line 52: Nitpick we like to keep comments aboce the code, not behind. And i think some other comments use unnecessary punctuation.

lib/Mojolicious.pm
```diff
@@ -41,14 +42,29 @@ has plugins            => sub { Mojolicious::Plugins->new };
...
+
+    return [@secrets];
+  }
+
+  # If no secret exists, generate one and attempt to write it back to
+  # secret_file, taking care that the file is only readable by the current user.
+  my $secret = urandom_token;
+  path($file)->touch->chmod(0600)->spew($secret);
+
+  $self->log->trace("Your secret passphrase has been set to strong random value and stored in $file");
```
kraih, line 65: Think we quote generated values in other log messages.

lib/Mojolicious.pm
```diff
@@ -41,14 +42,29 @@ has plugins            => sub { Mojolicious::Plugins->new };
...
-  return [$self->moniker];
+    die "Your secrets_file $file does not contain any acceptable secret (of 22 chars or more)" unless @secrets;
+
+    return [@secrets];
+  }
+
+  # If no secret exists, generate one and attempt to write it back to
+  # secret_file, taking care that the file is only readable by the current user.
+  my $secret = urandom_token;
+  path($file)->touch->chmod(0600)->spew($secret);
```
kraih, line 63: Probably reading and writing secrets files should be `Mojo::Util` functions too if we do want to go in that direction.

t/mojolicious/secret/lite-create.t
```diff
@@ -0,0 +1,16 @@
...
+use Test::More;
+use Mojolicious::Lite;
+
+$ENV{MOJO_SECRETS_FILE} = tempfile();
+
+like app->secrets->[0], qr/^[A-Za-z0-9]{44}$/, 'urandom_token generated secret';
+
+END {
+  unlink $ENV{MOJO_SECRETS_FILE};
+}
```
kraih, line 14: Isn't the point of using `tempfile` that it gets cleaned up automatically?

stigtsp: `tempfile` doesn't seem to clean up tempfiles, I've changed it to use `tempdir` instead.
  
Example with `tempfile` and the `unlink` removed:
  ```
  $ prove -Ilib  -r t/mojolicious/secret/lite-create.t -v
  t/mojolicious/secret/lite-create.t .. [2024-09-29 14:01:09.00902] [71360] [trace] Your secret passphrase has been set to strong random value and stored in "/tmp/Sp9NjlS0Es"
  [..]
  
  $ ls -l /tmp/Sp9NjlS0Es
  .rw------- sgo users 43 B Sun Sep 29 14:01:09 2024 /tmp/Sp9NjlS0Es
  ```

lib/Mojolicious/Command/Author/generate/secret.pm
stigtsp: Tests have been added for `mojo generate secret`

lib/Mojo/Util.pm
stigtsp: Tests are running `SystemFunction036` (RtlGenRandom) on Windows to get random bytes. It's used in the same way as for example [Crypt::URandom](https://metacpan.org/release/DDICK/Crypt-URandom-0.40/source/lib/Crypt/URandom.pm#L92-97). However, It is a legacy API, so we can consider using `ProcessPrng` instead.
   
Here are some discussions in rust and golang:
   - https://github.com/rust-random/rand/issues/111
   - https://github.com/golang/go/issues/53192

lib/Mojo/Util.pm
stigtsp: - `urandom_token` is replaced with `urandom_urlsafe`: As pointed out by @latk and @guest20, the mod reduction and rejection sampling is a bit complex and not neccecary for solving this PR.

   A secondary purpose of`urandom_token` was to provide a "password generator" with custom alphabet support similar to `Session::Token`, which would be useful not really needed for HMAC secrets.
  
   I've replaced it with `urandom_urlsafe` as suggested by @latk, where 256 bits are encoded as url safe base64.
   
   Tests checking mod bias has also been removed as they are no longer neccecary.
  
- `urandom_bytes`: Afaik it should be ok with `sysread` to keep an open file handle across forks, but it now opens `/dev/urandom` on each invocation rather than
   keeping the fh around, as suggested by @Grinnz. This should be less of a problem now that `urandom_urlsafe` encodes 32 bytes directly. 
   
   The Win32::API `SystemFunction036` object is kept using `state`.

lib/Mojo/Util.pm
stigtsp: - `urandom_bytes`: Afaik it should be ok with `sysread` to keep an open file handle across forks, but it now opens `/dev/urandom` on each invocation rather than
   keeping the fh around, as suggested by @Grinnz. This should be less of a problem now that `urandom_urlsafe` encodes 32 bytes directly. 
   
   The Win32::API `SystemFunction036` object is kept using `state`.

lib/Mojo/Util.pm
stigtsp: A simple check for this has been added:

```perl
    sysread($urandom, $bytes, $num) == $num or die "sysread() from /dev/urandom didn't return $num bytes";

```

lib/Mojolicious/Command/Author/generate/secret.pm
stigtsp: This is fixed

lib/Mojolicious.pm
stigtsp: This has been updated

lib/Mojo/Util.pm
stigtsp: This is fixed

lib/Mojolicious.pm
stigtsp: Fixed

lib/Mojolicious.pm
stigtsp: Fixed

stigtsp:
I've (hopefully) addressed all comments and pushed a single commit to this PR.

One problem that remains is that a `mojo.secrets` file is left over in the `t/mojo/` directory after tests are run, likely due to something calling `app->secrets` there. A quick fix for this is to copy the one from `t/mojolicious/` but it seems a bit messy. Any better ideas on how to fix this?

lib/Mojo/Util.pm
stigtsp: For password generation, maybe. But I've removed `urandom_token` (and custom alfabets) in favor of `urandom_urlsafe` as described in https://github.com/mojolicious/mojo/pull/2200#discussion_r1780049397

lib/Mojo/Util.pm
guest20: @stigtsp You don't need de-biasing if your alphabet has 256 characters in it

lib/Mojo/Util.pm
```diff
@@ -540,7 +563,7 @@ Mojo::Util - Portable utility functions
 
 =head1 SYNOPSIS
 
-  use Mojo::Util qw(b64_encode url_escape url_unescape);
+  use Mojo::Util go(b64_encode url_escape url_unescape);
```
latk, line 567: Unintended change of the quote operator?

lib/Mojo/Util.pm
```diff
@@ -958,6 +981,21 @@ Unindent multi-line string.
...
+Returns random bytes. On Linux and UNIX-like systems /dev/urandom is used as a
+source of random data, on Windows RtlGenRandom/SystemFunction036 is used. The
+default number of bytes returned is 32 bytes.
+
+=head2 urandom_urlsafe
+
+  my $token = urandom_urlsafe;
+
+Generates a base64url encoded string of 32 random bytes suitable for session
+tokens and similar.
```
latk, line 997: The signatures/examples of these functions could show that *both* take an optional argument. Here, it looks like `urandom_bytes()` always needs an argument, and that `urandom_urlsafe()` could never receive an argument.

lib/Mojolicious/Command/Author/generate/secret.pm
```diff
@@ -0,0 +1,82 @@
...
+has usage       => sub { shift->extract_usage };
+
+sub run {
+  my ($self, $secret_file) = (shift, shift);
+
+  $secret_file //= $self->app->secrets_file;
+
+  my $token = urandom_urlsafe();
+
+  print "Writing @{[ length($token) ]} byte to $secret_file\n";
```
latk, line 16: What is the intent of this message? It's probably good to inform users that the file is being written, but showing the expected file size may be less informative?

lib/Mojolicious.pm
```diff
@@ -41,14 +42,31 @@ has plugins            => sub { Mojolicious::Plugins->new };
...
 
-  # Warn developers about insecure default
-  $self->log->trace('Your secret passphrase needs to be changed (see FAQ for more)');
+  if (-f $file) {
 
-  # Default to moniker
-  return [$self->moniker];
+    # Read secrets and filter out those who are less than 22 characters long
+    # (~128 bits), as they are not likely to be sufficiently strong.
+    my @secrets = grep { length $_ >= 22 } split /\n/, path($file)->slurp;
```
latk, line 54: Should there be a warning message if a candidate secret is skipped? Right now, they're discarded silently.

t/mojo/util.t
stigtsp: This test has been removed as it's no longer needed

lib/Mojo/Util.pm
stigtsp: Fixed

lib/Mojo/Util.pm
stigtsp: Fixed

lib/Mojolicious/Command/Author/generate/secret.pm
stigtsp: I agree, removed the number of bytes

lib/Mojolicious.pm
Grinnz: It is useful so that instances using the session key can be upgraded to use the new secret over even a short period of time without immediately invalidating everything.

Grinnz: Also, regular secret rotation is good practice even when not found to be compromised.

lib/Mojo/Util.pm
kraih: This part of the code will probably end up being a blocker, since i don't see anyone on the team agreeing to take responsibility for maintaining it in the future.

jkramarz: I'd not expect finding such code in web framework. In the same time, AFAIK the alternatives are:
- using some small old library with single maintainer and with no recent releases, that does exactly this call, but conveniently wrapped, 
- overkill of making Net-SSLeay a dependency to get it through openssl binding.

kraih: `Net::SSLeay` is already a soft dependency, maybe there's a clean way to make this feature optional? A hard dependency is pretty much out of the question. It's unfortunate Perl doesn't ship with some kind of OpenSSL binding, for mojo.js we've chosen a different solution and [just encrypt all sessions](https://github.com/mojolicious/mojo.js/blob/main/src/session.ts).

guest20: `Crypt::URandom` has a pile of different window support calls. Adding that as an optional dependency for windows enjoyers might unblock this.

stigtsp: `Crypt::URandom` appears to [sets up some context](https://metacpan.org/release/DDICK/Crypt-URandom-0.40/source/lib/Crypt/URandom.pm#L63-88) for `CryptGenRandom` for Win2k, and uses `RtlGenRandom` for later Windows versions.

According to a discussion in golang crypto/rand, it seems fine to use `RtlGenRandom` (SystemFunction036) https://github.com/golang/go/issues/33542#issuecomment-715642267

stigtsp: What about something like this?

```perl

use constant CRYPT_URANDOM => !!eval { require Crypt::URandom };

...

sub urandom_bytes {
  my $num   = shift || 32;

  return Crypt::URandom::urandom($num) if CRYPT_URANDOM;

  croak 'Cannot find /dev/urandom. On Windows? Try installing "Crypt::URandom" from CPAN' 
    unless -f '/dev/urandom';

  open(my $urandom, '<', '/dev/urandom')  or croak "Cannot open /dev/urandom: $!";
  sysread($urandom, $bytes, $num) == $num or die "sysread() from /dev/urandom didn't return $num bytes";
  close($urandom);

  return $bytes;
}

```

lib/Mojolicious.pm
guest20: I think only the first secret (the one used to set new session cookies) should be subject to the length restriction, since you might still want to restore sessions cookies that were set with crappy secrets so you can re-issue them with new stronger key...

lib/Mojo/Util.pm
stigtsp: I've updated the PR with logic similar to the above, avoiding any Win32 specific code.

Also added `Crypt::URandom` to the Windows GH actions workflow, should the be in Makefile.PL instead?

kraih:
Heads up, early November is SUSE hack week, and i might work on a port of the mojo.js encrypted session implementation to Mojolicious. That's probably the best solution for security.

daleif:
[like]  Lars Madsen reacted to your message:
________________________________

jkramarz:
> Heads up, early November is SUSE hack week, and i might work on a port of the mojo.js encrypted session implementation to Mojolicious. That's probably the best solution for security.

Assuming that we're talking about https://github.com/mojolicious/mojo.js/blob/436b3263690efe65b0a8b4ecbbb09e4a3c99af5e/src/session.ts#L77 , please consider if it's not vulnerable to IV nonce reuse (https://frereit.de/aes_gcm/) and is using scrypt key derivation algorithm correctly (https://nodejs.org/api/crypto.html#cryptoscryptpassword-salt-keylen-options-callback  , see notes about choosing proper salt).

If considering change of session cookies format, what about just using Crypt::JWT library instead?

kraih:
> If considering change of session cookies format, what about just using Crypt::JWT library instead?

Good idea. Worth investigating.

guest20:
I'm not sure if JWT gives us much more than an extra header section with:
  - "which crypto to use" field (a source of vulns of honoured) and
  -  "type" which is ... "jwt"

latk:
> i might work on a port of the mojo.js encrypted session implementation to Mojolicious. That's probably the best solution for security.

Even if the session cookie format is changed, a secure method for generating secrets is needed, as provided by this PR.

Having strong keys so that the cryptographic methods can uphold their security guarantees may be more important than deciding whether to offer only Integrity (via the current use of a HMAC), or Integrity+Confidentiality (by switching to an authenticated encryption method like AES-GCM).

If porting the JavaScript code, it could be made significantly more efficient:

```js
  static async encrypt(secret: string, value: string): Promise<string> {
    const key = await scrypt(secret, 'salt', 32);
    const iv = await randomBytes(12);
    const cipher = crypto.createCipheriv('aes-256-gcm', key as crypto.CipherKey, iv);
```

1. I am not convinced that this use of scrypt() has any security benefits over a fast cryptographic hash function like SHA-2, especially if the `secret` is already strong.
2. Due to using a constant salt, the scrypt() call could be pre-computed during application startup, instead of re-running this for each encrypt() and decrypt().

Mojo.js doesn't have benchmarks involving session cookies, but if it had some then caching (or removing) the scrypt() call might save on the order of 100ms per request.

> what about just using Crypt::JWT library instead?

+1 on sticking close to standard formats like JWT, as this neatly sidesteps many pitfalls from doing low-level crypto manually. (Aside from the weakness that a JWT decoder might fail to limit which algorithms are accepted.)

That particular library has a hard dependency on `CryptX`, which provides a `CryptX::PRNG::random_bytes()` function which would remove the need to manually open `/dev/urandom`. This would also solve the problem of a secure fallback on Windows.

But pulling in a large and security-critical XS dependency just for that might not be worth it? The current (and much simpler) approach to session cookies can already be secure (if strong keys are used).

kraih:
> 
> Even if the session cookie format is changed, a secure method for generating secrets is needed, as provided by this PR.

Is it really? Why can't we outsource that problem to plugins on CPAN? Any core mojo APIs missing?

stigtsp:
> > Even if the session cookie format is changed, a secure method for generating secrets is needed, as provided by this PR.
> 
> Is it really? Why can't we outsource that problem to plugins on CPAN? Any core mojo APIs missing?

@kraih Yes. Strong keys are required for HMAC signed cookies. I have no strong opinion on what CPAN modules or syscalls are used for randomness, as long as bits are from a csprng.

For example, NIST requires HMAC keys to be a minimum of 128 bits from a cryptographic random number generator: https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-224.ipd.pdf

Problems:

1. A well known string is used as default HMAC key. Even though its documented, and a warning is printed, it provides no security and could be a vulnerability if it ends up in production. [CWE-321](https://cwe.mitre.org/data/definitions/321.html)

2. No key length requirement for users when setting keys. This leads to weak keys in the wild vulnerable to dictionary or brute force attacks, as described in [Baking Mojolicious cookies](https://www.synacktiv.com/en/publications/baking-mojolicious-cookies.html) by Antoine Cervoise. [CWE-521](https://cwe.mitre.org/data/definitions/521.html)

3. `mojo generate app` generates a weak secret using `sha1_sum $$ . steady_time . rand`, which is not from a cryptographic number generator. [CWE-338](https://cwe.mitre.org/data/definitions/338.html)

This PR aims to start fixing these problems, so let me know if there is anything more I can do to help.

kraih:
> @kraih Yes. Strong keys are required for HMAC signed cookies. I have no strong opinion on what CPAN modules or syscalls are used for randomness, as long as bits are from a csprng.
> 

Perhaps there was a miscommunication then. We make no claim to be secure by default, that's why we ask users to change their secrets and suggest to rotate regularly. The secrets we generate by default are strictly for development purposes, just like the TLS certificate we ship. It is the users responsibility to change both for production.

The default secrets being better with an optional CryptX dependency because of an alternative session implementation would merely be a positive side effect, not an end goal.

jkramarz:
Out of curiosity, I checked how some top starred projects on Github with direct Mojolicious dependency are dealing with this problem:
- zonemaster and mojolicious.org website has no sessions at all :-)
- Convos is using sha1 of (usually) some random bytes from /dev/urandom: https://github.com/convos-chat/convos/commit/54d1763ac65c05aad27ad454b4e5a62ba8352d39#diff-14d3feb01f623517c2b3475efcc7cdbcc32705fa461ab5c2ecda1d529f627ff4R134
- LANraragi is using bytes from Perl non-cryptographic rand, then storing it in (probably) web-accessible temp directory: https://github.com/Difegue/LANraragi/commit/7ba8425e0945f06f9aadae0735191dd61e2f9843
- Yancy hopes there's secret defined: https://github.com/preaction/Yancy/blob/83ae73dc31aa0f72cf5c39b79df8a121c9370385/lib/Yancy/Plugin/Auth/Token.pm#L239
- openSUSE's MirrorCache is using a string formed with Perl non-cryptographic rand and not storing it at all: https://github.com/openSUSE/MirrorCache/blob/c8d8d53b3ec522d8372f44c807af11fa47fbee7a/lib/MirrorCache/WebAPI.pm#L88
- Ado hopes that adding some magic sprinkles works: https://github.com/kberov/Ado/blob/a93056a50a8719b4c5204444c3b929e3b4716516/lib/Ado.pm#L51
- Mojowka has a static string: https://github.com/shoorick/mojowka/blob/0159a8e8e5810836c91a4d6d5be8e00e9a634373/mojowka#L244
- USGCRP's gcis won't start without proper configuration (and I hope it's not this_should_be_replaced_with_a_server_secret in production deployment, but didn't dare to check): https://github.com/USGCRP/gcis/blob/b291522f1ab7cfff850cb5f25a7491039d04f6c5/lib/Tuba.pm#L46

s1037989:
This seems like a really neat solution and appears very Mojo.  The PR looks really well made and followed all the contributing guidelines.  I'd love to see this merged.
... are there any cons to merging this in?

kraih:
I'm more and more leaning against this proposal. Especially the secret files don't feel very Mojolicious. We'll see after encrypted sessions have been implemented.

christopherraa:
Personally I am :-1:  on file based solutions like the one proposed. Any writing to disk also introduce a potential additional attack-vector and on top of that deviates from the move we've been seeing in recent years towards more and more env-based approach to configuration of services, especially for containerized workloads. When doing application hardening and insulation we actively try to limit the applications access to and use of the file system, except where strictly necessary.

Those were my $.02

guest20:
@christopherraa If the secret isn't persisted, all cookies are invalidated every time the application restarts. The only other choice is making the secret mandatory and aborting startup if it's not set... which gets awful CrashLoopBackoff'y

Grinnz:
Not to mention that there are numerous applications out there which have never touched the session cookie or at least have not used it for anything sensitive. I personally think it is very important that it be secure by default, but I also have reservations about writing to disk for permissions/deployment reasons; but it would go along with the default behavior of hypnotoad to write a pidfile to the project directory if not configured otherwise. (The default behavior of the application log to log to the project directory has been removed, as it simply logs to STDERR by default now)

kraih:
I have to agree with @christopherraa that ephemeral parts of the config, such as secrets, need to be easily updatable in container environments.

s1037989:
What about using a machine-generated unique ID, and then fall back on the moniker if the file can be read?

/var/lib/dbus/machine-id

kraih:
#2212 is what i had in mind for encrypted sessions.

lib/Mojo/Util.pm
```diff
@@ -379,6 +381,25 @@ sub unquote {
   return $str;
 }
 
+sub urandom_bytes {
+  my $num = shift || 32;
+
+  return Crypt::URandom::urandom($num) if CRYPT_URANDOM;
+
+  croak 'Cannot find /dev/urandom, install Crypt::URandom from CPAN' unless -e '/dev/urandom';
```
robrwo, line 389: Is it worth having a fallback to read `/dev/urandom` vs simply requiring the user to have this module installed if they are using this feature?

carnil:
CVE-2024-58135 is related to this merge request.

https://lists.security.metacpan.org/cve-announce/msg/29241187/

Grinnz:
If this proceeds, I would recommend now using (instead of Crypt::URandom) Crypt::SysRandom which can optionally use the preferred Crypt::SysRandom::XS if that is also installed.

lib/Mojo/Util.pm
Grinnz: The point of this is to be secure by default, so if it doesn't work without optional dependencies then the default state isn't really any better than it was.

kraih:
> CVE-2024-58135 is related to this merge request.
> 
> https://lists.security.metacpan.org/cve-announce/msg/29241187/

Should have been resolved already with [a previous change](https://github.com/mojolicious/mojo/commit/c82071556c569a251152892c8cc2fd0ad5a4be54). This PR is merely about doing it without dependencies.

Grinnz:
Given that change, perhaps it would be simpler to use the CryptX functions for all optional dependencies, and focus on what can be provided if CryptX is not installed.

Leont:
> If this proceeds, I would recommend now using (instead of Crypt::URandom) Crypt::SysRandom which can optionally use the preferred Crypt::SysRandom::XS if that is also installed.

Honestly Crypt::SysRandom is [small enough}(https://metacpan.org/dist/Crypt-SysRandom/source/lib/Crypt/SysRandom.pm) that the most Mojolicious solution might be to inline it.

kraih:
Closing this issue since it looks like it's been abandoned. The filed CVEs present no concrete attack vectors, so no immediate action is necessary. As far as i'm concerned, this was resolved with the addition of CryptX support.

PR #2212, opened 2024-11-20
Add support for encrypted sessions with CryptX

kraih:
SUSE Hack Week has started, so this is my first proposal for sessions with encrypted cookies. I've tried to stay as close to the mojo.js implementation as possible. To address #2200 i've also added a `Mojo::Util::generate_secret` function.

The feature is only available if [CryptX](https://metacpan.org/pod/CryptX) is installed, and needs to be enabled with `$app->sessions->encrypted(1)`.

### DIFF
--- a/lib/Mojo/Util.pm
+++ b/lib/Mojo/Util.pm
@@ -21,6 +21,17 @@ use Symbol             qw(delete_package);
 use Time::HiRes        ();
 use Unicode::Normalize ();
 
+# Encryption support requires CryptX 0.080+
+use constant CRYPTX => $ENV{MOJO_NO_CRYPTX} ? 0 : !!(eval {
+  require CryptX;
+  require Crypt::AuthEnc::ChaCha20Poly1305;
+  require Crypt::KeyDerivation;
+  require Crypt::Misc;
+  require Crypt::PRNG;
+  CryptX->VERSION('0.080');
+  1;
+});
+
 # Check for monotonic clock support
 use constant MONOTONIC => !!eval { Time::HiRes::clock_gettime(Time::HiRes::CLOCK_MONOTONIC()) };
 
@@ -64,15 +75,15 @@ my $UNQUOTED_VALUE_RE = qr/\G=\s*([^;, ]*)/;
 # HTML entities
 my $ENTITY_RE = qr/&(?:\#((?:[0-9]{1,7}|x[0-9a-fA-F]{1,6}));|(\w+[;=]?))/;
 
-# Encoding and pattern cache
-my (%ENCODING, %PATTERN);
+# Encoding, encryption and pattern caches
+my (%ENCODING, %ENCRYPTION, %PATTERN);
 
 our @EXPORT_OK = (
-  qw(b64_decode b64_encode camelize class_to_file class_to_path decamelize decode deprecated dumper encode),
-  qw(extract_usage getopt gunzip gzip header_params hmac_sha1_sum html_attr_unescape html_unescape humanize_bytes),
-  qw(md5_bytes md5_sum monkey_patch network_contains punycode_decode punycode_encode quote scope_guard secure_compare),
-  qw(sha1_bytes sha1_sum slugify split_cookie_header split_header steady_time tablify term_escape trim unindent),
-  qw(unquote url_escape url_unescape xml_escape xor_encode)
+  qw(b64_decode b64_encode camelize class_to_file class_to_path decamelize decode decrypt_cookie deprecated dumper),
+  qw(encode encrypt_cookie extract_usage generate_secret getopt gunzip gzip header_params hmac_sha1_sum),
+  qw(html_attr_unescape html_unescape humanize_bytes md5_bytes md5_sum monkey_patch network_contains punycode_decode),
+  qw(punycode_encode quote scope_guard secure_compare sha1_bytes sha1_sum slugify split_cookie_header split_header),
+  qw(steady_time tablify term_escape trim unindent unquote url_escape url_unescape xml_escape xor_encode)
 );
 
 # Aliases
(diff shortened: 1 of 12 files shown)


kraih:
As an alternative i will now look into using JWT for the rest of Hack Week.

lib/Mojolicious/Sessions.pm
```diff
@@ -58,16 +59,13 @@ sub store {
     samesite => $self->samesite,
     secure   => $self->secure
   };
-  $c->signed_cookie($self->cookie_name, $value, $options);
+  my $method = $self->encrypted ? 'encrypted_cookie' : 'signed_cookie';
+  $c->$method($self->cookie_name, $value, $options);
 }
 
-sub _deserialize { Mojo::JSON::decode_json($_[0] =~ s/\}\KZ*$//r) }
+sub _deserialize { Mojo::JSON::decode_json($_[0]) }
```
Grinnz, line 66: This will fail to deserialize existing signed sessions with the padding support removed

kraih: Well spotted.

lib/Mojolicious/Command/Author/generate/app.pm
```diff
@@ -196,7 +196,7 @@ done_testing();
 </p>
 
 @@ config
-% use Mojo::Util qw(sha1_sum steady_time);
+% use Mojo::Util qw(generate_secret);
 ---
 secrets:
-  - <%= sha1_sum $$ . steady_time . rand  %>
+  - <%= generate_secret() %>
```
latk, line 202: The generated secret remains insecure unless CryptX is installed.

This would be fixed by #2200.

lib/Mojo/Util.pm
```diff
@@ -130,6 +153,17 @@ sub dumper { Data::Dumper->new([@_])->Indent(1)->Sortkeys(1)->Terse(1)->Useqq(1)
 
 sub encode { _encoding($_[0])->encode("$_[1]", 0) }
 
+sub encrypt_cookie {
+  my ($value, $key, $salt) = @_;
+  croak 'CryptX 0.080+ required for encrypted cookie support' unless CRYPTX;
+
+  my $dk = Crypt::KeyDerivation::pbkdf2($key, $salt);
```
latk, line 160: The expanded key only has to be computed once at application start up. Running pbkdf2 each time a cookie is encrypted or decrypted is computationally expensive and will tank Mojo performance.

This is the same performance problem as in mojo.js, as discussed in https://github.com/mojolicious/mojo/pull/2200#issuecomment-2407646319. However, mojo.js uses the more modern (and more computationally expensive) scrypt key-derivation function.

It could also be sensible to tune the PBKDF2 cost parameter. The `Crypt::KeyDerivation` documentation says that it defaults to 5000 iterations, whereas 2023-era OWASP guidelines suggest 600,000 iterations for password storage purposes. However, a low cost parameter may actually be appropriate here.

latk (commented):
The encryption part looks reasonably sound. :+1:

AES-GCM is a slightly dated but reasonable choice and seems to be used correctly here. Alternatively, the more modern ChaCha20-Poly1305 could be used.

However, the signed cookie mechanism remains insecure, so the fixes in #2200 would remain valuable.

Also, the unnecessary repeated invocations of PBKDF2 are a major performance problem. This value should likely be cached at application startup.

(PBKDF2 itself is no longer a state of the art key derivation function, and the default cost parameter of 5000 is a 2005-era security choice. But its use here seems acceptable as we only need key-stretching, not cracking-resistant password hashes.)

lib/Mojo/Util.pm
kraih: It's now cached.

kraih:
I've switched to `ChaCha20-Poly1305`.

lib/Mojolicious/Command/Author/generate/app.pm
kraih: And we are ok with that.

abraxxa: That invalidates Mojolicious as a valid framework to use in the EU as NIS2 requires secure-by-default and no predefines secrets or credentials allowed.

kraih:
My experiments with JWT today have been rather disappointing. Primary focus for most modules seems to be JWS and not JWE, and i've run into more than a few inconsistencies and errors. Also the cookies end up much larger for the simple user login case, which i consider the most common use.

So i expect this PR to end up as my main proposal for a new sessions implementation.

lib/Mojolicious/Command/Author/generate/app.pm
kraih: @abraxxa No.

kraih: @abraxxa The burden of proof is on you. If you make a handwavy claim you get a handwavy rebuttal from me. :)

abraxxa: If you search https://eur-lex.europa.eu/legal-content/EN/TXT/HTML/?uri=CELEX:32022L2555 for 'by default' you see the points I'm referring to, (98) and (104).

kraih: @abraxxa I don't see how any of that applies to the situation here. So I'm just going to assume you've misunderstood the feature.

kraih: And on that note we should end the topic before things get out of hand.

jberger (commented):
I'm a bit late but 👍

guest20:
Will this be able to automatically upgrade existing, signed but non-encrypted sessions or does everybody have to log in again?

kraih:
Everybody has to log in again.

PR #2237, opened 2025-03-11
Mojolicious::Guides: explain the convention of method examples that return the invocant

Grinnz:
### Summary
Add a bit of documentation to the main guides CONVENTIONS section explaining the common example of a method that returns its invocant for chaining.

### Motivation
This construct is used a lot in doc examples, but doesn't represent how the method should actually be called and its intent may not be obvious.


kraih (approved):

christopherraa (approved):
:+1:

PR #2239, opened 2025-03-28
psgi env: fix double encoding of path parts

sparrow2009:
Fix double encoding of path parts when parsing a PSGI env with a `SCRIPT_NAME` set.

### DIFF
--- a/lib/Mojo/Message/Request.pm
+++ b/lib/Mojo/Message/Request.pm
@@ -197,6 +197,8 @@ sub _parse_env {
     $base->path->parse($value =~ m!/$! ? $value : "$value/");
 
     # Remove SCRIPT_NAME prefix if necessary
+    # Make sure $path is decoded and unescaped as expected by to_string
+    $path->parts;
     my $buffer = $path->to_string;
     $value  =~ s!^/|/$!!g;
     $buffer =~ s!^/?\Q$value\E/?!!;
--- a/t/mojo/request_cgi.t
+++ b/t/mojo/request_cgi.t
@@ -2,6 +2,7 @@ use Mojo::Base -strict;
 
 use Test::More;
 use Mojo::Message::Request;
+use Mojo::Util 'encode', 'url_escape';
 
 subtest 'Parse Lighttpd CGI environment variables and body' => sub {
   my $req = Mojo::Message::Request->new;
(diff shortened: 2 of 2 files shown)


lib/Mojo/Message/Request.pm
```diff
@@ -197,6 +197,8 @@ sub _parse_env {
     $base->path->parse($value =~ m!/$! ? $value : "$value/");
 
     # Remove SCRIPT_NAME prefix if necessary
+    # Make sure $path is decoded and unescaped as expected by to_string
```
kraih, line 200: Reads like you're putting things that should be in a nice detailed commit message into comments.

t/mojo/request_cgi.t
```diff
@@ -124,6 +125,29 @@ subtest 'Parse CGI environment with maximum message size' => sub {
...
+subtest 'Parse CGI environment with SCRIPT_NAME set and path part with UTF8 character' => sub {
+  my $req = Mojo::Message::Request->new;
+  my $script_name = '/app/';
+  my $path_stub   = 'some/action/blub/';
+  my $part_utf8   = encode(q(utf-8), "\x{1d120}");
+  my $path        = $path_stub.url_escape($part_utf8);
+  # test $env is a stripped down version of the result of:
+  #   use HTTP::Request::Common 'GET';
+  #   use HTTP::Message::PSGI 'req_to_psgi';
+  #   $env = req_to_psgi( GET( 'http://www.example.com'.$script_name.$path ), SCRIPT_NAME => $script_name );
```
kraih, line 137: Same here, this is ok in a commit message, but serves no real purpose in a test.

sparrow2009: Also has now been incorporated into the commit message. That part was never actually meant to be part of the final version to be merged. 

As from browsing the commit history I got under the impression that it is expected to keep commit messages lean. So I did not want this to be part of it, but decided to put it in intermittently as a code comment for making reviewers life easier.

t/mojo/request_cgi.t
```diff
@@ -124,6 +125,25 @@ subtest 'Parse CGI environment with maximum message size' => sub {
     'right absolute URL';
 };
 
+subtest 'Parse CGI environment with SCRIPT_NAME set and path part with UTF8 character' => sub {
```
kraih, line 128: It's `UTF-8`.

PR #2243, opened 2025-04-10
Add download methods to Mojo::UserAgent and Mojo::UserAgent::Transactor

kraih:
This is something that has come up for me at work. While trying to explain how to build a client for resumable large file downloads, i realized how tricky doing it right actually is. We already support range requests on the server-side in Mojolicious, so from a consistency point of view the feature would probably fit in.

The `download` and `download_p`methods in `Mojo::UserAgent`  are intentionally very bare-bones. But the option argument (that currently only allows for headers to be set) leaves room for future additions. And the `Mojo::UserAgent::Transactor` method makes more advanced use cases possible.

An argument against the feature might be that it broadens the scope of `Mojo::UserAgent` too much, but i'm ok with that.

### DIFF
--- a/lib/Mojo/UserAgent.pm
+++ b/lib/Mojo/UserAgent.pm
@@ -3,6 +3,7 @@ use Mojo::Base 'Mojo::EventEmitter';
 
 # "Fry: Since when is the Internet about robbing people of their privacy?
 #  Bender: August 6, 1991."
+use Carp qw(croak);
 use Mojo::IOLoop;
 use Mojo::Promise;
 use Mojo::Util qw(monkey_patch term_escape);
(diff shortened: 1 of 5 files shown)


kraih:
Another option i was considering:
```
use Mojo::File qw(path);

path('/home/sri/test.tar.gz')->download('https://example.com/test.tar.gz');
path('/home/sri/test.tar.gz')->download('https://example.com/test.tar.gz', {headers => {Accept => '*/*'}});
```
As only option that would be too limited, but might make sense in combination with `Mojo::UserAgent::Transactor::download`. Then only the simple cases would use `Mojo::File`, and the rest simply works with transactions.

kraih:
Went with the other solution.

PR #2244, opened 2025-04-11
Add downloads method to Mojo::File and Mojo::UserAgent::Transactor

kraih:
Alternative to #2243. Personally, i think i prefer this solution.


t/mojo/file_download.t
```diff
@@ -0,0 +1,131 @@
...
+  is $file->slurp, '', 'right content';
+
+  my $ua2   = Mojo::UserAgent->new(max_redirects => 5);
+  my $file2 = $dir->child('redirect2.txt');
+  ok $file2->download('/redirect', {ua => $ua2}), 'file downloaded';
+  ok -e $file2,                                   'file exists';
+  is $file2->slurp, 'CHUNK1CHUNK2CHUNK3', 'right content';
+};
+
+subtest 'Exsiting file is larger' => sub {
```
christopherraa, line 97:
```suggestion
subtest 'Existing file is larger' => sub {
```

christopherraa (commented):
To me this looks good, and it seems like a good addition to the user agent / transactor. Only thing I noticed was a small type. Added a suggested fix for that.

PR #2252, opened 2025-05-20
Refuse to set signed or encrypted cookies with an insecure default secret

Grinnz:
### Summary
Rough attempt at taking the approach of dying if encrypted or signed cookies are requested to be set but the application secret is still set to the default of the application moniker.

This is fairly straightforward to do in the signed_cookie and encrypted_cookie methods, but this also requires a check in the session method, because setting the session cookie occurs too late to error out sensibly. I'm not satisfied with this check in the session method, because: there still may be instances where this session method is called while a response is already being rendered; and session data can be set by assigning to `$c->session->{foo}` and there's no way to know that will happen at this juncture.

This also affects use of features using the session implicitly, such as the flash and csrf_token default helpers.

### Motivation
The risk of insecure secrets being used in applications using signed or encrypted cookies is not sufficiently mitigated by the annoying warning in the logs. But we don't want to cause applications to fail if they are not using these features.

### References
Implements a part of https://github.com/mojolicious/mojo/pull/2200 by refusing to set signed or encrypted cookies by default unless the secret has been changed from the default.

### DIFF
--- a/lib/Mojolicious/Controller.pm
+++ b/lib/Mojolicious/Controller.pm
@@ -59,6 +59,10 @@ sub encrypted_cookie {
   my $app     = $self->app;
   my $secret  = $app->secrets->[0];
   my $moniker = $app->moniker;
+
+  Carp::croak 'Your secret passphrase must be changed to set encrypted cookies (see FAQ for more)'
+    if !$ENV{MOJO_ALLOW_INSECURE_SECRET} and (!length $secret or $secret eq $moniker);
+
   return $self->cookie($name, Mojo::Util::encrypt_cookie($value, $secret, $moniker), $options);
 }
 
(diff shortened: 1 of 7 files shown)


kraih:
It seems a little overengineered, but i also don't know a better solution.

lib/Mojolicious/Controller.pm
```diff
@@ -242,12 +246,24 @@ sub session {
...
   my $session = $stash->{'mojo.session'} //= {};
+
+  # Require changed secret if session data will be set
+  my $set_operation = !!(@_ > 1 || ref $_[0]);
+  if (!$ENV{MOJO_ALLOW_INSECURE_SECRET} and (keys %$session or $set_operation)) {
+    my $app     = $self->app;
+    my $secret  = $app->secrets->[0];
+    my $moniker = $app->moniker;
+
+    Carp::croak 'Your secret passphrase must be changed to set session data (see FAQ for more)'
```
guest20, line 258: I don't think these secrets are "passphrase"s (and the FAQ doesn't call them that either)

lib/Mojolicious/Controller.pm
```diff
@@ -262,8 +278,15 @@ sub signed_cookie {
   # Request cookie
   return $self->every_signed_cookie($name)->[-1] unless defined $value;
 
+  my $app     = $self->app;
+  my $secret  = $app->secrets->[0];
+  my $moniker = $app->moniker;
+
+  Carp::croak 'Your secret passphrase must be changed to set signed cookies (see FAQ for more)'
+    if !$ENV{MOJO_ALLOW_INSECURE_SECRET} and (!length $secret or $secret eq $moniker);
```
guest20, line 286: Would it make sense to move this check into `$app->session`?

guest20: yes, sorry, github threw away my comment, and i crankly re-typed it wrong.

I mean should it be in `$app->secrets`

Grinnz: That attribute may be accessed by applications that don't actually do anything with it, such as Mojolicious::Plugin::Mount. But perhaps we can work around that more easily than the problems with this approach.

PR #2280, opened 2025-10-06
Both HTTP method and query can have a value of '0'

karenetheridge:
Method ABNF is in https://datatracker.ietf.org/doc/html/rfc9110#appendix-A and equivalent to the regex: ^[a-zA-Z0-9!#$%&'*+.^_`|~-]+$

I couldn't find a good place to test this, as parse_env is only called from the cgi and psgi tests; open to suggestion and will happily add some.

### DIFF
--- a/lib/Mojo/Message/Request.pm
+++ b/lib/Mojo/Message/Request.pm
@@ -176,10 +176,10 @@ sub _parse_env {
   $headers->content_length($env->{CONTENT_LENGTH}) if $env->{CONTENT_LENGTH};
 
   # Query
-  $url->query->parse($env->{QUERY_STRING}) if $env->{QUERY_STRING};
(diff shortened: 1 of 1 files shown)


lib/Mojo/Message/Request.pm
```diff
@@ -176,10 +176,10 @@ sub _parse_env {
   $headers->content_length($env->{CONTENT_LENGTH}) if $env->{CONTENT_LENGTH};
 
   # Query
-  $url->query->parse($env->{QUERY_STRING}) if $env->{QUERY_STRING};
+  $url->query->parse($env->{QUERY_STRING}) if defined $env->{QUERY_STRING};
```
Grinnz: `length` may be a better test

karenetheridge: Fair, an empty Mojo::Parameters in the url object is equivalent to not having one defined at all.

PR #2290, opened 2025-11-22
Add support for 'webp' image format

niceperl:
### Summary

Add support for 'webp' image format

### Motivation

It's a common mime type 

### References

https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/MIME_types/Common_types


rawleyfowler:
Can you add a test for this @niceperl ?

PR #2291, opened 2025-11-22
Add silent option to daemon command

melo:
### Summary
Adds a `-s` or `--silent` option to the `daemon` command.

This will disable the startup "Web application available at..." message.

### Motivation
On production environments, specially on K8s or ECS, the message is not particularly useful.


kraih:
This should be very easy to test.

melo:
Yes, I'll look for the proper place to put the tests.

Thank you.

PR #2300, opened 2026-01-06
ensure timely cleanup of UserAgent temporary files

dakkar:
### Summary

Ensure that temporary files created by `Mojo::UserAgent` are cleaned up as soon as the transaction is destroyed.

### Motivation

the `$cb` that `sub start` passes to `sub _start` is:

    sub { shift->ioloop->stop; $tx = shift }

which closes over `$tx`, which owns the response, its content, its asset, and its tempfile (when the asset is a `Mojo::Asset::File`)

if we leave the callback inside the `{connections}` hashref, the tempfile will not be cleaned up until the connection is next re-used, which may not happen any time soon, especially for long-lived applications that download data from many sources

deleting the callback just before calling it, ensures a timely cleanup

I've also added a test to prove the problem and the fix. All tests pass.

### DIFF
--- a/t/mojo/user_agent.t
+++ b/t/mojo/user_agent.t
@@ -363,16 +363,18 @@ ok $tx->res->headers->content_length, 'has "Content-Length" value';
 ok !$tx->is_empty,                    'transaction is not empty';
 is $tx->res->body, 'x' x 262144, 'right content';
 
-# memory->file asset promotion, plus timely cleanup
-{
+subtest 'memory->file asset promotion, plus timely cleanup' => sub {
     local $ENV{MOJO_MAX_MEMORY_SIZE} = 262144/2;
     $tx = $ua->get('/huge');
-}
-is $tx->res->body, 'x' x 262144, 'right content';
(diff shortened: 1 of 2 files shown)


t/mojo/user_agent.t
```diff
@@ -363,6 +363,17 @@ ok $tx->res->headers->content_length, 'has "Content-Length" value';
...
+# memory->file asset promotion, plus timely cleanup
+{
+    local $ENV{MOJO_MAX_MEMORY_SIZE} = 262144/2;
+    $tx = $ua->get('/huge');
+}
+is $tx->res->body, 'x' x 262144, 'right content';
+my $filename = $tx->res->content->asset->path;
+ok $filename, 'the huge body should be saved to a temporary file';
+undef $tx;
+ok !(-e $filename), 'the temporary file should be deleted as soon as the transaction is destroyed';
```
kraih: Subtests for new tests, instead of comments.

dakkar: fair, I kept to the same style as the rest of the file

I'll make this a subtest

PR #2307, opened 2026-03-20
fix serialization of cookies

karenetheridge:
- only one Cookie header is permitted; all cookies must be in a single string (see https://www.rfc-editor.org/rfc/rfc6265#section-5.4)
- the delimiter between individual cookie-pairs is "; ", not ", " (see https://www.rfc-editor.org/rfc/rfc6265#section-4.2.1)

### DIFF
--- a/lib/Mojo/Headers.pm
+++ b/lib/Mojo/Headers.pm
@@ -20,11 +20,12 @@ my %NAMES = map { lc() => $_ } (
 for my $header (keys %NAMES) {
   my $name = $header;
   $name =~ y/-/_/;
+  my $delim = $name eq 'cookie' ? '; ' : ', ';
   monkey_patch __PACKAGE__, $name, sub {
     my $self = shift;
     $self->{headers}{$header} = [@_] and return $self if @_;
     return undef unless my $headers = $self->{headers}{$header};
-    return join ', ', @$headers;
+    return join($delim, @$headers);
   };
 }
 
@@ -48,7 +49,7 @@ sub add {
 sub append {
   my ($self, $name, $value) = @_;
   my $old = $self->header($name);
-  return $self->header($name => defined $old ? "$old, $value" : $value);
+  return $self->header($name => defined $old ? ($old . (lc $name eq 'cookie' ? '; ' : ', ') . $value) : $value);
 }
 
 sub clone {
@@ -90,8 +91,9 @@ sub header {
   # Replace
   return $self->remove($name)->add($name, @_) if @_;
 
-  return undef unless my $headers = $self->{headers}{lc $name};
-  return join ', ', @$headers;
+  $name = lc $name;
+  return undef unless my $headers = $self->{headers}{$name};
+  return join(($name eq 'cookie' ? '; ' : ', '), @$headers);
 }
 
 sub is_finished { (shift->{state} // '') eq 'finished' }
@@ -178,7 +180,14 @@ sub to_string {
 
   # Make sure multi-line values are formatted correctly
   my @headers;
-  for my $name (@{$self->names}) { push @headers, "$name: $_" for @{$self->{headers}{lc $name}} }
+  for my $name (@{$self->names}) {
+    if ($name eq 'Cookie') {
+      push @headers, "$name: " . join('; ', @{$self->{headers}{lc $name}});
+    }
+    else {
+      push @headers, "$name: $_" for @{$self->{headers}{lc $name}};
+    }
+  }
 
   return join "\x0d\x0a", @headers;
 }
--- a/t/mojo/headers.t
+++ b/t/mojo/headers.t
@@ -64,7 +64,6 @@ subtest 'Common headers' => sub {
   is $headers->content_range('foo')->content_range,                             'foo', 'right value';
   is $headers->content_security_policy('foo')->content_security_policy,         'foo', 'right value';
   is $headers->content_type('foo')->content_type,                               'foo', 'right value';
-  is $headers->cookie('foo')->cookie,                                           'foo', 'right value';
   is $headers->dnt('foo')->dnt,                                                 'foo', 'right value';
   is $headers->date('foo')->date,                                               'foo', 'right value';
   is $headers->etag('foo')->etag,                                               'foo', 'right value';
@@ -101,6 +100,9 @@ subtest 'Common headers' => sub {
   is $headers->referer,                   'foo', 'right value';
   is $headers->referer('bar')->referer,   'bar', 'right value';
   is $headers->referrer,                  'bar', 'right value';
+
+  is $headers->cookie('foo')->cookie,    'foo',  'right value for single cookie';
+  is $headers->cookie('a', 'b')->cookie, 'a; b', 'right value for multiple cookies';
 };
 
 subtest 'Clone' => sub {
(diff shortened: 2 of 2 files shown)


t/mojo/headers.t
```diff
@@ -177,6 +177,15 @@ subtest 'Set headers from hash' => sub {
   my $headers = Mojo::Headers->new;
   $headers->from_hash({Connection => 'close', 'Content-Type' => 'text/html'});
   is_deeply $headers->to_hash, {Connection => 'close', 'Content-Type' => 'text/html'}, 'right structure';
+  is $headers->to_string, "Connection: close\r\nContent-Type: text/html", 'right string';
```
Grinnz, line 182: Is the order of these headers when stringified guaranteed?

karenetheridge: The values are stored in an arrayref, so they're output in the same order that they were added. However, `from_hash` uses a hashref so the order there is not deterministic.  I'll fix the test.

karenetheridge: Ah, no, I'm wrong, all accesses to the data structure use the `names` method, which sorts the entries, so the order is consistent.

kraih:
~~I think there was a reason we don't do this, but i don't quite remember. Cookie headers had to be split up into multiple headers with the same name and not joined as a string or it caused some bugs. 🤔~~

Edit: I was mixing up `Cookie` and `Set-Cookie`.

lib/Mojo/Headers.pm
```diff
@@ -47,8 +47,9 @@ sub add {
 
 sub append {
   my ($self, $name, $value) = @_;
-  my $old = $self->header($name);
-  return $self->header($name => defined $old ? "$old, $value" : $value);
+  my $old   = $self->header($name);
+  my $delim = $name eq 'Cookie' ? '; ' : ',';
```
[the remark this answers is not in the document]
karenetheridge: thanks copilot, I'll make these fixes.

t/mojo/headers.t
```diff
@@ -177,6 +177,15 @@ subtest 'Set headers from hash' => sub {
...
+
+  $headers = Mojo::Headers->new;
+  $headers->from_hash({'Cookie' => ['a=b', 'c=d']});
+  is_deeply $headers->to_hash, {Cookie => 'a=b; c=d'}, 'right structure';
+  is $headers->to_string, "Cookie: a=b; c=d", 'right string';
+  $headers->from_hash({'Cookie' => ['e=f']});
+  is_deeply $headers->to_hash, {Cookie => 'a=b; c=d; e=f'}, 'right structure';
+  is $headers->to_string, "Cookie: a=b; c=d; e=f", 'right string';
```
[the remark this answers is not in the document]
karenetheridge: fixed; thanks copilot!

lib/Mojo/Headers.pm
```diff
@@ -178,7 +179,13 @@ sub to_string {
 
   # Make sure multi-line values are formatted correctly
   my @headers;
-  for my $name (@{$self->names}) { push @headers, "$name: $_" for @{$self->{headers}{lc $name}} }
+  for my $name (@{$self->names}) {
+    if ($name eq 'Cookie') {
+      push @headers, "$name: " . join('; ', @{$self->{headers}{lc $name}});
+      next;
```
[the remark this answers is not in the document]
karenetheridge: fixed; thanks copilot!

lib/Mojo/Headers.pm
```diff
@@ -91,7 +92,7 @@ sub header {
   return $self->remove($name)->add($name, @_) if @_;
 
   return undef unless my $headers = $self->{headers}{lc $name};
-  return join ', ', @$headers;
+  return join(($name eq 'Cookie' ? '; ' : ', '), @$headers);
```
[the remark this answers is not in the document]
karenetheridge: I'm not sure if this is valid, but it's harmless to do a lc comparison so I added it anyway.

lib/Mojo/Headers.pm
```diff
@@ -178,7 +180,13 @@ sub to_string {
 
   # Make sure multi-line values are formatted correctly
   my @headers;
-  for my $name (@{$self->names}) { push @headers, "$name: $_" for @{$self->{headers}{lc $name}} }
+  for my $name (@{$self->names}) {
+    if (lc $name eq 'cookie') {
+      push @headers, "$name: " . join('; ', @{$self->{headers}{lc $name}});
+      next;
+    }
+    push @headers, "$name: $_" for @{$self->{headers}{lc $name}};
```
kraih: Doesn't `lc` run twice here when it doesn't have to?

lib/Mojo/Headers.pm
```diff
@@ -24,7 +24,7 @@ for my $header (keys %NAMES) {
     my $self = shift;
     $self->{headers}{$header} = [@_] and return $self if @_;
     return undef unless my $headers = $self->{headers}{$header};
-    return join ', ', @$headers;
+    return join((lc $name eq 'cookie' ? '; ' : ', '), @$headers);
```
kraih: This `lc` seems pointless.

lib/Mojo/Headers.pm
```diff
@@ -47,8 +47,10 @@ sub add {
 
 sub append {
   my ($self, $name, $value) = @_;
-  my $old = $self->header($name);
-  return $self->header($name => defined $old ? "$old, $value" : $value);
+  my $old   = $self->header($name);
+  my $delim = lc $name eq 'cookie' ? '; ' : ', ';
+  my $ret   = $self->header($name => defined $old ? ($old . $delim . $value) : $value);
+  return $ret;
```
kraih: Why store `$ret` in a variable?

lib/Mojo/Headers.pm
```diff
@@ -91,7 +93,7 @@ sub header {
   return $self->remove($name)->add($name, @_) if @_;
 
   return undef unless my $headers = $self->{headers}{lc $name};
-  return join ', ', @$headers;
+  return join((lc $name eq 'cookie' ? '; ' : ', '), @$headers);
```
kraih: This also calls `lc` twice unnecessarily?

kraih:
The header code is quite performance critical, any CPU cycles you can save will add up.

lib/Mojo/Headers.pm
karenetheridge: oops, looks like an artifact from when I had debugging information in there.

I will go over everything thoroughly tomorrow.

karenetheridge:
cleaned up, with unnecessary lc removed:

- in the monkey-patched named subs, $name is always in lowercase
- in `append` and `header`, name could be in any case and must be normalized
- in `to_string`, the `names` method always returns uppercase names

I also inlined a local variable so a string comparison is done only when needed, and created a local var in the monkey-patched method to avoid a runtime calculation.

lib/Mojo/Headers.pm
karenetheridge: fixed.

lib/Mojo/Headers.pm
karenetheridge: fixed.

lib/Mojo/Headers.pm
karenetheridge: fixed.

kraih:
Rebase the commit to get tests to pass.

karenetheridge:
rebased; not sure why CI is still failing.

rawleyfowler:
@karenetheridge Your branch is still behind upstream

karenetheridge:
Not "still" - it was up to date at the time.

rawleyfowler:
Oh my bad, I just realized this is a month old. Sorry.

PR #2314, opened 2026-04-29
do not unconditionally override tls_options

krauro:
### Summary
Only set the `SSL_hostname` option when it is not provided

### Motivation
This respects the users' `$ua->tls_options` configuration


tagg:
Partly the same as #2148

t/mojo/certs/client.key
krauro, line 1: I have updated all certs to use 2048 bit keys. Otherwise on my system I was running into the following openssl error:
```
[format:PEM] error:0A00018F:SSL routines::ee key too small
```
While there are workarounds for most systems defaulting to openssl sec-level 2, I think its probably better to update the test certs here. I suspect @tagg was having trouble running the tests for the same reason

t/mojo/ioloop_tls.t
```diff
@@ -223,6 +228,43 @@ is $client_close, 1,         'client emitted close event once';
 ok $running,     'loop was running';
 ok !$server_err, 'no error';
 
+# Valid client and server certificate with custom verifycn_name
```
kraih: Use a subtest block for new tests.

kraih:
Squash commits.

krauro:
Squashed and now using a subtest block

lib/Mojo/IOLoop/TLS.pm
```diff
@@ -58,8 +58,8 @@ sub _expand {
     $tls->{SSL_key_file}  ||= $KEY;
   }
   else {
-    $tls->{SSL_hostname}      = IO::Socket::SSL->can_client_sni ? $args->{address} : '';
-    $tls->{SSL_verifycn_name} = $args->{address};
+    $tls->{SSL_hostname}      ||= IO::Socket::SSL->can_client_sni ? $args->{address} : '';
+    $tls->{SSL_verifycn_name} ||= $args->{address};
```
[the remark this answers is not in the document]
krauro: `SSL_verifycn_name` should _never_ be set to an empty string as the common name of a certificate can never be an empty string and verification against an empty string will always fail.

While `SSL_hostname` can also never be an empty string, it looks like IO::Socket::SSL uses it as a way to [disable setting the TLS SNI extension](https://github.com/noxxi/p5-io-socket-ssl/blob/master/lib/IO/Socket/SSL.pm#L780-L783).

krauro:
The windows CI failure looks unrelated

kraih:
You can rebase to fix the Windows failure.

krauro:
Hello! Would appreciate if someone could take a look to review and merge this 😄

Grinnz (approved):

kraih (approved):

PR #2315, opened 2026-05-11
Add test coverage for text_* Test::Mojo methods

charliesheh:
Adds coverage for the following Test::Mojo methods in t/test/mojo.t:

- text_is
- text_isnt
- text_like
- text_unlike

The new tests validate:
- exact text matching
- negative text matching
- regex similarity matching
- regex negative matching
- custom description handling

This improves Test::Mojo method coverage consistency alongside the existing attr_* and content_* tests.


kraih:
Patch doesn't do what you describe it does.

PR #2317, opened 2026-05-28
Provide high quality randomness, even in absence of CryptX

Leont:
### Summary
This adds a new utility function `generate_random`, that returns a given amount of cryptographic randomness, and refactors `generate_secret` to use it.

Previously `generate_secret` would fall back to `rand()` if `CryptX` wasn't available, now it will use either `/dev/urandom` (Unix), or `RtlGenRandom` (Windows) as a fallback.

This code is based on Crypt::SysRandom but adapted to Mojo's style. Most importantly, `Crypt::SysRandom::XS` is replaced by `Crypt::PRNG`. This is a proven piece of logic that is known to work on all widely deployed platforms.

### Motivation
The security of many things rely on having cryptographically good randomness. This makes sure that is available in all situations and to all Mojo applications.

The new function is not only useful inside of Mojolicious itself, but also has many use-cases in the wider ecosystem (e.g. Auth, CSRF, Captcha).

### References

* https://github.com/mojolicious/mojo/pull/2200
* [CVE-2024-58135](https://www.cve.org/CVERecord?id=CVE-2024-58135)

### DIFF
--- a/lib/Mojo/Util.pm
+++ b/lib/Mojo/Util.pm
@@ -6,6 +6,7 @@ use Data::Dumper   ();
 use Digest::MD5    qw(md5 md5_hex);
 use Digest::SHA    qw(hmac_sha1_hex sha1 sha1_hex);
 use Encode         qw(find_encoding);
+use Errno          ();
 use Exporter       qw(import);
 use File::Basename qw(dirname);
 use Getopt::Long   qw(GetOptionsFromArray);
@@ -13,7 +14,7 @@ use IO::Compress::Gzip;
 use IO::Poll qw(POLLIN POLLPRI);
 use IO::Uncompress::Gunzip;
 use List::Util         qw(min);
-use MIME::Base64       qw(decode_base64 encode_base64);
+use MIME::Base64       qw(decode_base64 encode_base64 encode_base64url);
 use Mojo::BaseUtil     qw(class_to_path monkey_patch);
 use Pod::Usage         qw(pod2usage);
 use Socket             qw(inet_pton AF_INET6 AF_INET);
@@ -31,6 +32,7 @@ use constant CRYPTX => $ENV{MOJO_NO_CRYPTX} ? 0 : !!(eval {
   CryptX->VERSION('0.080');
   1;
 });
+use constant WIN32_API => eval { require Win32::API };
 
 # Check for monotonic clock support
 use constant MONOTONIC => !!eval { Time::HiRes::clock_gettime(Time::HiRes::CLOCK_MONOTONIC()) };
@@ -82,8 +84,8 @@ our @EXPORT_OK = (
   qw(b64_decode b64_encode camelize class_to_file class_to_path decamelize decode decrypt_cookie deprecated dumper),
   qw(encode encrypt_cookie extract_usage generate_secret getopt gunzip gzip header_params hmac_sha1_sum),
   qw(html_attr_unescape html_unescape humanize_bytes md5_bytes md5_sum monkey_patch network_contains punycode_decode),
-  qw(punycode_encode quote scope_guard secure_compare sha1_bytes sha1_sum slugify split_cookie_header split_header),
-  qw(steady_time tablify term_escape trim unindent unquote url_escape url_unescape xml_escape xor_encode)
+  qw(punycode_encode quote random_bytes scope_guard secure_compare sha1_bytes sha1_sum slugify split_cookie_header),
+  qw(split_header steady_time tablify term_escape trim unindent unquote url_escape url_unescape xml_escape xor_encode)
 );
 
 # Aliases
@@ -176,9 +178,7 @@ sub extract_usage {
 }
 
 sub generate_secret {
-  return Crypt::Misc::encode_b64u(Crypt::PRNG::random_bytes(128)) if CRYPTX;
-  srand;
-  return sha1_sum($$ . steady_time() . rand);
+  return encode_base64url(random_bytes(128));
 }
 
 sub getopt {
@@ -342,6 +342,36 @@ sub quote {
   my $str = shift;
   $str =~ s/(["\\])/\\$1/g;
   return qq{"$str"};
+}
+
+sub random_bytes {
+  my ($count) = @_;
+  croak 'Byte count must be a defined, non-negative integer' unless defined $count && $count =~ /\A\d+\z/a;
+  if (CRYPTX) {
+    return Crypt::PRNG::random_bytes($count);
+  }
+  elsif (WIN32_API) {
+    state $genrand = Win32::API->new('advapi32', 'INT SystemFunction036(PVOID RandomBuffer, ULONG RandomBufferLength)')
+      or die "Could not import SystemFunction036: $^E";
+    return '' if $count == 0;
+    my $buffer = chr(0) x $count;
+    $genrand->Call($buffer, $count) or croak("Could not read random bytes");
+    return $buffer;
+  }
+  elsif (-e '/dev/urandom') {
+    open my $fh, '<:raw', '/dev/urandom' or die "Couldn't open /dev/urandom: $!";
+    my ($result, $offset) = ('', 0);
+    while ($offset < $count) {
+      my $read = sysread $fh, $result, $count - $offset, $offset;
+      next                                 if not defined $read and $!{EINTR};
+      croak("Could not read random bytes") if not defined $read or $read == 0;
+      $offset += $read;
+    }
+    return $result;
+  }
+  else {
+    die "Can't find source of randomness on your system, consider installing CryptX";
+  }
 }
 
 sub scope_guard { Mojo::Util::_Guard->new(cb => shift) }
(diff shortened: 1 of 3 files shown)


lib/Mojo/Util.pm
```diff
@@ -175,10 +177,31 @@ sub extract_usage {
   return unindent($output);
 }
 
+sub generate_random {
```
robrwo: Why not call it "random_bytes" for consistency with other modules? It's not really "generating" bytes so much as reading them from something else.

lib/Mojo/Util.pm
```diff
@@ -175,10 +177,31 @@ sub extract_usage {
...
+    }
+    return $result;
+  }
+}
+
 sub generate_secret {
-  return Crypt::Misc::encode_b64u(Crypt::PRNG::random_bytes(128)) if CRYPTX;
-  srand;
-  return sha1_sum($$ . steady_time() . rand);
+  return encode_base64url(generate_random(128));
```
robrwo: Note that this fixes CVE-2024-58135

robrwo (approved):

lib/Mojo/Util.pm
Leont: Because there already was a `generate_secret` so I tried to be consistent with that, but you may have a point.

lib/Mojo/Util.pm
Leont: I do note that in the PR (but could also do that in the commit message).

kraih: Note that i do not consider that CVE a security issue, it misrepresents the facts and is not exploitable.

kraih: To be clear, that means this PR will have to be accepted as a new feature and pass a vote. Personally, i'm especially concerned about value to the majority of our users and long term maintainability.

lib/Mojo/Util.pm
```diff
@@ -6,14 +6,15 @@ use Data::Dumper   ();
...
 use Exporter       qw(import);
 use File::Basename qw(dirname);
 use Getopt::Long   qw(GetOptionsFromArray);
 use IO::Compress::Gzip;
 use IO::Poll qw(POLLIN POLLPRI);
 use IO::Uncompress::Gunzip;
 use List::Util         qw(min);
-use MIME::Base64       qw(decode_base64 encode_base64);
+use MIME::Base64       qw(decode_base64 encode_base64 encode_base64url);
 use Mojo::BaseUtil     qw(class_to_path monkey_patch);
```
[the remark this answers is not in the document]
Grinnz: Perl 5.16 comes with MIME::Base64 3.13, which supports this method. Non-issue

lib/Mojo/Util.pm
```diff
@@ -175,10 +177,31 @@ sub extract_usage {
...
+  if (CRYPTX) {
+    return Crypt::PRNG::random_bytes($count);
+  } elsif (WIN32_API) {
+    state $genrand = Win32::API->new('advapi32', 'INT SystemFunction036(PVOID RandomBuffer, ULONG RandomBufferLength)') or die "Could not import SystemFunction036: $^E";
+    return '' if $count == 0;
+    my $buffer = chr(0) x $count;
+    $genrand->Call($buffer, $count) or Carp::croak("Could not read random bytes");
+    return $buffer;
+  } else {
+    open my $fh, '<:raw', '/dev/urandom' or die "Couldn't open /dev/urandom: $!";
```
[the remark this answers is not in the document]
Grinnz: These are indications of general issues with retrieving randomness, not issues relevant to the caller. I think die is appropriate.

lib/Mojo/Util.pm
Grinnz: I agree, random_bytes would be clearer.

lib/Mojo/Util.pm
```diff
@@ -175,10 +177,31 @@ sub extract_usage {
   return unindent($output);
 }
 
+sub generate_random {
+  my ($count) = @_;
+  if (CRYPTX) {
+    return Crypt::PRNG::random_bytes($count);
+  } elsif (WIN32_API) {
+    state $genrand = Win32::API->new('advapi32', 'INT SystemFunction036(PVOID RandomBuffer, ULONG RandomBufferLength)') or die "Could not import SystemFunction036: $^E";
```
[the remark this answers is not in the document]
Grinnz: I tend to agree with this, there is no reason to try the /dev/urandom path on windows regardless of the presence of WIN32_API, and it would be better to provide a clearer error message in this specific case so the user can remedy the situation.

lib/Mojo/Util.pm
```diff
@@ -722,12 +745,17 @@ function was called from.
 
   =cut
 
+=head2 generate_random
+
+ my $random = generate_random($bytes);
+
+Generate C<$bytes> worth of high entropy randomness
```
kraih: It really annoys me that @Copilot does not pick up on style inconsistencies like missing punctuation here.

lib/Mojo/Util.pm
Leont: I agree it's not exploitable. I do believe it's very useful functionality to have: needing randomness is a recurring thing in web development IME, for a variety of reasons.

lib/Mojo/Util.pm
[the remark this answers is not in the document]
Leont: Windows without `Win32::API` is not really a thing (non-Windows with Win32::API is apparently a thing because it does work on cygwin). I can see the argument for writing this more defensively, but I don't think it really matters in practice.

lib/Mojo/Util.pm
Leont: Fixed now.

lib/Mojo/Util.pm
```diff
@@ -175,10 +177,31 @@ sub extract_usage {
   return unindent($output);
 }
 
+sub generate_random {
+  my ($count) = @_;
```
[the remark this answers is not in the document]
Leont: Fair enough

lib/Mojo/Util.pm
```diff
@@ -175,10 +177,35 @@ sub extract_usage {
...
+  croak 'Byte count must be a defined, non-negative integer' unless defined $count && $count =~ /\A\d+\z/;
+  if (CRYPTX) {
+    return Crypt::PRNG::random_bytes($count);
```
guest20: What's the point of dragging around all the code to get good random and having the dependency? 

It's paying for the milk while renting the cow.

Leont: It's an optional dependency?

guest20: but why depend on it at all if there's mainline code about it? 

**Edit**: Or non-snippily, does the default RNG in this external module (ChaCha20) do a better job than `/dev/urandom` does?

Leont: It would be more faster, and also would work in bad/weird containerized environments or chroots where a decent /dev is not available.

guest20: Containers. Cool. That'd do it.

Grinnz:
I'm approving this once a decision is made on the function name - any other opinions on that?

lib/Mojo/Util.pm
kraih: It's not like we could get rid of the CryptX dependency, we introduced it for encrypted sessions and only use it for default config values because it was there already.

lib/Mojo/Util.pm
Leont: Renamed it to random_bytes

lib/Mojo/Util.pm
```diff
@@ -344,6 +344,36 @@ sub quote {
   return qq{"$str"};
 }
 
+sub random_bytes {
+  my ($count) = @_;
+  croak 'Byte count must be a defined, non-negative integer' unless defined $count && $count =~ /\A\d+\z/;
```
Grinnz:
```suggestion
  croak 'Byte count must be a defined, non-negative integer' unless defined $count && $count =~ /\A[0-9]+\z/;
```

lib/Mojo/Util.pm
```diff
@@ -344,6 +344,36 @@ sub quote {
...
+  if (CRYPTX) {
+    return Crypt::PRNG::random_bytes($count);
+  }
+  elsif (WIN32_API) {
+    state $genrand = Win32::API->new('advapi32', 'INT SystemFunction036(PVOID RandomBuffer, ULONG RandomBufferLength)')
+      or die "Could not import SystemFunction036: $^E";
+    return '' if $count == 0;
```
Grinnz, line 356: Maybe this should just be before the if/elsif; the /dev/urandom case also may blow up on count 0

Grinnz: Nevermind, I see that the while loop would not execute in that circumstance, all is well

lib/Mojo/Util.pm
Leont: Fixed (using //a)

Grinnz:
Commit messages need to be updated for change in function name

lib/Mojo/Util.pm
Leont: The /dev/urandom case handles that just fine, it's just that RtlGenRandom is picky about being passed a 0 length.

Grinnz (approved):

kraih (approved):

PR #2326, opened 2026-07-26
Use the distribution's CSPRNG for the WebSocket mask and handshake nonce

robrwo:
Mojo::WebSocket derived two RFC 6455 security values from Perl's rand():

    my $mask      = pack 'N', int(rand 9 x 7);                  # masking key
    my $challenge = b64_encode sprintf('%16u', int(rand 9 x 16)), '';

RFC 6455 s5.3 says of the masking key: "The masking key needs to be unpredictable; thus, the masking key MUST be derived from a strong source of entropy, and the masking key for a given frame MUST NOT make it simple for a server/proxy to predict the masking key for a subsequent frame." s4.1 says the Sec-WebSocket-Key "MUST be a nonce consisting of a randomly selected 16-byte value".

Neither holds:

  * '9 x 7' is the string "9999999", so the mask is drawn from [0, 9999999) -- about 23.3 bits of the 32-bit field, 0.23% of the specified space, and always with a zero top byte. Measured over 20000 frames: one distinct value of the top byte.

  * The nonce is 16 ASCII decimal digits, space-padded, then base64-encoded. Measured: 11 distinct byte values out of 256, all in 0x20..0x39.

  * rand() is drand48, seeded by perl from a 32-bit value, so both streams follow from at most 2**32 possibilities and a linear congruential generator's output constrains what comes next -- the specific property s5.3 forbids.

Masking is a defence-in-depth mechanism: it stops a non-WebSocket-aware intermediary being tricked into misreading frame bytes as a separate HTTP request. That is why the RFC states the randomness requirement as a MUST rather than a SHOULD, and why an exploit path runs through a mis-parsing proxy rather than directly against the peer.

The fix needs nothing new. Mojo::Util::random_bytes is already part of this distribution -- it prefers Crypt::PRNG, falls back to the Win32 API or /dev/urandom, and dies rather than degrading -- and Mojolicious already uses it for CSRF tokens in Mojolicious::Plugin::DefaultHelpers::_csrf_token. Only the WebSocket code was left on rand(). Both call sites now request exactly the width the RFC specifies: 4 bytes for the mask, 16 for the nonce.

Only clients mask (Mojo::Transaction::WebSocket sets `masked` for client transactions), so the affected path is Mojo::UserAgent->websocket.

Verified: the top mask byte goes from 1 distinct value to 256, distinct masks over 20000 frames from 19977 to 20000, and nonce byte values from 11 of 256 (0x20..0x39) to 256 of 256 (0x00..0xff). Masked frames still round-trip. Mojolicious' own test suite passes unchanged -- t/mojo/websocket*.t 4 files / 49 tests, and the full t/mojo/ 63 files / 1573 tests, identical before and after.

### DIFF
--- a/lib/Mojo/WebSocket.pm
+++ b/lib/Mojo/WebSocket.pm
@@ -55,8 +55,6 @@ sub build_frame {
 
   # Mask payload
   if ($masked) {
-    # RFC 6455 5.3: the masking key MUST be derived from a strong source of
-    # entropy and MUST NOT let a server or proxy predict the next one.
     my $mask = random_bytes(4);
     $payload = $mask . xor_encode( $payload, $mask x 128 );
   }
@@ -83,7 +81,6 @@ sub client_handshake {
   $headers->sec_websocket_version(13) unless $headers->sec_websocket_version;
 
   # Generate 16 byte WebSocket challenge
-  # RFC 6455 4.1: a nonce consisting of a randomly selected 16-byte value.
   my $challenge = b64_encode random_bytes(16), '';
   $headers->sec_websocket_key($challenge) unless $headers->sec_websocket_key;
 
--- a/lib/Mojo/WebSocket.pm
+++ b/lib/Mojo/WebSocket.pm
@@ -3,7 +3,7 @@ use Mojo::Base -strict;
 
 use Config;
 use Exporter   qw(import);
-use Mojo::Util qw(b64_encode dumper sha1_bytes xor_encode);
+use Mojo::Util qw(b64_encode dumper random_bytes sha1_bytes xor_encode);
 
 use constant DEBUG => $ENV{MOJO_WEBSOCKET_DEBUG} || 0;
 
@@ -55,8 +55,10 @@ sub build_frame {
 
   # Mask payload
   if ($masked) {
-    my $mask = pack 'N', int(rand 9 x 7);
-    $payload = $mask . xor_encode($payload, $mask x 128);
+    # RFC 6455 5.3: the masking key MUST be derived from a strong source of
+    # entropy and MUST NOT let a server or proxy predict the next one.
+    my $mask = random_bytes(4);
+    $payload = $mask . xor_encode( $payload, $mask x 128 );
   }
 
   return $frame . $payload;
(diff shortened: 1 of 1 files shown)


lib/Mojo/WebSocket.pm
```diff
@@ -55,8 +55,10 @@ sub build_frame {
 
   # Mask payload
   if ($masked) {
-    my $mask = pack 'N', int(rand 9 x 7);
-    $payload = $mask . xor_encode($payload, $mask x 128);
+    # RFC 6455 5.3: the masking key MUST be derived from a strong source of
```
Grinnz: Unnecessary comments. The need for strong entropy is clear, please keep to code changes.

kraih:
I have yet to hear a good explanation for why WebSocket masks need something stronger than plain old `rand()`. There is no attack vector here, the mask is merely obfuscating WebSocket messages so they cannot be mistaken for HTTP traffic. It's a backwards compatibility feature for old infrastructure that has not been upgraded with WebSocket support. In that context, does this PR offer any actual improvements? And i don't think anyone has ever been able to create a functioning cache poisoning attack with WebSockets.

kraih:
Of course you could prove with a benchmark that this has absolutely no performance impact, in which case i would not care about an actual attack vector.

robrwo:
It protects against cache poisoning by proxies. The RFC says to use an unpredictable source.  Built-in  rand is predictable and according to CPANSec analysis tools, the next frame can be predicted after 3 frames. Since this is unlikely to be used without HTTPS, it's below the security issue threshold but still worth hardening since random_bytes is now part of Mojo::Uril

kraih:
Just saw the `Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>`, afraid the contribution has to be rejected based on that. While I have nothing against AI assistants, Mojolicious requires a human to take full responsibility for every contribution.

robrwo:
> Just saw the `Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>`, afraid the contribution has to be rejected based on that. While I have nothing against AI assistants, Mojolicious requires a human to take full responsibility for every contribution.

I am taking full responsibility for it, and signed off on it. It was manually submitted by me, and edited by me.

Edit: I can re-submit it without the Co-Authored, but that would be a lie. While I think it's important to push back and question people who submit PRs that are identified as being or seem to be co-authored by AI, blindly rejecting PRs is just going to encourage people to hide this.

PR #2328, opened 2026-08-12
Replace invalid code points with replacement character in Mojo::Util->html_unescape

nielsproest:
### Summary
Numeric character references are passed straight to `chr`, so "&#xD800;" produces a lone surrogate, "&#1114112;" produces code point beyond Unicode, and "&#0;" produces a literal NUL.

This commit respects the WHATWG HTML standard requirements for character replacement.

### Motivation
The WHATWG HTML standard requires the parser to resolve surrogate, codepoints beyond unicode and NUL to U+FFFD according to: https://html.spec.whatwg.org/#numeric-character-reference-end-state

According to https://infra.spec.whatwg.org/#leading-surrogate surrogates are defined as:
```
A leading surrogate is a code point that is in the range U+D800 to U+DBFF, inclusive.
A trailing surrogate is a code point that is in the range U+DC00 to U+DFFF, inclusive.
A surrogate is a leading surrogate or a trailing surrogate.
```

With this patch, references to 0x00, surrogates and code points beyond 0x10FFFF now resolve to U+FFFD.

The spec also requires that references to the C1 control range (0x80 to 0x9F) are interpreted as windows-1252 bytes, so these are remapped through windows-1252 and therefore "&#x80;" now yields U+20AC.

### References
Incidentally this fixes the issue reported in mojolicious#2284


lib/Mojo/Util.pm
```diff
@@ -508,7 +516,11 @@ sub _entity {
   my ($point, $name, $attr) = @_;
 
   # Code point
-  return chr($point !~ /^x/ ? $point : hex $point) unless defined $name;
+  unless (defined $name) {
+    my $cp = $point !~ /^x/ ? $point + 0 : hex $point;
+    return "\x{FFFD}" if $cp == 0 || $cp > 0x10FFFF || ($cp >= 0xD800 && $cp <= 0xDFFF);
```
Grinnz: I don't believe it makes a difference for codepoints above 0xFF, but it would be more idiomatic to write this string as `"\N{U+FFFD}"`

t/mojo/util.t
```diff
@@ -242,6 +242,36 @@ subtest 'html_unescape (multi-character entity)' => sub {
   is html_unescape('&acE;'), "\x{223e}\x{0333}", 'right HTML unescaped result';
 };
 
+subtest 'html_unescape (invalid code points)' => sub {
+  is html_unescape('&#0;'),             "\x{fffd}",           'null character reference';
```
Grinnz: I would similarly recommend using the `\N{U+####}` syntax for these test strings.

t/mojo/util.t
```diff
@@ -242,6 +242,36 @@ subtest 'html_unescape (multi-character entity)' => sub {
...
+};
+
+subtest 'html_unescape (windows-1252 remap)' => sub {
+  is html_unescape('&#x80;'),  "\x{20ac}", 'euro sign';
+  is html_unescape('&#128;'),  "\x{20ac}", 'euro sign in decimal';
+  is html_unescape('&#0128;'), "\x{20ac}", 'euro sign with leading zero';
+  is html_unescape('&#x93;'),  "\x{201c}", 'left double quotation mark';
+  is html_unescape('&#x99;'),  "\x{2122}", 'trade mark sign';
+  is html_unescape('&#x9F;'),  "\x{0178}", 'latin capital letter y with diaeresis';
+  is html_unescape('&#x81;'),  "\x{81}",   'unmapped control is preserved';
```
Grinnz: For these test strings it could actually make a difference, as `\N{U+7F}` will return a different string (but equivalent character) on EBCDIC systems than ASCII systems. I think that is the behavior you want, since `chr` will also return a different string on these systems. Unfortunately I don't have a way to test and Mojolicious doesn't particularly care about EBCDIC so it doesn't matter that much, but `\N` should be the correct option for these sort of tests.

nielsproest: I have changed all the \x into \N.
If we want EBCDIC support we would probably need to use unicode_to_native, but i have not done this.

PR #2332, opened 2026-09-04
Added to_deep_hash() method into Mojo::Parameters

buralien:
### Summary
Allow deeply expanding query parameters which use bracketed notation using Mojo::Parameters::to_deep_hash()

### Motivation
Some HTML forms and frontend JavaScript libraries post data using query parameters such as "`foo[bar]=baz`", but currently there is no way for Mojolicious to expand "foo" into a hashref with "bar" as key and "baz" as it's value. Current `to_hash()` method returns this as `{"foo[bar]" => "baz"}`.


rawleyfowler:
Query parameters like that aren't really specified in any specification, and are not super common in the Perl world. I think this would be better as a Mojolicious plugin.

kraih:
If i remember correctly a plugin for this already exists.

PR #2333, opened 2026-09-19
Fix confusing crash when an unknown filter name is passed to optional/required

### LINKED ISSUE #2129: Can't locate object method "" via package "Mojolicious::Validator::Validation"
* Mojolicious version: 9.35
* Perl version: 5.34

In this example I'm using a filter that is not defined (say a typo)

```
#!/usr/bin/env perl

get '/' => sub ($c) {
  $c->render(template => 'index');
};
post '/' => sub ($c) {
    $c->validation->optional('test','foo');
    return $c->render(text => 'foo');
};
app->start;

__DATA__

@@ index.html.ep
% layout 'default';

<form action="/" method="POST">
    <input type="text" name="test" />
    <input type="submit" value="Submit" />
</form>

@@ layouts/default.html.ep
<!DOCTYPE html>
<html>
  <head><title><%= title %></title></head>
  <body><%= content %></body>
</html>
```

### Expected behavior
Ought to be some message in the log that the filter `foo` does not exist.

### Actual behavior
Run under `morbo` or so and hit the button. This gives us the rather unuseful error

```
Use of uninitialized value $cb in method lookup at .../Mojolicious/Validator/Validation.pm line 74.
[2023-11-13 16:32:52.60911] [38589] [error] [mSTxzMBC76Is] Can't locate object method "" via package "Mojolicious::Validator::Validation" ...
```

It comes from this code
```
  for my $cb (map { $self->validator->filters->{$_} } @filters) {
      @input = map { $self->$cb($name, $_) } @input;
  }
```
The `filters->{$_}` part is returning `undef` of the filter is unknown to the list. And thus we pass on `undef` as a call back to be used inside the loop. Where it fails the code crashes.

Note sure what the best bethod would be here, but something like

```
      Carp::croak sprintf("Error in filter list for '%s', at least one of these does not exist: '%s'\n",$name,join(',',@filters)) 
            unless defined $cb;
      @input = map { $self->$cb($name, $_) } @input;
```

might at least help to debug something like this.

hikmetba-bit:
## Summary

Fixes #2129.

`Mojolicious::Validator::Validation::optional()` (and `required()`, which calls it) mapped the list of filter names straight to callbacks before calling each one:

```perl
for my $cb (map { $self->validator->filters->{$_} } @filters) {
  @input = map { $self->$cb($name, $_) } @input;
}
```

If a filter name is unknown (typo, or the filter was never registered), `$self->validator->filters->{$_}` is `undef`, and `$self->$cb(...)` then crashes with:

```
Use of uninitialized value $cb in method lookup at .../Validation.pm line 74.
Can't locate object method "" via package "Mojolicious::Validator::Validation"
```

— which gives no indication of which filter name was actually the problem, as the original reporter found while debugging a typo'd filter in production.

## Fix

The loop now walks the filter names directly and croaks with a clear message naming the offending filter as soon as it finds one with no matching entry, before ever attempting the method call:

```perl
for my $filter (@filters) {
  my $cb = $self->validator->filters->{$filter};
  Carp::croak qq{Unknown filter "$filter"} unless defined $cb;
  @input = map { $self->$cb($name, $_) } @input;
}
```

`Carp` was already imported (`use Carp ();`) and this same `Carp::croak` pattern is already used elsewhere in the same file (`BUILD_DYNAMIC`), so this stays consistent with the existing code style.

Added a `Changes` entry and a regression test in `t/mojolicious/validation_lite_app.t` (`Unknown filter` subtest) asserting the new, clear error message.

## Verification

**This environment's Perl install could not load Mojolicious itself** — core modules it depends on transitively (`Pod::Usage`, and something in `File::Spec` resolution) are broken/missing in this particular install, so I could not run the real test suite (`prove -l t/mojolicious/validation_lite_app.t`) here.

To still verify the actual logic change (not just eyeball it), I ported just the affected loop — old and new versions — into a standalone script using only `Carp` and `Test::More` (both of which *do* load correctly here), and confirmed:
1. The old logic really does crash with the exact confusing `Can't locate object method "" via package ...` message on an unknown filter name.
2. The new logic croaks with the clear `Unknown filter "<name>"` message instead.
3. The new logic is unchanged for a known filter (no regression).
4. The new logic is unchanged when no filters are passed at all (the common case).

All four checks passed. I could not run the actual `t/mojolicious/validation_lite_app.t` file I added the regression test to, or the wider suite, in this environment — please have CI/a reviewer confirm it passes before merge.

## Test plan
- [ ] CI: `prove -l t/mojolicious/validation_lite_app.t` passes, including the new "Unknown filter" subtest
- [ ] CI: full test suite shows no regressions
- [ ] Manual: reproduce the original crash on `main`, confirm this branch instead raises a clear `Unknown filter "..."` error

🤖 Generated with [Claude Code](https://claude.com/claude-code)


kraih:
Afraid we do not accept PRs with a `Co-Authored-By` an AI agent. You may of course use AI assistants, but a human has to take full responsibility for the patch.

PR #2334, opened 2026-09-19
Document that Mojolicious::Lite sets the default Mojo::UserAgent::Server app

### LINKED ISSUE #2283: Document somewhere that Mojolicious::Lite initializes Mojo::UserAgent::Server.
It is very surprising how UA knows which server to send requests.

```perl
use Mojolicious::Lite;
use Mojo::UserAgent;


get '/test' => sub{ shift->render( text => 'Ok' ) };

print "Response: " .Mojo::UserAgent->new->get( '/test' )->result->body ."\n";

1;
```

Document somewhere that Mojolicious::Lite changes Mojo::UserAgent::Server.

I found nothing here https://docs.mojolicious.org/Mojolicious/Lite or here https://docs.mojolicious.org/Mojo/UserAgent
and even here https://docs.mojolicious.org/Mojo/UserAgent/Server

It seems this could be bit dangerous, because if a 3rd party module loads Mojolicious::Lite, then UA in current module module is not just clean UA:

```perl
use Mojo::UserAgent;
use Some::Module;

my $ua =  Mojo::UserAgent->new;
```

Though, overall I like the idea that if the URL is not absolute, then the request goes to the current application.

hikmetba-bit:
## Summary

Fixes #2283.

`Mojolicious::Lite::import()` has always contained:

```perl
# Make sure there's a default application for testing
Mojo::UserAgent::Server->app($app) unless Mojo::UserAgent::Server->app;
```

which makes the first `Mojolicious::Lite` app loaded in a process become the process-wide default for `Mojo::UserAgent::Server`, so relative URLs passed to `Mojo::UserAgent` reach it automatically.

As the reporter found, this was previously undocumented in `Mojolicious::Lite`, `Mojo::UserAgent`, or `Mojo::UserAgent::Server`'s POD (I confirmed this by grepping all three before writing anything). It's a genuinely useful behavior, but also a real gotcha they pointed out: if a third-party module loads `Mojolicious::Lite` before your own application does, `Mojo::UserAgent::Server->app` may already point at that module's app instead of yours, since it's a process-wide, load-order-dependent global.

## Change

Added a short note to:
- `Mojolicious::Lite`'s `DESCRIPTION`, right after the existing SYNOPSIS/tutorial pointer.
- `Mojo::UserAgent::Server`'s `app` method POD, where the reporter specifically looked and found nothing.

Both cross-reference each other.

## Verification

Documentation-only change. Verified by reading the exact code path (`Mojolicious::Lite::import()`, the `unless Mojo::UserAgent::Server->app` guard) that performs this behavior, and confirming with `grep -rn "UserAgent::Server" lib/Mojolicious/Lite.pm lib/Mojolicious.pm lib/Mojo/UserAgent/Server.pm` that it was genuinely undocumented before this change. No test suite run needed for a POD-only change (nothing here changes runtime behavior).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


kraih:
Afraid we do not accept PRs with a Co-Authored-By an AI agent. You may of course use AI assistants, but a human has to take full responsibility for the patch.

